Issue #212, opened 2014-12-02
Switch to files earlier in the "Intro to Python" chapter


bmispelon:
As I mentionned in #207, I think switching to files earlier in the chapter would be beneficial:

* The usefulness of `print()` becomes more obvious (and we can remove the weird sentence that tries to explain `repr()` vs `print()` without talking about `repr`)
* We can format dictionaries with one key:value pair per line

mostaszewski:
I have submitted a pull request for second part of it.

olasitarska:
Is this now resolved @bmispelon?

olasitarska:
Confirmed with bmispelon that this is resolved.

PR #227, opened 2015-01-05
Format dictionaries

mostaszewski:
Format dictionaries with one key:value pair per line.
Regarding to issue #212


asendecka:
Thank you!

PR #229, opened 2015-01-06
Minor formatting corrections

papisz:
Some keywords were displayed like this: ~~~for~~~ instead of this: `for`.


asendecka:
Thank you!

PR #231, opened 2015-01-07
Update CONTRIBUTING.md

annaoz:
Hey Ola!

I was going through the tutorial on how to contribute and I noticed that it's not possible anymore to download the Gitbook editor. Instead you have to sign up on their website and open the book in your browser I guess. I actually couldn't do that and therefore couldn't get a new screenshot because I can't manage to download the tutorial to my computer via git :(. I always get the same error message:

$ git clone git@github.com:annaoz/organizer-manual.git
Cloning into 'organizer-manual' ...
Warning: Permanently added the RSA host key for IP address '192.30.252.130' to the list of known hosts.
Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights and the repository exists.

Either there's something wrong with the command or I'm making a mistake.

Hope it can be fixed!

Anna

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,7 +86,7 @@ Example:
     To git@github.com:miohtama/tutorial.git
        b37ca59..fe36152  contributing -> contributing
        
-If you don't want to download the Gitbook Editor app you can also go to the [Gitbook Website](http://gitbook.com), sign up for free and work direclty in your browser.
+If you don't want to download the Gitbook Editor app you can also go to the [Gitbook website](http://gitbook.com), sign up for free and work directly in your browser.
 
 # Making a pull request
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,13 +53,11 @@ Either use the Github app for your operating system (mentioned above) or `git` c
 
     git clone git@github.com:yourgithubusername/tutorial.git
 
-Download the [Gitbook Editor](http://help.gitbook.io/editor/README.html) app to your computer.
+Go to the [Gitbook Website](http://gitbook.com) and sign up for free.
 
-Then you can open the tutorial in Gitbook Editor (*File* > *Open book*).
+Then you can open the tutorial in your browser (*File* > *Open book*).
 
-![Gitbook](contributing/images/gitbook.png)
-
-Make any changes in the tutorial using the editor and then save changes (*Book* > *Save all*).
+Make any changes in the tutorial using Gitbook and then save changes (*Book* > *Save all*).
 
 Then commit the changes using `git` and push the changes to your remote Github repository.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -53,9 +53,9 @@ Either use the Github app for your operating system (mentioned above) or `git` c
 
     git clone git@github.com:yourgithubusername/tutorial.git
 
-Go to the [Gitbook Website](http://gitbook.com) and sign up for free.
+Download the [Gitbook Editor](https://github.com/GitbookIO/editor/releases) app to your computer.
 
-Then you can open the tutorial in your browser (*File* > *Open book*).
+Then you can open the tutorial in Gitbook Editor (*File* > *Open book*).
 
 Make any changes in the tutorial using Gitbook and then save changes (*Book* > *Save all*).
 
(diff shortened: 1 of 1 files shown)


kerstin:
Hm, the repo for the editor [still exists](https://github.com/GitbookIO/editor), as does this [download page](https://github.com/GitbookIO/editor/releases) where it says the latest version is from autumn, don't the downloads there work anymore?

kerstin:
Regarding cloning not working: could it be you're using an SSH key pair the public key of which you haven't added to GitHub yet? Have you already tried cloning via https instead of ssh?

I can clone fine from your fork as well as the [original DG](https://github.com/DjangoGirls/organizer-manual) remote, so the repositories work alright.

You can check your ssh keys uploaded to GH [here](https://github.com/settings/ssh). On your HD, you'll find your keys in a hidden directory called ".ssh" in your home/user directory. The public key (ending in .pub) is the one you'll want to compare. To try cloning with https, go to the repo you want to clone on GH and toggle the git clone url displayed on the lower right-hand side (just click on https to get the https version of it).

Btw. it's always a good idea to create separate GitHub issues for separate problems (like the GitBook issue/pull request vs. the git clone issue here). Makes it easier to keep track of things. (:

annaoz:
Thanks for the tips with the ssh keys. I'll look into that another time. I really only started contributing here a couple of days ago so I'm still learning :)

annaoz:
The link to the dowload of the Gitbook editor doesn't work for me. It tells me "file not found". When I go to http://gitbook.com/editor it tells me "The desktop Editor is no longer supported and being replaced by this new web editor, source code is still available on GitHub." Hence I figured that one probably should use the new web editor instead of dowloading an old version of the Gitbook editor.

kerstin:
I simply get redirected to the Gitbook website when I click the link to the editor you posted, I don't even get to see that note! But true, there's no word about the editor on their website anymore.

And I dunno why the download page on GitHub won't work for you, too bad. It's linked from the repository's README and contains downloads to a number of old versions. I reckon what they meant by not supporting the stand-alone editor anymore is that they won't add new features or provide support for it henceforth, not necessarily that one must/should ditch it right away. (I personally prefer native apps over web-/browser based ones, so I'm sad to see the stand-alone app go.)

annaoz:
When I click the hyperlink in the sentence "Download the Gitbook Editor app to your computer." which is http://help.gitbook.io/editor/README.html I get the error message "404 File not found", I don't get redirected anywhere so either it only doesn't work for me which would be strange since I'm using Google Chrome and everything else works fine or the link needs to be updated.

Right, anyway it would be helpful to add one sentence that there are two options: Download the editor or use the web editor so people can choose what they are most comfortable working with.

kerstin:
That's not the link I posted though! Check my comment above, I'm linking to the GitHub repo for the editor.

And yup, listing both options in the DG README sounds like a good idea. Maybe including a note that the stand-alone app is being phased out but can still be used, just so people don't get confused?

annaoz:
I know that's not the link you sent but that's the link that is currently in the tutorial and that people will probably click on and then discover as well that it doesn't work that's why it needs to be changed.

That sounds good!

kerstin:
Ahh, got it now!

annaoz:
Thank you for helping me with the ssh key. I was able to generate and add one to GitHub and successfully cloned the repository :) What do you think about putting 1-2 sentences about ssh keys into the "Contribute" section? Since the tutorial is aimed at beginners I can imagine that other people that have never worked with GitHub before might run into the same problem.

kerstin:
You're welcome!

About including it in Contribute: I'm not sure, but I think I personally wouldn't. Using git and GitHub works fine with https, which should be the default (which is why I was guessing you accidentally switched the git clone URLs to ssh at some point) and doesn't require changing any settings or an understanding of ssh. In the long run, I think ssh is preferable but if you don't really know what you're doing, it might create more confusion and problems than just sticking with https, especially as a beginner (e.g. accidental sharing of private key, forgetting about replicating the process on other machines or websites etc.).

keimlink:
GitHub has an excellent documentation on how to fork and clone a repository and how to create a pull request. So I would suggest to link to GitHub's documentation instead explaining it in the contribution guide.

olasitarska:
good point @keimlink !

annaoz:
I updated the link to the download site for Gitbook editor (old one wasn't working), added a sentence and link that you can also use Gitbook in your browser and added a link to the documentation like Markus suggested.

kerstin:
:+1:

asendecka:
Thank you @annaoz :) :cake: :heart_eyes:. Thank you @kerstin and @keimlink!

PR #239, opened 2015-01-25
startproject commands are the same for win/posix

HonzaKral:
This means we don't have to have two lines for windows and posix systems.

Thanks to @lenarother and Nika for pl/uk translation!


bmispelon:
Thanks, dziękuję, дякую! :penguin:

PR #241, opened 2015-01-25
Don't include empty date div if there is no date

HonzaKral:
One of the attendees submitted this change to not have the empty div appear if there is no date.

No idea about frontend, just passing things along.


bmispelon:
I guess it makes sense.

Thanks, Honza.
Thonza! :penguin:

PR #247, opened 2015-02-05
Better instructions to run `django-admin startproject`

encukou:
Move the note about virtualenv back, so the command is right after the instruction to run it.

Also, include the prompt in the command, since the instructions explicitly say to not type it.


aniav:
:+1: :+1:

PR #254, opened 2015-02-06
Deploy - Merge changes in mysite/settings.py

martinsvoboda:
Merge two changes of mysite/settings.py file together to one block of code. Previous version was confusing for girls.


aniav:
Thank you! :)

Issue #249, opened 2015-02-05
Use __lte=now() instead of __isnull=False when filtering published Posts


HonzaKral:
It's easier to explain and more correct.

bmispelon:
I think it makes sense :pineapple:

aniav:
That is now fixed and merged. Thank you! :ok_hand:

PR #259, opened 2015-02-12
Introduce __lte=now() instead of isnull=False

aniav:
In #249 there is a suggestion to move from `isnull=False` to `__lte=now()`. I have done this and slightly changed the text in English and Polish version. I need someone to look at the Ukrainian version and tell me what change should I make to the one sentence that has been changed.

Previously there was:
>You can also get a list of all published posts. We do it by filtering all the posts that have `published_date`:

Now we have:
>You can also get a list of all published posts. We do it by filtering all the posts that have `published_date` set in the past:

In Ukrainian the sentence is currently:
>Також можна отримати список усіх опублікованих постів. Зробимо це відфільтрувавши усі пости, що мають `published_date`:


aniav:
@chupakabra can you help me with that? Should be an easy one ;)

aniav:
Ukrainian version updated thanks to @ruslansavenok. Woohoo! :)

PR #261, opened 2015-02-14
fix broken link in the project-wide README

ana-balica:
The link to `how_internet_works/README.md` from the main README gives 404.

Also [on the index page](http://tutorial.djangogirls.org/en/index.html) the same URL is broken. It's actually rather peculiar that the page is generated from the global `README.md` and not the from the language folder `en/README.md`.


aniav:
Thank you for you submission. The problem with internal links is that they won't work properly with the `.md` extension while not having the dot at the front. So unfortunatelly your commit won't work either. Imust have broken the links while fixing the i18n links in other places. Let me fix that and close your pull request when it will be done :)

aniav:
I have fixed that in 5be7ebd094bc1dc838ff06c8b7db7222eb31cb09. Thank you! :rabbit:

PR #265, opened 2015-02-26
Update Python to the latests version (from 3.4.2 to 3.4.3)

ana-balica:


olasitarska:
Thanks @ana-balica! :tulip:

PR #266, opened 2015-02-27
Update python version to 3.4.3 in Chapter 11 - Deploy!

ProProgrammer:

PR #268, opened 2015-02-27
Revert back to python version to 3.4.2 in Chapter 11 - Deploy! as 3.4.3 not supported by Heroku as of now

ProProgrammer:
Reverts DjangoGirls/tutorial#266
Reverting the changes as it turns out that Heroku doesn't support `python-3.4.3` runtime as of now (https://devcenter.heroku.com/articles/python-support#supported-runtimes)
Will keep tracking updates to this document.


asendecka:
I already reverted the previous commit :). Thank you!

jorisoverzet:
Ok, thank you for the reference. My pullrequest does not touch chapter 11, only the local installation, so this probably won't affect the functionality of deploying on Heroku. But it could be a little bit confusing for the participant.

PR #269, opened 2015-03-06
Change some potentially confusing naming

AlexHill:
I feel that the line ``NameError: name 'names' is not defined`` could be a bit confusing. Too many ``name``s! I think it's clearer if the variable we're using has nothing to do with the string "name". I've also made the incorrect variable name a typo of the correct one, which ties in with the explanatory text that follows.

On the other hand, the example now differs from those before and after it, so I totally understand if you'd rather not make this change.


kerstin:
One solution would be to change the other examples as well. (;

aniav:
Ah damn, not this pull request.

aniav:
I am merging this and I will change the example in other languages. Thank you! :zap:

PR #270, opened 2015-03-06
Link not working

mauricioabreu:
Added 'en' as default redirect.


bmispelon:
Hi,

I don't think we need to add a default language.
The current link uses `./` at the beginning which makes it relative and so it should work in all languages.

Thanks! :dash:

mauricioabreu:
Ok, but the link still does not work. 
1 - Go to https://github.com/DjangoGirls/tutorial
2 - Try to click the link
3 - Page not found

:-)

bmispelon:
Oh, I see. You're navigating directly from github.

Our tutorial is meant to be viewed from http://tutorial.djangogirls.org.
I don't know if there's a way to support both github and our normal tutorial (which is translated in 3 languages).

mauricioabreu:
Yep! It is ok! 
I did not know you were moving. Last time I checked this repo it had no translation to other languages.

Thanks for reviewing my patch though! :-)

PR #279, opened 2015-03-19
Python 3.4.3 fixes

jorisoverzet:
Bump python version to 3.4.3 for spanish and fix the mac os x download link for the different languages.

### DIFF
--- a/en/python_installation/README.md
+++ b/en/python_installation/README.md
@@ -39,7 +39,7 @@ Use this command in your console:
 
 ### OS X
 
-You need to go to the website https://www.python.org/downloads/release/python-342/ and download the Python installer:
+You need to go to the website https://www.python.org/downloads/release/python-343/ and download the Python installer:
 
 * download the *Mac OS X 64-bit/32-bit installer* *DMG* file,
 * double click to open it,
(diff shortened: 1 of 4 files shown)


ProProgrammer:
Python 3.4.3 not yet supported by Heroku
Here you go: https://devcenter.heroku.com/articles/python-runtimes

jorisoverzet:
@ProProgrammer Could be, but the tutorial already mentioned 3.4.3 since some other pull request, so I just aligned/finished it. If you think it won't work, or it bothers you, write your own pull request to 3.4.2

ProProgrammer:
Actually I had written a pull request myself to update python versions to 3.4.3 (#268) which was first accepted but had to immediately request for reverting the pull request as I later found this Heroku compatibility issue.

Lets hope one of the admins looks into this before merging this pull request

jorisoverzet:
Ok, thank you for the reference. My pullrequest does not touch chapter 11, only the local installation, so this probably won't affect the functionality of deploying on Heroku. But the different versions in local and in requirements.txt could be a little bit confusing for the participant.

olasitarska:
What about this now? We switched from Heroku to PythonAnywhere. Is there still a blocker?

aniav:
PythonAnywhere has 3.4.0 support only. I would close this for now. How do you think?

olasitarska:
@aniav +1

Issue #282, opened 2015-03-21
At the point where we are asked to get all posts, we would probably have existing posts


aychedee:
>>> Post.objects.all()
    [  ]

Would probably be populated with the posts that the user created at an earlier step... Same with the user objects a little later. They've already created a superuser

phalt:
@aychedee :+1: Spotted that literally seconds after you, made a PR for it!

bmispelon:
PR merged.

Thanks everyone! :bee:

PR #283, opened 2015-03-21
Reword tutorial to support stuff in database

phalt:
As mentioned in issue #282, we now have stuff in the dev database at this point in the tutorial.

These changes reword things slightly so it makes a bit more sense and avoids duplicate data being created or confusion.

:cake:


bmispelon:
Looking good!

Thanks! :frog:

PR #284, opened 2015-03-21
Activate virtualenv after command prompt restart

doriczapari:
My team got quite confused when they had to reopen the command line after doing the Heroku thingies and the next commands weren't working. It would be nice to have a reminder here because activating the virtualenv was ages ago :)

PR #285, opened 2015-03-21
activate virtualenv in Start Dj Project

eppyk:
Attendees usually find it confusing to get back to Django Installation chapter each time they need to activate the virtualenv, so I thought it may be handy to have it here too (as usually this is the chapter they go to directly after Intro to Python).


bmispelon:
This is the same as #284, isn't it?

eppyk:
Dori's in Deploy section, and mine is in Starting Django project.

aniav:
Thank you! :bee:

PR #286, opened 2015-03-21
Update README.md

eppyk:
Added "http" in the link to the Lobster font; without it, it doesn't seem to work.


phalt:
Localhost not playing nicely with protocol relative links? Usually there is no problem with this. If this prevents the issue then :+1:

aniav:
Great catch! :sparkles:

PR #287, opened 2015-03-21
Wrong usage of word "set"

haxoza:
Get rid of improper usage of word "set" in favour of word "sequence". From Python point of view a set keeps unique values. A string is not set of characters, it's a sequence. The same about "instructions" in a function.


asendecka:
Thank you! :dancer: :balloon:

haxoza:
Thanks for merging it! :tada:

PR #288, opened 2015-03-22
Fix broken link for getting started

AnSavvides:
It looks like there are four entries for `how_internet_works` (one each for English, Spanish, Polish & Ukrainian?) - I am defaulting to the English one.

#DjangoSprintLondon


phalt:
:+1: Good spot.

phalt:
:question:  Looking at this a bit more it seems that when browsing the docs on github this fixes the broken link, but I think gitbook works out the language for us and sets up the right link? @bmispelon would know more, I think.

AnSavvides:
Aha! I see! Let's see what @bmispelon has to say then. I will try entice him with some emoji: :cat: :snake:

kerstin:
Is there a chance to correct `how_internet_work` to `how_the_internet_works` in the process? It's been bugging me from the beginning. :P

AnSavvides:
Sure thing! Makes perfect sense :P

AnSavvides:
@kerstin that's all done :)

kerstin:
Cool! I just remembered there was something about this at some point in the past, so went looking for it after I wrote my comment.

It's #119 though I don't know what the exact problems (plural) were that were referred to in that issue except for broken links IF someone, somewhere had previously linked to only that one chapter. (Which I honestly think is unlikely, or at least wouldn't affect a myriad of people.)

AnSavvides:
Ah! Well, let's wait and see what @bmispelon and/or @olasitarska think :)

aniav:
Ok, so we have two things here.

:zap: First is changing the link - I can't merge that. The links in gitbook are relative and are generated during build time. I have checked them - they work properly on gitbook. They won't always work properly here, especially for the main README.md file.

:zap: The second thing is adding the "the" in how_the_internet_works. I have added this in 80b38f1.

I am closing this pull request. Thank you all!

AnSavvides:
:

PR #292, opened 2015-03-25
Fixing Python Version Link

StephenDonaldson:
Changing URL from 342 to 343 so that it links to version 3.4.3 and not 3.4.2.

### DIFF
--- a/en/python_installation/README.md
+++ b/en/python_installation/README.md
@@ -39,7 +39,7 @@ Use this command in your console:
 
 ### OS X
 
-You need to go to the website https://www.python.org/downloads/release/python-342/ and download the Python installer:
+You need to go to the website https://www.python.org/downloads/release/python-343/ and download the Python installer:
 
 * download the *Mac OS X 64-bit/32-bit installer* *DMG* file,
 * double click to open it,


ProProgrammer:
I know it can be tedious but please look for previously created pull requests before submitting a new one.
https://github.com/DjangoGirls/tutorial/pull/279 and https://github.com/DjangoGirls/tutorial/pull/268

StephenDonaldson:
I did; the issue still persists that this is inconsistent with the text afterwards. Either this link needs to be changed or the text after needs to be changed. Currently it's causing confusion for users of the tutorial.

ProProgrammer:
@asendecka Should all such pull requests be closed and instead one be created for change all mentions of Python 3.4.3 to Python 3.4.2 and we keep that consistent at least till heroku starts supporting Python 3.4.3 environment.

Although it doesn't make much of a difference if we have Python 3.4.3 on our local system and 3.4.2 in runtime for heroku but this can be confusing keeping in mind this tutorial is aimed for beginners.

aniav:
I believe we don't have any specification as how such situations should be treated. Until this point most pull requests consisted of a full change. I believe we can just wait a bit until Heroku updates its runtime and then merge both pull requests. Is this a good solution? :)

aniav:
Ok, sorry about me being messy. I am closing this pull request in favour of #279 which has this change and more - it will be cleaner with one PR for this. Thank you!

ProProgrammer:
Very well :+1:

ProProgrammer:
However a pull request has to be submitted to change all references of 3.4.3 to 3.4.2 until Heroku starts supporting python 3.4.3 runtime as I believe people find it confusing to first having to specifically be instructed to install and setup 3.4.3 on their local machines but then be asked to mention 3.4.2 in runtime well at the time of deployment.

aniav:
There i fixed it for you in 83fd4e5 :)

ProProgrammer:
Superb :+1:

PR #293, opened 2015-03-27
Minor updates to "Starting a django project"

sodevious:
* Reorder timezone settings to the order they are displayed in settings.py
* Explain why the `.` is needed when using the startproject command


aniav:
Thank you! :sparkles: :bouquet:

PR #303, opened 2015-03-29
Fixed #302 -- Added .pyc files to gitignore

### LINKED ISSUE #302: Add .pyc files to .gitignore

MarkusH:
Thanks @antonagestam for the report


aniav:
I am really not sure we need to add this in this specific tutorial where we work on python 3. We had a talk about it in #291. Can someone else give their opinion on it? @bmispelon maybe?

MarkusH:
Ah, must have missed that issue. I'll add `__pycache__` instead.

bmispelon:
I agree that this should not be necessary with Python3 but I've seen it happen a few times now and I guess it doesn't cost us much to put it in the `.gitignore` anyway.

While we're there, we might as well add `.pyo` files too (I'm not sure how those get generated but I've seen them before in some cases).

aniav:
Ok, seems legit. I was wondering if this wouldn't make the thing more complicated instead of easier. But if you say it's ok I am also ok with it! :)

MarkusH:
I haven't seen anybody asking about the content of that file, tbh. It's mostly literally copied and pasted into their code base.

bmispelon:
It's been my experience as well. Especially in the deploy chapter, people just copy/paste without much thinking (which sucks but it's another issue altogether).

I say :shipit: :)

bmispelon:
Danke! :snail:

aniav:
:cat2: :tiger2:

PR #305, opened 2015-03-30
Advise readers not to use non-ascii characters

xavierdutreilh:
Yesterday, I was helping my team to set up their laptop for the first DjangoGirls Paris and we encountered an issue during one of the installations. It appears that virtual environments do not work well with paths containing non-ascii characters (in particular, French accents).

So, I updated the tutorial to advise readers to be careful when creating their virtual environment in a directory whose path may contains accents and special characters.

### DIFF
--- a/en/django_installation/README.md
+++ b/en/django_installation/README.md
@@ -31,7 +31,7 @@ To create a new `virtualenv`, you need to open the console (we told you about th
 
     C:\Users\Name\djangogirls> C:\Python34\python -m venv myvenv
 
-where `C:\Python34\python` is the directory in which you previously installed Python and `myvenv` is the name of your `virtualenv`. You can use any other name, but stick to lowercase and use no spaces. It is also good idea to keep the name short - you'll be referencing it a lot!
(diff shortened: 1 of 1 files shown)


bmispelon:
Which operating system was it?

I've just tried on my linux machine and using accented letters wasn't an issue.

aniav:
I believe we had exactly the same problem for Windows machines.

xavierdutreilh:
Sorry, @bmispelon, I forgot to specify the environment. It was under Windows 7. And yes, like you said, in any *nix environments, it works well.

en/django_installation/README.md
```diff
@@ -40,7 +40,7 @@ It will look like this:
 
     ~/djangogirls$ python3 -m venv myvenv
 
-`myvenv` is the name of your `virtualenv`. You can use any other name, but stick to lowercase and use no spaces. It is also good idea to keep the name short - you'll be referencing it a lot!
+`myvenv` is the name of your `virtualenv`. You can use any other name, but stick to lowercase and use no spaces, accents or special characters. It is also good idea to keep the name short - you'll be referencing it a lot!
```
bmispelon: If the problem is only on windows, then this shouldn't be needed, right?

xavierdutreilh: Yes, you are right. I should not have edited this paragraph.

bmispelon:
Merci ! :rooster:

PR #307, opened 2015-03-30
Fix links in polish and ukrainian version

mrfuxi:
Fix links in polish and ukrainian version by making file structure as in other languages


aniav:
Well spotted! :sparkles: :ok_hand:

PR #310, opened 2015-04-03
Ukrainian language in this context should have different ending

webknjaz:


aniav:
Verified by @ruslansavenok. Thank you! :sparkles:

PR #311, opened 2015-04-03
Explicitly state the directory we should be in

BeataBak:
Hi there, love the tutorial! Had only this one minor issue (might be just me but worth checking):
When I came to this point in the tutorial I tried running the "git add -A ." command and it wouldn't stage all of my modified files. It turns out I was in the wrong directory and I didn't realise this was a problem.


asendecka:
Awesome! :confetti_ball: :gift_heart:

PR #315, opened 2015-04-04
add an explanation + screenshot re adding python to the path on windows

hjwp:
the windows python installer, for some insane reason, has an option to "add pyton.exe to the path" which is unchecked by default.  this change adds a note + screenshot to help ppl catch it.


hjwp:
will re-submit a clean-mergeable one.

PR #318, opened 2015-04-05
(tiny thing) "less than" vs "smaller than" as words for <

hjwp:
Tiny suggestion, assuming original author has English as a second language, apologies if not.  "less than" and "less than or equal to" are the ways that < and <= are usually put into words, at least in my experience.  compare, eg, the unittest library which has "assertGreater" and "assertLess".

With that said, I like "smaller" too!


olasitarska:
Make sense to me! Thanks Harry :+1: :penguin:

PR #326, opened 2015-04-09
'python3' doesn't work on Windows

eppyk:
'python3' command doesn't work on Windows :( must be 'python'.

### DIFF
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -212,6 +212,6 @@ The error you saw was because we when we deployed to Heroku, we created a new da
 
     $ heroku run python manage.py createsuperuser
 
-You should now be able to see your website in a browser! Congrats :)!
+You should now be able to see the admin page of your website in a browser! Congrats :)!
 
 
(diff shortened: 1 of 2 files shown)


aniav:
I am sure we have changed it there and back from `python` to `python3` and so on more than once already. I feel this should be throughly investigated and we should find the situations where you have `python` and not `python3` commands on Windows. Could someone give some more feedback?

eppyk:
I'm not qualified enough to investigate this, but I hope someone knows what the problem might be! :) Thanks for your reply Ania.

eppyk:
I suspected it's because on MacOS you already have Python2 installed, so you kind of have to differ between Python2 (being the default) and Python3 and that's why you need the command to be python3. Maybe it depends on whether you have another version of python already installed on your computer?
I don't really know, just guessing.

aniav:
I can't merge it automatically, but adding your changes by hand in a second! :heart: :sparkles:

PR #330, opened 2015-04-12
Conflating terms: POST (http method) vs Post (Django model)

Matt-Deacalion:
My students had a few difficulties comprehending some of the form logic. These difficulties came from the fact that they didn't understand how the `POST` attribute on the `request` object was related to a blog `Post`. This was made clear once I explained the different types of HTTP request (GET, POST etc.) and that the conflating terms were just a coincidence.


aniav:
Thank you! :+1:

PR #333, opened 2015-04-13
Changes in startproject section

nnkps:
Changed path of django-admin script for Windows and moved note section before command (so note about appending '.' to the end of the command will be read earlier and will not mess project - one of workshop participants told me it will be more clear)


olasitarska:
Good point, thanks! :+1: :bee:

PR #335, opened 2015-04-14
Syntax-highlight Python code blocks

hjwp:
In markdown / gitbook, there is a syntax for specifying that a codeblock uses a particular syntax -- it uses

    ```python
     code here
     ```

I've made the changes to the python intro (cf "save it!" section), just for the bits where we save code into a file.  Can do more chapters if y'all are happy with it.


bmispelon:
Awesomesauce! :cow2:

PR #337, opened 2015-04-14
Syntax-highlight ALL the things.

hjwp:
following #335, this puts all the other codeblocks into syntax-highlighted chunks.

PR #345, opened 2015-04-26
Syntax-highlight all the code listings.

hjwp:
This is a replacement for #337, removing a couple of things that had nothing to do with syntax highlighting.  Should be clean.  To double-check, I've found `git diff -w` gives good output.


olasitarska:
Thanks @hjwp! Massive amount of work. You're the best. :heart_eyes_cat:

PR #352, opened 2015-05-10
add published date with filled forms

juliabiroprezi:
if we dont add published date for new and edited posts, they get filtered out in the post_list section


aniav:
Thank you! :sparkling_heart:

PR #353, opened 2015-05-12
Force plugins versions

SamyPesse:
This PR should fix your build that failed.

The problem is probably coming from one of your plugins that got installed twice (once by the `package.json` and once by the `gitbook install`), a new version of a plugin is probably adapted for GitBook 2.0.0 and failed with the legacy version.

BTW: :warning: :bomb: you should update to GitBook 2.0.0!!!


bmispelon:
I'm curious about the deletion of `package.json`. Is that file not necessary. It seems to have some useful info in there but I have no idea if it's needed or not...

SamyPesse:
It's not necessary since the introduction of the commande `gitbook install` (GitBook ~1.2.0), `gitbook install` will read the list of plugins from `book.json` and install it the correct versions.

Previously, your book was installing the plugins both using `npm install` and `gitbook install` (same results), but I have the impression that the `gitbook install` was replacing the plugins with newer versions since no versions were specified in the `book.json`.

Even if this correct is not fixing the issue, it will make the builds a little faster.

bmispelon:
Alright, let's try it!

Thanks for your help in debugging this! :gift:

bmispelon:
It worked! :tada: :balloon: :confetti_ball:

SamyPesse:
Yeah! Looks like it's working :) https://www.gitbook.com/book/djangogirls/djangogirls-tutorial/activity

olasitarska:
Thanks @SamyPesse! :penguin: 

I'm gonna look into upgrading to 2.0.0 later today :)

asendecka:
@SamyPesse Yay! Thank you for your help! :dancer: :tulip: :star: :rainbow:

SamyPesse:
You're welcome :beers:
@olasitarska let me know if you need help upgrading to 2.0.0.

PR #354, opened 2015-05-12
fix a markdown rendering issue in deploy chapter

hjwp:
without this, the `<your-username>` gets swalled because markdown things it's an html tag.


bmispelon:
Good catch, thanks! :snail: :guardsman:

hjwp:
:runner:

PR #357, opened 2015-05-14
minor pl-en inconsistencies

mkowalski:
1. It's better to confirm current working directory before initializing Git repo
2. In Chapter 10 some Posts have been added (via Django admin panel), so in Chapter 11 we will see them in console


asendecka:
Amazing work! Thank you :sunny: :hand: :cookie:

PR #358, opened 2015-05-15
fixing broken github link

gszjulcsi:
Github link contained an additional ' which broke the link when clicking on it.


bmispelon:
:jack_o_lantern: jó, köszi!

PR #359, opened 2015-05-16
Provide alternative runserver command for Windows

Almad:
On Windows, hostname may include non-ascii characters encoded by locale-specific encoding.

`runserver` tries to do `gethostbyaddr`, which then fails with `UnicodeDecodeError` as it always try to decode with `utf-8`

Providing IP explicitly (even if it's `0.0.0.0`) enforces Django to skip this resolv, serving as a workaround for [Python's bug #9377](http://bugs.python.org/issue9377).

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -101,6 +101,11 @@ You need to be in the directory that contains the `manage.py` file (the `djangog
 
     (myvenv) ~/djangogirls$ python manage.py runserver
 
+If you are on Windows and this fails with `UnicodeDecodeError`, use this command instead:
+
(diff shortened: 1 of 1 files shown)


bmispelon:
I remember seeing this error (and tracking it down to this very bug) but never knew that there was such an easy workaround.

What I had to do the last time it happened was to change the name of the computer (which is what windows uses to get the hostname).

I'm +1 on that change. Anyone disagrees?

olasitarska:
Go ahead!

keimlink:
Slick fix for this weird problem! :+1:

bmispelon:
Thanks! :koala:

PR #365, opened 2015-05-29
Fix readme link

garaud:


aniav:
Hello @garaud. Links in the turorial are prepared for gitbook and they are relative. Therefore I believe I can't accept your pull request. Sorry and thank you for your effort! :)

garaud:
Thanks you for the answer. It seemed me strange to get a 404 after clicking on the link from the Github page. I understand this should be relative for gitbook, you don't have to be sorry.

Great contents by the way !! Keep going.

kerstin:
This has been irritating me as well and I believe it confuses people who see the GH repo first.

Would it make sense to add a link to the tutorial on Gitbook for users who aren't already reading it from there? It could be put in brackets + explanation added so Gitbook readers don't get confused.

What do you think?

aniav:
I believe the README.md from the root folder isn't required by gitbook. We could overwrite it with other contents :)

kerstin:
Ahh! So is the "Details" page on Gitbook https://www.gitbook.com/book/djangogirls/djangogirls-tutorial/details based on the /en README https://github.com/DjangoGirls/tutorial/tree/master/en, not on the main README?

In that case, editing the main/root README to be less confusing for GitHub users might really make sense.

aniav:
Yes aaaand done! :)

garaud:
Nice !! Thanks. So my PR was not so useless.

kerstin:
:+1:

PR #385, opened 2015-06-04
Run python installation chapter through hemingway

Videocard:
Improved hard to read text and grammar in Python Installation chapter.

:tada:


aniav:
Crazy! :sparkles:

Issue #382, opened 2015-06-04
Incomplete/unclear sentence


Videocard:
In file python_installation/README.md line 19

See [this commit](https://github.com/DjangoGirls/tutorial/commit/a4670a521f086db53104eb2ea520669aa5c197fb)

:smile:

mcbhenwood:
Looking at this now. I don't think it's incomplete, I think it is just badly worded. Will issue a PR shortly.

mcbhenwood:
https://github.com/DjangoGirls/tutorial/pull/390

aniav:
:+1: :whale:

PR #390, opened 2015-06-04
Clarify connection between paragraph and its following helper image

mcbhenwood:
Lack of explicit phrasing and colon rendered the previous
version a little unclear. Fixes issue #382.


phalt:
:+1: nice one!

mcbhenwood:
Thanks - this Issue wasn't tagged at all. Should I only look for accepted Issues or just wade in, in general?

aniav:
I think you can just grab any issue. Some might be blocked by something and some mightnot be tagged at all. We haven't worked out a good policy yet :)

PR #393, opened 2015-06-04
Clarify whether commands should be run on PythonAnywhere

treyhunner:
While coaching at the Ensenada workshop I received questions from multiple students about whether commands needed to be run on PythonAnywhere.

This pull request provides some suggestions for answering some of these questions before they arise.  I am not particularly satisfied with the wording, but I haven't thought of anything better yet.

A couple alternative solutions:
- Add a section explaining that all commands after the deployment section are to be run in your local console unless it specifically says to run them on PythonAnywhere
- Change the appearance of code sections that should be run on PythonAnywhere (maybe change the color or font) and explain the difference

Feel free to close this pull request in favor of a better solution.

I have never coached students with the Heroku tutorial so I am not sure if these are new problems.

### DIFF
--- a/en/django_orm/README.md
+++ b/en/django_orm/README.md
@@ -12,7 +12,7 @@ It's easiest to learn by example. Let's try this, shall we?
 
 ## Django shell
 
-Open up your console and type this command:
+Open up your local console (not on PythonAnywhere) and type this command:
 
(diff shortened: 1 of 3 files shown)


aniav:
I think this is perfectly fine for the time being :)

PR #396, opened 2015-06-05
Fixed 377 - Drew more attention to the period in the command

mcbhenwood:
To draw attention to the missing period, I've made it the only thing which we really hammer home to the student in that section. I agree with the comments by @helenst that reminding the student not to type ```(myenv)``` etc is less important. If they do type that it will be an immediate failure (without repercussions) and so I have moved the explanation about this concept to *below* the command instructions.


aniav:
Awesome! :sparkles:

PR #401, opened 2015-06-06
Added list delete example and minor changes

blossomica:
I read the tutorial a couple days ago and noticed it has an example  to delete an item in a dictionary but non for the lists and that can leave a newbie wondering what is its equivalent so I added an example of list item removal.

I removed the programmers are lazy mostly because it was not mentioned previously as stated in that sentence and because I believe that, it is a stereotype that can give the wrong impression to some people. 

I added an additional explanation on why range does what it does because I felt that was incomplete since the notice told us to notice that but didn't clearly tell us that, that's how that function works. So I tried to kind of fix that.

Also I believe "programmers" sounded better than "computer people" so I changed that too.

I was thinking that it will be nice to have a glossary or dictionary for some of the technical terms on an appendix or at least show newbies on where to get access of a technical dictionary online. I'd like to hear your thoughts on that. 

These are my suggested edits I am open to feedback if there is anything that needs changes :)

Thank you.

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -216,7 +216,7 @@ Easy, right? If you want to add something to your list, you can do this by typin
     >>> print(lottery)
     [59, 42, 30, 19, 12, 3, 199]
 
-If you want to show only the first number, you can do this by using __indexes__. An index is the number that says where in a list an item occurs. Computer people like to start counting at 0, so the first object in your list is at index 0, the next one is at 1, and so on. Try this:
+If you want to show only the first number, you can do this by using __indexes__. An index is the number that says where in a list an item occurs. Programmers prefer to start counting at 0, so the first object in your list is at index 0, the next one is at 1, and so on. Try this:
 
     >>> print(lottery[0])
     59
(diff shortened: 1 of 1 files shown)


aniav:
Hey @blossomica! Thank you for the changes, good work done! :)
As for the glossary, we have recently added one, it's the `GLOSSARY.md` file. The terms defined there should be linked in the text in all occurances. There is just one term now but feel free to add more :)

blossomica:
Oops. I didn't see that. Thanks for bringing it to my attention and also letting me know I did a good work @ania :)

I have another question too. In the command line chapter we have move and copy in the table of commands but no example of them. Everything else in that table had an example previously done except for those two, and the way they are written in the table (with the line break probably because the table is too small) might confuse someone to think the two lines are two separate commands or that the line break is necessary. 

So I was thinking that to avoid confusion in that spot, it either gets removed because it also isn't used anywhere in the tutorial (plus there is a link already linking to additional commands if they are interested) with the pros of making the tutorial a bit shorter or we add in examples for move and copy at work (cons makes the tutorial longer but shows them how move/copy works). I can add in examples if needed but I'd like to see what others think is the best option.

aniav:
I believe we should avoid making the tutorial longer at all costs. We already have an issue about the tutorial being too long. From my understanding the additional commands are added there for consistency and because coaches might want to give them as additional excercises, so I would leave them the way they are, or less likely remove them.

blossomica:
I see. That's good to know. Thanks for the input on this :)

PR #404, opened 2015-06-07
django-admin not django-admin.py for windows

ShaguftaMethwani:
Writing django-admin.py simply opens the file with the default text editor.


aniav:
Thanks! :sparkles:

PR #408, opened 2015-06-19
Create separate installation chapter

helenst:
Install instructions are split out into separate markdowns and included (yay gitbook 2) from both the installation section and inline in their original positions in the tutorial.

I'm not 100% about this - I *think* it flows reasonably well in both contexts... the alternative would be to take the installation bits out of the main tutorial completely and just have them all in one place together...

I've also inserted this at the start - it means the chapters get renumbered, it didn't seem to have any bad effects but perhaps somebody else will know better :smile:


aniav:
I think this does need some checking and building the book locally. Please do ping us here if we don't merge/comment it in the next weeks. Thank you! Hugs! :sparkles:

helenst:
of course :-) I will try not to let it get forgotten!

tpinecone:
I'm currently going through the tutorial to prepare for my first time coaching, and I think this will make the installation process much simpler for attendees and coaches!

PR #418, opened 2015-07-10
Cleanup some wording and example output

stennie:
Cleanup some wording and example output based on running through first part of tutorial with Python 3.4.3 and Django 1.8.


aniav:
Thank you! Bam! :sparkles:

PR #420, opened 2015-07-11
Add in OS specific terminal prompting for the changing directories

alexlatchford:


aniav:
Thank you! :sparkles:

Issue #419, opened 2015-07-10
Incorrect commit message


medullaskyline:
In the django_templates section, the (commit message) [https://github.com/DjangoGirls/tutorial/blob/master/en/django_templates/README.md#one-more-thing] `added views to create/edit blog post` should be something like `modified templates to display posts from database`.

williln:
Thanks @medullaskyline! I updated this in [422](https://github.com/DjangoGirls/tutorial/pull/422); hopefully it will get merged :+1:

asendecka:
Thank you @medullaskyline for spotting it and @williln for pull request :D. You are amazing! :mega: :strawberry: :rainbow:

PR #422, opened 2015-07-12
Update example commit comment

williln:
Addresses issue [#419](https://github.com/DjangoGirls/tutorial/issues/419); this is more correct with what the changes actually were.


asendecka:
:heart: :sunny: Thank you!

PR #426, opened 2015-07-18
Escape template tags

lhuriguen:
Add escape tags to form.as_p because it's not showing up.


helenst:
Thank you! :smile: :whale:

PR #427, opened 2015-07-19
clarify what to replace in the template section

mmchugh:
I was coaching the tutorial today, and everyone I watched go through this step replaced all 4 tags with `{{ posts }}` instead of replacing the full content, and were confused when their output did not match the screenshot.

This was the clearest I could word it, but open to changes to make it more clear.


aniav:
Thank you! :sparkles:

PR #428, opened 2015-07-19
Avoid confusion by using different directory name

craigsdennis:
While coaching this, I had 2 of my 3 students almost wipe their installation directory that we setup in the Installation step.  The difference between ~/djangogirls and ~/Desktop/djangogirls wasn't clear enough.  This rename should remove that confusion.

Thanks for all the awesome work!

### DIFF
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -136,15 +136,15 @@ Here it is!
 
 ### Create directory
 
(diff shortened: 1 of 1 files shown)


williln:
+1 to this! @craigsdennis I think everyone with merge access is bringing down the house at EuroPython, but this won't get lost in the ether!

helenst:
One merger here who is not at Europython, although I wish I was!

Anyway - this is awesome, thank you! :tulip:

PR #430, opened 2015-07-20
Spanish: Use the @ to refer to "otras y otros"

goliveirab:
I propose to use the @ to refer to "otras y otros" for example.


rapoulson:
I like this a lot--does anyone have thoughts on implementation? We'd have to make the change in Crowdin before publishing the new version of the translation.

rapoulson:
This never gained traction so I'm closing it but I'd be happy to revisit it if someone has the time to make the changes :smile:

PR #438, opened 2015-07-25
Clarify that there is no difference between single and double quotes

donaldharvey:


phalt:
:+1:

olasitarska:
Thanks @donaldharvey! Thanks @phalt for review! :)

PR #440, opened 2015-07-25
elif more detail and example: fixes #248

### LINKED ISSUE #248: elif is not really clear
While ```if``` and ```else``` parts are nice and easy to understand, Karolina had a question in which case does ```elif``` gets run.

We should make sure that ```if``` and ```else``` are covered first and then ```elif``` is introduced

donaldharvey:


phalt:
This is great. Is "test" the right word to use here? Perhaps "conditions" is nicer?

donaldharvey:
Yeah much better, thanks!

olasitarska:
Fantastic! Thanks! :palm_tree:

PR #447, opened 2015-08-03
Update README.md

henziger:
The git commit message was erroneously the same as for the next chapter. :)


helenst:
Nicely spotted :-) I think the adding of CSS should probably have its own commit step but this is an improvement so let's go with it for now :sunglasses:

PR #448, opened 2015-08-03
Fix paragraph regarding publish_date when creating post.

henziger:
The published_date is clearly set, fixed documentation to follow code behavior.


phalt:
Nice find @henziger, the change makes much more sense to me :dizzy:

PR #449, opened 2015-08-04
Don't set published_date when creating new post.

henziger:
Actually, for this code to work well with the first Extentions homework (http://djangogirls.gitbooks.io/django-girls-tutorial-extensions/content/homework/index.html) it's probably best to not set the published_date here. Otherwise it won't be possible to create drafts, except from the admin site.

Sorry for all the back and forth turns with this publish_date thing.

### DIFF
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -221,7 +221,6 @@ def post_new(request):
         if form.is_valid():
             post = form.save(commit=False)
             post.author = request.user
-            post.published_date = timezone.now()
             post.save()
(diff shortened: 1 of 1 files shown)


asendecka:
I agree that this is confusing in extensions. At the very beginning we did not set publish date in form at all. But we've changed that later, because not many people during the workshop managed to finish the tutorial and tutorial extensions. 

What about leaving this line in basic tutorial, but change it in tutorial extensions? So when we start talking about drafts and publish buttons?

henziger:
Totally agree with you, closing this request for now.

PR #450, opened 2015-08-05
Remove `form.save(commit=False)` from tutorial code and text.

candeira:
Remove `form.save(commit=False)` from tutorial code and text.

Rationale:

While doing the djangogirls tutorial, I noticed that ModelForm.save(commit=False) is given to newcomers as a reasonable way to acquire a form's populated model. This is an antipattern for several reasons:

  - It's counterintuitive. To a newcomer, it's the same as `save(save=no)`. 
  - It's a leaky abstraction. Understanding it requires understanding that the save method does two things: a) prepare the model, and b) save it to the database.
  - It doesn't name its effect or return well.
  - It forces the tutorial to explain why "save" means "not save, not really", while "save()" means "really save, really".

Separately, I've written a Django patch adding a new utility method, `ModelForm.get_updated_model()`, which addresses these issues.

The present patch modifies the django_forms chapter so the new `ModelForm.get_updated_model()` method is used, and the explanation is made clearer.

Changes:

  - used `post = form.get_updated_model()` in forms tutorial code.
  - changed English tutorial text correspondingly.
  - changed Spanish tutorial text correspondingly.
  - updated Spanish tutorial a bit where it lagged behind the English one.

Dependencies:

The present patch depends on a proposed patch to Django. It can only be merged if/when the Django patch is merged and the tutorial is upgraded to a later Django version.

https://github.com/django/django/pull/5105

### DIFF
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -189,14 +189,13 @@ We check if the form is valid and if so, we can save it!
 
 ```python
 if form.is_valid():
-    post = form.save(commit=False)
+    post = form.get_updated_model()
     post.author = request.user
     post.published_date = timezone.now()
     post.save()
 ```
 
-Basically, we have two things here: we save the form with `form.save` and we add an author (since there was no `author` field in the `PostForm` and this field is required!). `commit=False` means that we don't want to save `Post` model yet - we want to add author first. Most of the time you will use `form.save()`, without `commit=False`, but in this case, we need to do that.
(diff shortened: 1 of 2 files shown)


olasitarska:
Hi @candeira, thank you for taking the time to submit this pull request! As this is something we can't merge until your PR against Django master will be merged, and then new version of Django released, I'm gonna close this off now. Please reopen once Django 1.9 with your change is released. 

Thanks! :) :dolphin:

PR #452, opened 2015-08-08
Update README.md

synemitchell:
Mac OS command line change needs 'source' command at the beginning to work. I'm guessing this is also needed for the Windows command as well, but have no way to check.


helenst:
Thanks! :tulip: I think the windows one is fine.

PR #454, opened 2015-08-09
add virtualenv name to prompt

dgales4130:
Add virtualenv name to prompt for the collectstatic example.


williln:
Thanks @dgales4130! :bouquet: :dolphin:

PR #459, opened 2015-08-16
Suggest user friendly edits for Extend chapter

willingc:
Hi! I was reviewing the tutorial for mentoring DjangoGirls at DjangoCon :tada:.

I stumbled a bit over the Extend Your Application chapter  because I kept losing my place in the page scroll. There's a lot of code snippets that look very similar and very few section headings as a visual cue for where you in the page.

I suggest adding a few subsection titles and a few other edits to be a bit more specific with terms like "it". Please use what you think might be helpful. I like this section quite a bit and making it more user friendly would help :sunflower: :rabbit: :tulip:

### DIFF
--- a/en/extend_your_application/README.md
+++ b/en/extend_your_application/README.md
@@ -8,7 +8,7 @@ The first thing we need in our blog is, obviously, a page to display one post, r
 
 We already have a `Post` model, so we don't need to add anything to `models.py`.
 
-## Create a link in the template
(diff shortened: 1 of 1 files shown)


asendecka:
Amazing work @willingc! You rock :D. :zap: :star: :strawberry: :cherries: :checkered_flag:

willingc:
Thanks for the merge @asendecka :ocean:

Closes #460

Issue #188, opened 2014-11-16
Gitbook plugin for toggling between Mac OS X / Windows / Linux versions of code examples


olasitarska:
It'd be awesome if we could make some kind of tabs to toggle between Mac OS X / Windows / Linux versions of code examples. Preferably it should also save it in cookies and defaults to chosen version.

helenst:
I've started looking into this.

helenst:
I stopped looking at this... it is probably best to address #314 first to minimise the need for this in the first place - and then possibly revisit.

ryangallen:
@olasitarska @helenst Hi Ladies!

What about a simple, subtle color difference between Unix and Windows examples? It does a pretty good job differentiating the two and would be a much faster fix for now. By introducing this convention at the start of the tutorial, we could eliminate the many `<p>Windows:</p>` elements as well. 

We could implement a class on the `pre` elements such as `os-unix`/`os-windows` to set a background and border color:

    pre.os-unix {
        background-color: honeydew;
        border: 1px solid lightgreen;
    }
    pre.os-windows {
        background-color: aliceblue;
        border: 1px solid lightblue;
    }

resulting in:

<img width="787" alt="screen shot 2015-09-08 at 5 23 28 pm" src="https://cloud.githubusercontent.com/assets/1379681/9749037/fd9e7314-564e-11e5-9d0c-c6d3da926801.png">

patjouk:
Hi evreyone!

Ola found [this](https://github.com/GitbookIO/plugin-codetabs#readme) the other day. Maybe we could give it a try?

patjouk:
PR #874 introduced a cool plugin called `sectionx`.

To create a block per operating system, use:

```
<!--sec data-title="Windows" data-id="{NameOfSection}_windows" data-collapse=true ces-->
Instruction for windows
<!--endsec-->

<!--sec data-title="Linux" data-id="{NameOfSection}_linux" data-collapse=true ces-->
Instructions for Linux
<!--endsec-->

<!--sec data-title="OS X" data-id="{NameOfSection}_OSX" data-collapse=true ces-->
Instructions for Mac OS X
<!--endsec-->
````

More instructions for `sectionx` are [available here](https://github.com/ymcatar/gitbook-plugin-sectionx).

We could also use it on those two chapters:

- [Introduction to Command Line](https://github.com/DjangoGirls/tutorial/tree/master/en/intro_to_command_line)
[...]

PR #462, opened 2015-08-20
install python on popular unix distrubions

sidrg:
FreeBSD and OpenBSD are important distributions of UNIX. OS X is also a UNIX. Installing an and using python on *BSD is quite normal.

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -4,6 +4,39 @@
 Django is written in Python. We need Python to do anything in Django. Let's start with installing it! We want you to install Python 3.4, so if you have any earlier version, you will need to upgrade it.
 
 
+### GNU/Linux
+
+It is very likely that you already have Python installed out of the box. To check if you have it installed (and which version it is), open a console and type the following command:
+
+    $ python3 --version
+    Python 3.4.3
+
+If you don't have Python installed, or if you want a different version, you can install it. How to install software like Python depends on you distribution.
+
+The following example is for Debian or Ubuntu.
+
+Type this command into your console:
+
+    $ sudo apt-get install python3.4
+
+
+#### OS X
+
+You need to go to the website https://www.python.org/downloads/release/python-343/ and download the Python installer:
+
+* Download the *Mac OS X 64-bit/32-bit installer* file,
+* Double click *python-3.4.3-macosx10.6.pkg* to run the installer.
+
+Verify the installation was successful by opening the *Terminal* application and running the `python3` command:
+
+    $ python3 --version
+    Python 3.4.3
+
+----
+
+If you have any doubts, or if something went wrong and you have no idea what to do next - please ask your coach! Sometimes things don't go smoothly and it's better to ask for help from someone with more experience.
+
+
 ### Windows
 
 You can download Python for Windows from the website https://www.python.org/downloads/release/python-343/. After downloading the ***.msi** file, you should run it (double-click on it) and follow the instructions there. It is important to remember the path (the directory) where you installed Python. It will be needed later!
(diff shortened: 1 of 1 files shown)


hjwp:
The tutorial is aimed at absolute beginners -- people who've never don't know any programming, and who've never used a command-line. I'm guessing hardly any of them will be using BSD? (but maybe there's something I don't know?).

The kind of person that I imagine uses BSD, I also imagine already knows how to install python -- that doesn't mean they can't get something out of the tutorial, just that I'm guessing they're probably going to skip the beginning sections about installation, what is a command-line, how the internet works, etc.

Maybe I'm wrong?

olasitarska:
I agree with @hjwp -- hardly anyone who applies for the workshop is using Linux, and if they're using Linux, they're usually already programming in different language or are using Ubuntu. 

I think this change would only add to confusion for people who don't know what BSD is, and will start to wonder what version is for them.

sidrg:
@hjwp , i think you are wrong. (sorry!)

do you know [pc-bsd](http://pcbsd.org) and stuff like that?

in my point of view freebsd/pc-bsd (in the bsd-world) is like debian/ubuntu (gnu/linux-world).
but you are absolutely right, if you wanna say that *bsd is not the mainstream (popular as linux). if mainstream is the basis of decisionmaking my effort is bad.

you are "promoting" fedora.
@olasitarska , how many are using fedora?
maybe naming fedora and explaining two different ways for installing packages there is confusing there too (or much more).

not being a djangogirl, i started reading this tutorial. yes, i am using *bsd and i am not a coder. i wanna learn (python and) django.

it was a try to make this tutorial better. (btw: thanks for creating stuff like that. i am not really sad if you do not merge it.)

sidrg:
@olasitarska , @hjwp : better? please check my wording! (my english is so horrible.)

aniav:
I can't agree to the changes made here. There are a few reasons to this:
* moving Debian and Ubuntu to be first instead of Windows - most of our attendees in Europe use Windows, in US it probably is Mac OS. There is no reason to place Debian and Ubuntu first. I mean - the order is not really important but if you are making any changes they should be better for the end users, not bringing more confusion
* introducing Debian and Ubuntu as something separate from "other systems" and placing Fedora there. Bringing confusion again

If you want to introduce FeeBSD and BSD systems to the tutorial please do that but don't change the whole chapter making it more confusing. It should be clear straightforward and not too complicated. The topic of Ooperating Systems is complicated by itself to many beginners and we don't want to make it less clear. I hope you understand that. Hugs! :)

yater:
@aniav :-/ :-( 

probably the order of items in a also important. for example, what is named first is suggested as the supposed best choice. or am i wrong?
btw: you can also look on it as alphabetical classification. :-D ;-)
to head windows is wrong (in my opinion). maybe it is political, but placing a product with that kind of license on top is wrong. we are talking about a tutorial for people who are new in this kind of world. (does this project is not part of the floss- movement? :-D i am just kidding, but maybe you think about it one moment.)
btw: maybe it is not wrong for someone you is learning django to have read the term 'Debian or Ubuntu' if they never heard it before. most likely these people a confused about reading it but want is quite normal because it is new for them and they are facing up to know that there are different operating systems.

btw: i think tat i did not "change the whole chapter". i just shifting (c&p) around to make it more clear. (when i had done this changes i was add the bsd-foo what i had to do on my machine.) i was confused by the focus on fedora (with the naming of two versions). (i am not using fedora but if people are not really familiar with it they were be also confused and there is no tip to figure out what version they are using. do you know what i mean?)

(@aniav "feebsd" instead of "freebsd" is the funniest typo i have seen for a while. :-) thanks! :-) that made (the start of) my day!)

olasitarska:
I agree with @aniav on this one. This tutorial is not about encouraging people to change their operating system, or promoting either closed or open software, but people learning to build websites with systems they arleady have. Mac OS X and Windows are winning when it comes to popularity among our attendees. Not even one on 10 people who apply for the workshop is using Linux.

aniav:
@yater keeping it straight: we are going to accept the PR only if the original order of the systems will stay the same and you will add *BSD systems at the end.

Additionally:
* we all agree people might not know how to check what version of Fedora they are using and that this should be added
* we also agree that you should be able to see only the system you are interested in and this issue is already placed in #188.

If you would like to introduce such changes we are happy to accept them ;)

sidrg:
@olasitarska sorry! i thought it is a online-tutorial.

with this arguments you are promoting fedora comparing with other operating systems in my point of view.

if popularity is ultimate, it is getting hard for django. (i can not find a statistic about usage share. i only found [something about jobs](http://www.indeed.com/jobtrends?q=django%2C+ruby+on+rails%2C+asp.net+mvc%2C+angularjs&l=) on the fly.)

once again: i thought it is a tutorial for online usage and people like me to learn django self-taught (alone or with other people around). if you text is used to having a workshop the situation is different.

PR #463, opened 2015-08-21
Update README.md - heroku python

czubik8805:
python-3.4.2 is not available on heroku

https://devcenter.heroku.com/articles/python-support#supported-runtimes


olasitarska:
Dzięki!

Issue #272, opened 2015-03-07
Design feedback: Style command prompt input/output differently


mattaustin:
Some feedback from the Django Girls Perth event, we had a few people who didn't realise the difference between the console/command prompt input and expected output (they continued to type in the 'output' text written in the docs).

Maybe we should style or separate the input/output text across the documentation?

danielquinn:
A lot of this can be fixed by enforcing the convention of prefixing "input" lines with `$`:

```bash
    $ python myscript.py
    This is the output of my script
```

It's pretty standard in forums and other documentation styles, so this would have the added benefit of making things less confusing when they're not in the tutorial.

aniav:
I think this is basically fixed as all prompt lines are now prefixed with `$` and additionally I have added a description above all code blocks explaining where this should be put: command-line, a specific file or PythonAnywhere to avoid confusion. I am going to close that for now, reopen it if you feel it's not enough :)

PR #473, opened 2015-09-14
Command prompt styling and other small improvements

infraredgirl:
Style the linux command prompt as `$` as opposed to including the current
directory name in the prompt.

Correct name of the python 3 package on Fedora 22 is `python3`.

Differentiate between Ubuntu/Debian/Fedora in Git installation
instructions.

### DIFF
--- a/en/deploy/install_git.md
+++ b/en/deploy/install_git.md
@@ -12,6 +12,17 @@ Download Git from [git-scm.com](http://git-scm.com/) and just follow the instruc
 
 If it isn't installed already, git should be available via your package manager, so try:
 
-    sudo apt-get install git
-    # or
-    sudo yum install git
+
+#### Debian or Ubuntu
+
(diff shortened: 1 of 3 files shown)


aniav:
Do you think it's better not to show the current directory in the Linux prompt examples?

infraredgirl:
I think it can be confusing for reasons described in issue https://github.com/DjangoGirls/tutorial/issues/272 .

Also keep in mind that the "style" of the command prompt is configurable and may or may not include current directory or number of other things. Different Linux distros style the prompt differently, for example.

The standard way of denoting the non-root prompt is the dollar sign `$` and I think it's good practice to promote this. The users also may see this elsewhere online and it's a good idea to get them used to it.

aniav:
If someone is able to configure the prompt then the changes commited here won't really be affecting them :)

I understand the reasoning now. Great input! Thank you :)))

Issue #474, opened 2015-09-16
Python 3.4.3 update instructions for Linux do not work


Winterflower:
Hi all!
The tutorial mentions that 

```
sudo apt-get install python3.4
````
can be used to update to python 3.4.
Ubuntu 14.04 comes with python 3.4.0, but trying to update using sudo apt-get install only gives

```
python3.4 is already the newest version.
```
The default version is python 3.4.0.
I'll look into updating the instructions, if that's ok.

alasdairnicol:
If Ubuntu 14.04 comes with Python 3.4.0, then the tutorial will run fine on it, and there's no need to upgrade. I've made a patch in #478. The alternative would be to add a note underneath the `apt-get install` command, saying something like

> if it says Python 3.4 is already the newest version, don't worry, it's fine.

alasdairnicol:
I think this can be closed, since #478 has been merged.

helenst:
Good call, thank you! :tulip::sparkles:

PR #478, opened 2015-09-19
Noted that there's no need to upgrade from 3.4.0 to 3.4.3

alasdairnicol:
Refs #474. Thanks @winterflower for the suggestion


helenst:
Thank you! :dolphin::rainbow: And thank you @Winterflower for reporting this!

PR #479, opened 2015-09-19
Change error name to reflect that shown to user

heldinz:
Minor change. Refer to error page as "TemplateDoesNotExist" instead of "Page not found", since that is the text shown to the user.


helenst:
Awesome, thank you! :sun_with_face::green_heart:

PR #482, opened 2015-09-25
migrate giving problem

taragurung:
Few system had the problem during my experience i dont know the exact reason behind but using the proposed code for such system worked fine.


olasitarska:
Hey @taragurung, thanks for this pull request!

However, I think `syncdb` has been deprecated with Django 1.7, and this tutorial uses Django 1.8. Please correct me if I'm wrong :)

taragurung:
The problem appeared in the latest version django too so we solved the problem with such solutions but Ok if it might be other cases thanks

PR #485, opened 2015-10-05
Adjust myvenv word

erichideki:


helenst:
Obrigada! :strawberry: :hibiscus:

PR #486, opened 2015-10-10
fix korean typos

sujinleeme:
edit 'syntax highlighting' code and remove artifacts in ko/README.me


rapoulson:
Hi :wave: @sujinleeme! Thanks for catching these, the `python` keyword determines how the code block gets processed but we have to make a few adjustments between Crowdin and Gitbook. Instead of removing it, what we want to do is format it like you see on line 19 of `ko/django_forms/README.md`, with three backticks that start at the left margin. We also close the block with three back ticks. Let me know if that's confusing, I'm happy to help :smile: Also, there are some merge artifacts in `ko/README.md` would you mind removing those before I update the tutorial with this PR? Thanks :fireworks:

sujinleeme:
Hi @rapoulson . Thanks for pointing out `python` keyword. I have added python code block again. Moreover, I have also removed artifacts in `ko/README.md` . Thanks :smile:

sujinleeme:
Hi @rapoulson Would you check again please?

rapoulson:
Looks good to me! :+1:

PR #487, opened 2015-10-12
grammar edit

wmoline:
Small grammar and clarity edits, especially around the use of commonly confused Latin abbreviations.


olasitarska:
Thanks @wmoline! :tulip:

PR #493, opened 2015-10-14
Pulled up note about logged out user error message

cemrich:
because this error could happen right after trying to save the form. Right now users will not go on reading the next section about form validation when this error is popping up and therefore will not see its solution.


helenst:
Yay! :tangerine::dolphin::tada:

PR #496, opened 2015-10-19
Use "function" instead of "method" where appropriate

domenkozar:


en/django_views/README.md
```diff
@@ -25,7 +25,7 @@ def post_list(request):
 ```
 
 
-As you can see, we created a method (`def`) called `post_list` that takes `request` and `return` a method `render` that will render (put together) our template `blog/post_list.html`.
+As you can see, we created a function (`def`) called `post_list` that takes `request` and `return` a function `render` that will render (put together) our template `blog/post_list.html`.
```
aniav: `post_list` is a method. No need to change that ;)

domenkozar: In Python terms, it's not a method: https://docs.python.org/2/tutorial/classes.html#method-objects

domenkozar: Correct me if I'm wrong, if it's a method, to what class does it belong?

aniav: Ah, sorrry! I messed this in my head with the class based views. You are totally right, thank you for correcting me!

domenkozar: :cake:

PR #502, opened 2015-10-27
Update README.md

sergio-garcia:
Updated the django version on the installation steps to use the same version (1.8.5 latest) on the instructions (1.7,1) and the command output (1.8).


sergio-garcia:
Replaced by #505.

PR #504, opened 2015-10-27
Updated django documentation link

sergio-garcia:
The tutorial version targets django 1.8 and all the tutorial should point to the correct version.


sergio-garcia:
Replace by #505

PR #505, opened 2015-10-27
Updated portuguese tutorial to point to Django 1.8.

sergio-garcia:
The portuguese tutorial use different versions of Django across files.
This patch updated all the references to current stable version 1.8.5.


helenst:
Wonderful, thank you! :hibiscus: :cake:

PR #506, opened 2015-10-30
prettyprint dict and fix method/function naming

domenkozar:
`participant` dict has a scrollbar on smaller screens

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -73,9 +73,9 @@ Nice, huh? To see your name in uppercase letters, simply type:
     >>> "Ola".upper()
     'OLA'
 
-You just used the `upper` __function__ on your string! A function (like `upper()`) is a sequence of instructions that Python has to perform on a given object (`"Ola"`) once you call it.
+You just used the `upper` __method__ on your string! A method (like `upper()`) is a sequence of instructions that Python has to perform on a given object (`"Ola"`) once you call it.
 
-If you want to know the number of letters contained in your name, there is a function for that too!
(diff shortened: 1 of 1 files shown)


en/python_introduction/README.md
```diff
@@ -252,7 +252,11 @@ This shows that you just created an empty dictionary. Hurray!
 
 Now, try writing the following command (try replacing your own information too):
 
-    >>> participant = {'name': 'Ola', 'country': 'Poland', 'favorite_numbers': [7, 42, 92]}
```
aniav: I am not sure if this doesn't bring problems for the user. There might be several problems with indentation when you force people to enter newlines. I would leave it as it was before...

domenkozar: But whitespace is so tied to Python we can't really escape it. We should rather explain it than hide it away. Later on if/else is introduces that uses whitespace for blocks

aniav: I believe we are intoruducing whitespace to them, but there is no need to make anything harder at this point. But I won't be arguing too much. Can someone taka a voice in this case please so we have some more inoput? :)

helenst: I'm inclined to think whitespace is better introduced at the `if` stage when we've started saving to files rather than doing it on the prompt - it seems like an extra complication at this point in the tutorial. I can see people getting unnecessarily frustrated, or trying to type in the dots.

domenkozar: Hmm, point taken.

aniav:
I merged it but changed the dictionary example to a oneliner. Thank you! :sparkles:

PR #532, opened 2015-11-12
Add note to create a branch before starting to edit in contributing README.md

glasnt:
The existing documentation referenced a `contributing` branch in the git output, but creating this branch was never detailed in the instructions. I've confirmed that GitBook doesn't appear to do this for you, so I've added it under the `git clone` command. 

I tried to keep the new copy in line with the current tone, but I'm more than happy to accept changes to make it match better.


aniav:
Good catch! :tada:

Issue #500, opened 2015-10-27
New syntax introduced and unexpained


knyghty:
Someone came into #django today asking what the difference between `{% endblock %}` and `{% endblock content %}` is, as the latter is introduced in the Template Extending section, but isn't explained at all (as far as I can see).

To avoid confusion, I'd suggest either removing it or explaining why it's used here but not elsewhere.

Log: https://botbot.me/freenode/django/2015-10-27/?msg=52833480&page=10

olasitarska:
:+1: for removing it! :)

olasitarska:
Thanks for reporting @knyghty! :tulip:

PR #534, opened 2015-11-13
template_extending: Name closed block in {% endblock %}

encukou:
{% endblock content %} was mentioned in the text, but the code had just {% endblock %}.
AFAIK it's good practice to name the closed block.

### DIFF
--- a/en/template_extending/README.md
+++ b/en/template_extending/README.md
@@ -63,7 +63,7 @@ Then in `base.html`, replace your whole `<body>` (everything between `<body>` an
         <div class="row">
             <div class="col-md-8">
             {% block content %}
-            {% endblock %}
+            {% endblock content %}
             </div>
         </div>
     </div>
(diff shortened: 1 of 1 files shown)


phalt:
Good spot @encukou ! :grin: 

I agree that explicitly naming the end block is good practice.

helenst:
Hmm... we just got rid of this in #500 (although the PR for that left that mention in the text) Not sure what the right thing is here - agreed about it being good practice. Maybe if the original complaint was of new syntax being introduced, we should make sure the closed block is named everywhere - then it won't be new syntax. :)

olasitarska:
@helenst, good suggestion! @encukou would you have some time to correct that in the whole tutorial -- just english version?

olasitarska:
Closing this now. @encukou please re-open once you've got a chance to update this pr :)

PR #535, opened 2015-11-13
Alphabetic order names and insert list names

erichideki:


olasitarska:
Thanks @erichideki! :)

PR #536, opened 2015-11-14
Done Django Girls Tutuorial Chinese Simplified version

jeffkit:
:)


tricoder42:
Will be fixed in #547

PR #537, opened 2015-11-14
Include note for Windows 10 pip install freeze

vekerdyb:


keimlink:
Wouldn't this install Django systemwide instead into the virtual environment?

vekerdyb:
You are right, it should be `python -m pip install django==1.8`. 
I will recreate the pull request.

PR #544, opened 2015-11-20
Update Templates

shacon:
Specifying to use the right Django Admin to add data for the Pythonanywhere site.  It took me a little while to figure this one out so thought it might be helpful for other people.


olasitarska:
Thanks @shacon! :palm_tree:

PR #548, opened 2015-11-22
Fix code example.

flash42:


olasitarska:
This has been fixed in #550. Thanks for subbmitting this though!

PR #552, opened 2015-11-28
Add a hint how to solve a common 'ImportError' problem in Python Anyw…

akloster:
At DjangoGirls Cologne we had a problem with several participants when they deployed their code. In the error logs there was an "ImportError" concerning "mysite".

The problem was that the string appended to the Python path was typed in wrongly.


olasitarska:
Thanks @akloster! :sparkles:

PR #553, opened 2015-11-29
Clarify the command line prompt

akloster:
I've had several issues where participants tried to input the prompt
itself. Therefore I suggest to make this point a bit more clearly. Also
In some places (I'll check later where) the tutorials (and their
translations) include much more of the prompt, which adds to the
confusion.


olasitarska:
@akloster perfect, really useful! Thank you.

Issue #512, opened 2015-11-07
choose the "Add python.exe to the Path" option is ambiguous


mulka:
In the Windows installation it doesn't say which option to select under "Add python.exe to the Path". Will be installed or install entire feature. I'm assuming install entire feature is the better option, but not sure what the difference would be.

PR #555, opened 2015-12-01
Updated Python install instructions for Windows

jmichelsen:
To be more specific about adding Python to your Windows path. Issue #512


olasitarska:
Perfect! Thanks @jmichelsen :tulip:

PR #557, opened 2015-12-02
use the url name for the redirect

pkimber:


pkimber:
Sorry to create the pull request.  I want to review this change a bit more.

pkimber:
In the `en/django_urls/README.md` we say *We will be using the named URLs later in the project so it is important to name each URL in the app*.  I thought it might be easier to use the name of the URL in the `view`?

olasitarska:
@pkimber good suggestion! Thank you :panda_face:

PR #558, opened 2015-12-03
Update README.md

dauricchio:
Fix 'Python' word


olasitarska:
Thank you @dauricchio!

PR #560, opened 2015-12-03
fix `TemplateSyntaxError at /post/2/edit/` `Invalid block tag: 'raw', expected 'endblock'`

F075kK:
fix by removing tags raw and endraw from fenced html block


olasitarska:
Thanks!

PR #561, opened 2015-12-03
Add security to edit button also.

purplediane:
Currently the tutorial only puts security around the add post functionality. Added instructions to put the same security around the edit button that shows up in the detail view. Also added a tiny bit of explanation of glyphicon for those like me who wonder where that came from.


olasitarska:
That's great, thank you Diane!

PR #565, opened 2015-12-04
Update README.md

cgl:
django-admin command should be written with a dash not a space.
Also . destination folder requires a space after project name.
Fixed


olasitarska:
Thank you!

PR #571, opened 2015-12-05
Move color-coding commentary

purplediane:
The description of color-coding is before saving of the file. Color
coding doesn’t happen until the file is saved and the editor knows what
file type it is.


olasitarska:
Thank you! :sparkling_heart:

PR #579, opened 2015-12-12
Fix error in string escaping

yakky:


yakky:
Closing this #581 is so much better

PR #582, opened 2015-12-12
Changed myenv to myvenv

patrick91:


patjouk:
Thanks again :smile_cat:

PR #585, opened 2015-12-12
Fixed wrong output for len command

patrick91:
was a bit confusing since lotteria was exactly 8 letters long :)


patrick91:
Duplicate, sorry ^^

PR #591, opened 2015-12-12
Removed line that generates misunderstanding

daivavt:
The text "Scegli la versione "HTTPS", copialo, e incollalo nel terminale a breve" does not make much sense in italian.
It can be removed since on the next lines the concept is explained well enough.


patjouk:
Thanks :cake:

PR #593, opened 2015-12-12
Fix code for Spanish version

lhuriguen:
The command block for Spanish is missing new lines and the collectstatic.


patjouk:
Thanks :wink:

PR #594, opened 2015-12-12
[IT] Fixed python anywhwere WSGI conf

nemesisdesign:
The first line of the WSGI configuration causes an error.
It's not clear if this is intended or not, it looks like a mistake in the tutorial.


patjouk:
fixed by #590

PR #600, opened 2015-12-15
Fixed instruction that was confusing

elifloresch:
'ls/dir' was a confusing for some of the girls on the workshops where I have participated. I think that separating ls and dir and indicating the operating system is a little clearer.
Thanks! :-)


olasitarska:
Thanks! :)

PR #601, opened 2015-12-15
Fix indentation , and remove python instruction on top of code.

jeffkit:
Fix indentation , and remove python instruction on top of code.


olasitarska:
Thanks! :)

PR #603, opened 2015-12-16
update latest version

fbchow:


olasitarska:
@fbchow have you also walked through the whole tutorial with a new Django version? there could be some differences that we're not aware of :)

PR #608, opened 2015-12-18
README.md modified

Satender052:


olasitarska:
I'm sorry, but this actually doesn't bring any value and breaks translation strings :(

PR #612, opened 2015-12-25
Added possible error message for Windows 10 users

lwyso:
Windows 10 computer might receive message “execution of scripts is
disabled on this system.” message to prevent users to continue or work
in venv. To correct this error, the user needs to set execution policy
to apply to All.

### DIFF
--- a/en/django_installation/README.md
+++ b/en/django_installation/README.md
@@ -58,7 +58,16 @@ The command above will create a directory called `myvenv` (or whatever name you 
 
     C:\Users\Name\djangogirls> myvenv\Scripts\activate
 
-on Windows, or:
+on Windows, 
+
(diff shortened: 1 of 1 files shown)


olasitarska:
I am not sure if it's safe to recommend that to users, as I'm not a Windows 10 user myself. Would love to get a review of that from some Windows users!

lwyso:
Agree, it would be a good idea for others to comment on this too. Apart from setting 'RemoteSigned', you could set 'Unrestricted' but this adds more risk if the user has not got enough knowledge about computers.

ProProgrammer:
I assume this is when people are using PowerShell to execute commands instead of command prompt (cmd). I use windows 10 on a day to day basis and so far haven't faced any issues personally. Although I do use command prompt all the time and not PowerShell.

Environment:
Windows 10 Pro Version 10.0.10586 Build 10586

aniav:
I have read the information on microsoft page and it seems the scripts downloaded from the internet still need to be signed but you can freely run scripts written locally, so it's fairly secure. I think we can merge that :)

PR #614, opened 2015-12-27
Change mvenv for myvenv since that's the name of the virtualenv created

jstitch:


helenst:
Nice, thank you! :sunflower::blue_heart:

PR #622, opened 2016-01-03
Update python install instructions for Fedora

abonas:
On Fedora 20 and Fedora 21, doing 'yum install python3.4' as currently documented, 
returns 'No package python3.4 available.'.
Doing 'yum install python3' installs python3.4.x


olasitarska:
Thanks! :dolphin:

PR #623, opened 2016-01-06
Use latest version of ga plugin

SamyPesse:
This should fix the analytics not being recorded in Google Analytics (https://github.com/GitbookIO/plugin-ga/commits/master).

CC @patjouk


patjouk:
Thanks :wink: !

PR #627, opened 2016-01-07
Edit to WSGI configuration file

bmsleight:
As per https://github.com/evansd/whitenoise/issues/31

Error from PythonAnywhere 
https://www.pythonanywhere.com/user/<USER NAME>/files/var/log/<USER NAME>.pythonanywhere.com.error.log
2016-01-07 13:17:08,392 :    from whitenoise.django import DjangoWhiteNoise
2016-01-07 13:17:08,392 :ImportError: No module named 'whitenoise'


olasitarska:
I think this is specific to your application (`prompt`), not general?

PR #629, opened 2016-01-09
es - `pat on the back`

chris-ramon:
#### Details.

- `Pat on the back` expression was translated to Spanish exactly word-by-word.
- IMHO `Palmada en la espalda` is not quite common to say, a better generic translation would be `Buen Trabajo` meaning `Good Job` in English.
  - Ref: http://www.wordreference.com/es/translation.asp?tranword=pat#enes:120275


olasitarska:
Thanks @chris-ramon! :snail:

PR #634, opened 2016-01-23
Coding is for girls links

asendecka:
Link to videos that covers bits of Django Girls tutorial added in correct places in tutorial. It's still work in progress, but it might be very useful for readers who are following tutorial at home. 

:heart: :palm_tree: :star:


phalt:
I like the idea :+1:

PR #642, opened 2016-02-01
The offer sounded unnatural, as the automatic translation.

gil9red:
The offer sounded unnatural, as the automatic translation.
Предложение звучало неестественно, как автоматический перевод.


alexanderad:
@olasitarska this is a re-iteration of #632 and is good to go from my point of view :+1: 
#632 can be closed after that.

olasitarska:
Thanks! :two_hearts:

PR #659, opened 2016-02-04
Update README.md

nampi:


patjouk:
Thanks a lot for all those fix. Next time, can you make only one PR? We're a small team and it would make our life much easier :wink:

nampi:
Thanks! I will certainly take into account your remark.

PR #663, opened 2016-02-06
Be more specific about what to replace and how

craigsdennis:
I saw two separate attendees type <their_name> including the greater than and less than.  Explicit!


olasitarska:
Thanks!

PR #664, opened 2016-02-07
Replace multiple occurrences of django 1.8 with 1.9 

alexanderad:
Refer latest version of django used in the tutorial.


olasitarska:
Thanks!

Issue #641, opened 2016-01-31
Simpler static assets


glenjamin:
Hello! I'm due to coach at a workshop in a few weeks, so I've been running through the tutorial myself first. It's great!

One of the things that did strike me as a bit awkward was the static asset serving stuff - having to introduce `whitenoise` as part of the deployment, and also needing to use `collectstatic` every time.

I had a poke around the django docs to investigate this further and found https://docs.djangoproject.com/en/1.9/howto/static-files/#serving-static-files-during-development 

I reckon this would simplify things a little, and reduce the number of steps which need to be performed.

I gave this a go with the following: https://github.com/glenjamin/djangogirls-tutorial/commit/0ede1209841f41ea5b1f54f09fc0c74312457f8e

This allowed me to reduce my pythonanywhere `wsgi` to the following, and stop using `collectstatic`

```python
import sys

path = '/home/glenjamin/djangogirls-tutorial'
if path not in sys.path:
    sys.path.append(path)

from mysite.wsgi import application
``` 

Is there a reason I've missed for not doing it this way already? Would you like me to send a PR for this?

olasitarska:
The `whitenoise` could be a leftover after we migrated from Heroku to PythonAnywhere...

@bmispelon @hjwp, can you advise here? I'm not entirely sure :)

glenjamin:
Hrm, I've now realised that this doesn't directly remove the need for `collectstatic`, it just removes the need for whitenoise.

glenjamin:
Ok, I did a bit more poking around and found that this approach enables the same static-file serving approach as the development server uses when deployed to pythonanywhere

https://github.com/glenjamin/djangogirls-tutorial/commit/07055a20fff5368a0adc407c1579b88fc80072d5


The docs have some scary warnings about "insecure" and "grossly inefficient", however this stack overflow post elaborates a bit: http://stackoverflow.com/questions/31097333/why-is-serving-static-files-insecure

My opinion is that it's an appropriate tradeoff for this application.


For a "real" application, personally I'd prefer to have my app serve up the static files anyway - and include caching headers so the next layer up can make things more efficient. I find Django core's hard-line on "handle assets yourself some other way" a little odd :)

hjwp:
@glenjamin that looks like it could work just fine.

[...]

PR #665, opened 2016-02-07
Remove need for whitenoise and collectstatic

glenjamin:
Enabling the same WSGI middleware that is used by runserver allows us to simplify the static files process, and remove a bunch of steps.

As discussed in #641

### DIFF
--- a/en/css/README.md
+++ b/en/css/README.md
@@ -38,7 +38,7 @@ Finally we will take a closer look at these things we've been calling __static f
 
 ### Where to put static files for Django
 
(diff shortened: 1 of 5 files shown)


olasitarska:
@glenjamin This looks good! Have you had a chance to actually test it with the rest of the Django Girls blog? Just making sure this approach actually works as described at PythonAnywhere, etc :)

glenjamin:
Yup, it's running on https://github.com/glenjamin/djangogirls-tutorial

If someone else is able to verify then that would be good too!

phalt:
@olasitarska ran through this with @glenjamin at DjangoGirls Sheffield to verify. Looks grand! :+1:

olasitarska:
Awesome! Thanks everyone :)

PR #666, opened 2016-02-09
Changes to Models and Deploy Chapter

wade-christensen:
In the Models chapter, I noticed participants of the PDX event this previous weekend seeing their posts with the title "Post object." It turned out to be an indentation error. I wanted to make it more clear that the methods needed to be indented inside the class to help future participants.

In the Deploy chapter, it seemed a little confusing to not explain that we would be moving back to our local environment to make changes. One participant made manual changes to her GitHub repo, which was confusing for her when she couldn't push without pulling changes down locally first. I thought this note might help.


olasitarska:
Thanks Wade, really helpful! Thanks for taking the time to include that.

wade-christensen:
Happy to contribute :)

PR #667, opened 2016-02-12
Change python version in a link for OS X

kiote:
For some reason everything here is about 3.4.3, but link was pointed to 3.4.2

I didn't change 59th line, but somehow it marked as changed as well :whale:


olasitarska:
Thanks! :) :100:

PR #670, opened 2016-02-17
Update Readme.md (french)

qlynq:
Line 42 : we need to create a folder (dossier) called "static", not a file (fichier).


phalt:
:+1: Looks like my French studies are paying off!

patjouk:
Thanks a lot! :cake:

PR #678, opened 2016-02-20
Make missing view function error more accurate for 1.9

nimasmi:
Looks like initialisation now tries to import view functions at server start, so you don't get to the DEBUG = True style error page any more. I've changed this, but kept the comments about friendly error pages elsewhere.

### DIFF
--- a/en/django_urls/README.md
+++ b/en/django_urls/README.md
@@ -111,12 +111,12 @@ As you can see, we're now assigning a `view` called `post_list` to `^$` URL. Thi
 
 The last part `name='post_list'` is the name of the URL that will be used to identify the view. This can be the same as the name of the view but it can also be something completely different. We will be using the named URLs later in the project so it is important to name each URL in the app. We should also try to keep the names of URLs unique and easy to remember.
 
-Everything all right? Open http://127.0.0.1:8000/ in your browser to see the result.
+If you try to visit http:127.0.0.1:8000/ now, then you'll find some sort of 'web page not available' message. Take a look at your server console window.
 
-![Error](images/error1.png)
+```bash
+AttributeError: 'module' object has no attribute 'post_list'
+```
 
-There is no "It works" anymore, huh? Don't worry, it's just an error page, nothing to be scared of! They're actually pretty useful:
-
-You can read that there is __no attribute 'post_list'__. Is *post_list* reminding you of anything? This is what we called our view! This means that everything is in place but we just haven't created our *view* yet. No worries, we will get there.
+It's not running anymore, huh? It's telling you that there is __no attribute 'post_list'__. That's the name of the *view* that Django is trying to find and use, but we haven't created it yet. No worries, we will get there.
 
 > If you want to know more about Django URLconfs, look at the official documentation: https://docs.djangoproject.com/en/1.9/topics/http/urls/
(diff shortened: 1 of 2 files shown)


phalt:
Been encountering the same issues today. I think we should mention to check the terminal / command prompt output to see the error instead?

phalt:
@nimasmi I've been seeing this with my work group today as well :+1: Should we mention in the second page (django_views/README.md) to check the command prompt / terminal?

nimasmi:
Thanks @phalt. But see my commit—that's exactly where I've moved it to. ;-)

nimasmi:
Oh, or actually have I misunderstood your suggestion? Not sure why we need to suggest checking the terminal in https://github.com/DjangoGirls/tutorial/blob/master/en/django_views/README.md, as the error will still be in the browser here.

en/django_urls/README.md
```diff
@@ -111,12 +111,12 @@ As you can see, we're now assigning a `view` called `post_list` to `^$` URL. Thi
 
 The last part `name='post_list'` is the name of the URL that will be used to identify the view. This can be the same as the name of the view but it can also be something completely different. We will be using the named URLs later in the project so it is important to name each URL in the app. We should also try to keep the names of URLs unique and easy to remember.
 
-Everything all right? Open http://127.0.0.1:8000/ in your browser to see the result.
+If you try to visit http:127.0.0.1:8000/ now, then you'll find some sort of 'web page not available' message. Take a look at your server console window.
```
helenst: Maybe say something like "That means the server has stopped running." - it might help them to make that connection in other situations and look to the console for help.

en/django_views/README.md
```diff
@@ -33,6 +33,6 @@ Another error! Read what's going on now:
 
 ![Error](images/error.png)
 
-This one is easy: *TemplateDoesNotExist*. Let's fix this bug and create a template in the next chapter!
+There is no "It works" anymore, huh? Don't worry, it's just an error page, nothing to be scared of! They're actually pretty useful.  You can read that *TemplateDoesNotExist*. This one is easy: Let's fix this bug and create a template in the next chapter!
```
helenst: Rather than referring back to "It works" (as other stuff happened since that), we could mention that this means the server is running again now.

helenst:
I checked and there's no need to look at the terminal in django_views section. However, there is another place this happens - we encountered this twice during the workshop - and I think it's when we start referring to `post_detail` later on.

en/django_urls/README.md
```diff
@@ -111,12 +111,12 @@ As you can see, we're now assigning a `view` called `post_list` to `^$` URL. Thi
...
 
-![Error](images/error1.png)
```
helenst: We could remove this file from the repo.

helenst:
Looks like this has got out of date because somebody updated the file and added a new screenshot of the terminal - but there are still some good changes in here so I'll merge manually :-)

helenst:
Thank you for your contribution! :tulip::penguin:

nimasmi:
Thanks for merging! You're right that I did let this stagnate, because I got stalled trying to create a screenshot with an appropriate favicon!

I have pushed some additional changes now to #717. Feel free to disregard if you think the current state of master is good enough.

PR #685, opened 2016-02-26
Django version

crazcalm:
in the django installation section, the instruction tell the user to install django 1.9. However, the deploy section does not specify which version of django to install. This change is to ensure that the local version of django and the deployed version of django are the same.


phalt:
Great suggestion, thanks @crazcalm ! :grin:

PR #688, opened 2016-02-26
Major and minor fixes on legibilty

jstitch:
from code not displaying good to typos and translation corrections


jstitch:
nope, need to merge with upstream first

PR #695, opened 2016-02-27
clarify AttributeError location

sternb0t:
Django 1.9 validation may kick in here, and cause the `AttributeError module has no attribute post_list` to appear in the command line console rather than on the site with the debug view.


aniav:
Looks great, thenk you! :100:

sternb0t:
You're welcome, and thanks from Django NYC! :statue_of_liberty:

PR #696, opened 2016-02-27
Update README.md

chrissyzic:
We were all confused at the end of this step because it was the first time the local server 127.0.0.1:8000 didn't match username.pythonanywhere.com, since we made changes to the local database that were not synced with the database stored on Python Anywhere. To prevent the same panic in others, we thought it would be helpful to add a note that the two sites may not match - and that's OK!

### DIFF
--- a/en/django_templates/README.md
+++ b/en/django_templates/README.md
@@ -86,7 +86,7 @@ $ git pull
 [...]
 ```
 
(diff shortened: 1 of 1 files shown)


aniav:
Thenk you! :sparkles:

PR #697, opened 2016-02-27
changed the error message of no attribute post_list

danagilliann:
This is my first open source pull request. Let me know if there's anything else you need.
The error message on the tutorial was not the same as the actual error. I replaced the error1 image and changed some of the text. I also removed the float right for the date class in CSS. It was misleading to the user and did not match the image.


olasitarska:
Thanks! Very appreciated.

Congrats on your first open source contribution! :tada:

PR #700, opened 2016-02-28
Remove more unneeded venv activations

saghul:


helenst:
Thank you! :panda_face::snowflake:

PR #701, opened 2016-02-28
Simplify regex

saghul:
Use \d instead of [0-9], which is explained in chapter 13.


phalt:
@saghul hey! thanks for the contribution. I agree that this easier to understand. I'll :+1: but I'd prefer to get a secondary :+1: just in case there is a reason we picked the first method originally :)

olasitarska:
Looks good! Thanks :)

PR #702, opened 2016-02-28
[pt] added missing published_date statement to post_new and post_edit

gustavomazevedo:
Also fixed 'mport' to 'import' in line 97


olasitarska:
Thanks!

PR #703, opened 2016-03-05
Add timezone import and associated text

gyermolenko:


helenst:
Thank you for contributing! :sparkles:

The timezone import already goes in when the `post_list` view is added in an earlier chapter (http://tutorial.djangogirls.org/en/dynamic_data_in_templates/index.html#queryset)

PR #704, opened 2016-03-06
Install latest release of Django 1.9

beli-sk:
Install latest release of Django 1.9 (1.9.4 as of the time of this writing) instead of the first 1.9 release. The original suggested install method may lack critical security fixes added in later patches.

This syntax should work even if there weren't any point releases yet.

Tested on Linux under Bash, please test on Mac and Windows, if the quoting style is portable.

### DIFF
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -167,7 +167,7 @@ Installing setuptools, pip...done.
 
 $ source myvenv/bin/activate
 
-(myvenv) $  pip install "django==1.9.*"
+(myvenv) $  pip install django~=1.9.0
 Collecting django
 [...]
 Successfully installed django-1.9
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -167,7 +167,7 @@ Installing setuptools, pip...done.
 
 $ source myvenv/bin/activate
 
-(myvenv) $  pip install django==1.9
+(myvenv) $  pip install "django==1.9.*"
 Collecting django
 [...]
 Successfully installed django-1.9
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -90,9 +90,9 @@ OK, we have all important dependencies in place. We can finally install Django!
 
 ## Installing Django
 
-Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install "django==1.9.*"` (note that we use a double equal sign: `==`).
+Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install django~=1.9.0` (note that we use a double equal sign: `==`).
 
-    (myvenv) ~$ pip install "django==1.9.*"
+    (myvenv) ~$ pip install django~=1.9.0
     Downloading/unpacking django==1.9
     Installing collected packages: django
     Successfully installed django
@@ -104,7 +104,7 @@ on Windows
 on Windows 8 and Windows 10
 > Your command line might freeze after when you try to install Django. If this happens, instead of the above command use:
 
->     C:\Users\Name\djangogirls> python -m pip install "django==1.9.*"
+>     C:\Users\Name\djangogirls> python -m pip install django~=1.9.0
 
 on Linux
 > If you get an error when calling pip on Ubuntu 12.04 please run `python -m pip install -U --force-reinstall pip` to fix the pip installation in the virtualenv.
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -90,9 +90,9 @@ OK, we have all important dependencies in place. We can finally install Django!
 
 ## Installing Django
 
-Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install django==1.9` (note that we use a double equal sign: `==`).
+Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install "django==1.9.*"` (note that we use a double equal sign: `==`).
 
-    (myvenv) ~$ pip install django==1.9
+    (myvenv) ~$ pip install "django==1.9.*"
     Downloading/unpacking django==1.9
     Installing collected packages: django
     Successfully installed django
@@ -104,7 +104,7 @@ on Windows
 on Windows 8 and Windows 10
 > Your command line might freeze after when you try to install Django. If this happens, instead of the above command use:
 
->     C:\Users\Name\djangogirls> python -m pip install django==1.9
+>     C:\Users\Name\djangogirls> python -m pip install "django==1.9.*"
 
 on Linux
 > If you get an error when calling pip on Ubuntu 12.04 please run `python -m pip install -U --force-reinstall pip` to fix the pip installation in the virtualenv.


en/deploy/README.md
```diff
@@ -167,7 +167,7 @@ Installing setuptools, pip...done.
 
 $ source myvenv/bin/activate
 
-(myvenv) $  pip install django==1.9
+(myvenv) $  pip install "django==1.9.*"
```
alexanderad: What about setting it to `django>=1.9` instead?

beli-sk:
@alexanderad, I wouldn't just do >=1.9 as that will pull 1.10 or whatever comes after 1.9 in the future and that version may be incompatible with this tutorial. I'd rather have the version change under control and manually change 1.9 to a higher minor version when the tutorial is checked and updated accordingly.

alexanderad:
@beli-sk this makes sense, thanks for the explanation.

phalt:
@alexanderad Hey, thanks for the contribution. My two cents: you should always pin the exact version of a package that you want, as it can lead to ambiguity. I think this is something we talk about during the workshops. Furthermore, Django doesn't run on a traditional release cycle, so you can't guarantee that point releases won't break APIs.

I'd recommend making this explicitly `1.9.4` - it's not hard to change it in the future if we need to :smiley: 

Thank you!

beli-sk:
The *patch releases* (the third number) is incremented only for bugfix and security releases and are 100% compatible whenever possible, as stated here: https://docs.djangoproject.com/en/1.9/internals/release-process/

So I think it's totally safe, I'd consider it even mandatory, to pull in latest 1.9 patch release. No functionality will be broken and the novice users will be safer from bugs and security vulnerabilities that may arise, without any delay caused by manually incrementing versions in the tutorial (supposing there is no editor watching full-time for bugs and vulnerabilities). And, relying on 1.9 for four months and 3 patch releases after it's release date is an example of that.

And there *were* security issues, as seen for example here: https://docs.djangoproject.com/en/1.9/releases/1.9.3/

phalt:
You make a good point :)

ProProgrammer:
@beli-sk +1 for ensuring compatibility with the tutorial.

ProProgrammer:
By the way, quoting from #714 can we use `django>=1.9,<1.10` or `django<1.10`

remram44:
[PEP 440](https://www.python.org/dev/peps/pep-0440/) mentions that the correct syntax is `django~=1.9`, which conveniently doesn't require escaping, but this doesn't work on my machine so I assume it needs a recent version of pip.

ProProgrammer:
Verified with pip 8.1.0, `pip install django~=1.9` installs `Django-1.9.4-py2.py3-none-any.whl`

phalt:
Is a specific version of pip bundled along with Py3?

remram44:
You mean, the PY3 installers you can get off python.org?

ProProgrammer:
Yeah I think pip 7.1.0 is bundled with Python 3.4.4

phalt:
Awesome, thanks for finding out. 

I think we should support minor versions of Django 1.9.*, as @beli-sk mentioned, it is useful to have minor security patches. So if we can get `django~=1.9` to install nicely with the version of pip that is bundled with Py3 (taking into account all the OS we support) and make sure there is no affect to the tutorial (there shouldn't be, but we should definitely check), then we should ship that :)


@ProProgrammer @remram44 @beli-sk sound like a plan? :smile:

ProProgrammer:
@phalt yup. definitely makes sense. I will confirm on this thread as and when I start the tutorial with django 1.9.4 in order to confirm there's no negative impact on tutorial. Meanwhile, if someone else starts before I do, please inform.

en/deploy/README.md
aniav: I would also opt on the setting proposed by @alexanderad. Using * in pip packages is not very popular and I actually had to check if something like that works... :)

beli-sk:
I would just like to point out, that if you would like to use the _compatible release_ operator `~=`, you want to use `django~=1.9.4`.

`~=1.9` is equivalent to `==1.*` and that is NOT what you want.

More details here: https://www.python.org/dev/peps/pep-0440/#compatible-release

Though I'm not sure if this is sufficiently supported in pip, I'm not sure if you can test all the possible versions you may encounter. I bet many still supported Linux distributions that come with Python 3 are still shipping pip without this functionality. I'm for `==1.9.*`, that has the best compatibility.

remram44:
@phalt Linux distributions won't have a version of pip recent enough (next Ubuntu release will)

@beli-sk Oops, yeah should be `~=1.9.0` or similar.

ProProgrammer:
Thanks @beli-sk for pointing that out. The default exact pip version with Python 3.4.4 is 7.1.2 and using that I am able to use the `django~=1.9.0` syntax
Here are my findings with the default pip on python 3.4.4 (pip 7.1.2):
* `pip install django~=1.9.0` install Django 1.9.4
* `pip install django~=1.8.0` install Django 1.8.11
* `pip install django~=1.7.0` install Django 1.7.11
* `pip install django~=1.5.0` install Django 1.5.12

However, if I use django~=1.8.*, it fails with error:
```
pip._vendor.pkg_resources.RequirementParseError: Expected version spec in django~='1.8.*' at ~='1.8.*'
```

I think we are good to go with `.0` at the end instead of `.*` to keep things simple.

beli-sk:
@ProProgrammer you either use `~=1.9.0` or `==1.9.*`. Using `~=` with a wildcard is not supported.

ProProgrammer:
@beli-sk Oh alright, I verified now, Similar behavior observed for wildcard with `==`. We are good to go with either of them (wildcard or `~=`)
Thanks for pointing that out!

helenst:
It seems like either of those options would be fine but people are slightly preferring `django~=1.9.0` and it sounds like the pip version compatibility is fine.

@beli-sk please could you update your PR - it would be really great to get this merged! :)

ProProgrammer:
Finished deploying django app according to the tutorial with Django 1.9.4 installed using `django~=1.9.0` No issues observed in tutorial

beli-sk:
Updated PR to use `~=` as that way seems to be more popular here.

phalt:
Amazing. Thank you so much @beli-sk and @ProProgrammer! :heart:

en/django_installation/instructions.md
```diff
@@ -90,9 +90,9 @@ OK, we have all important dependencies in place. We can finally install Django!
 
 ## Installing Django
 
-Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install django==1.9` (note that we use a double equal sign: `==`).
+Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install django~=1.9.0` (note that we use a double equal sign: `==`).
```
remram44: How is that "a double equal sign"? See #714 for additional changes

beli-sk:
@remram44, well now it's obviously not "a double equal sign". As this is already merged I suggest you rebase your PR on top of this, i.e. change the note in parentheses and have that merged along with your french update. Thank you for pointing this out.

PR #714, opened 2016-03-15
Update Django package specifier

remram44:
`==1.9` will install the 1.9 version exactly, ignoring minor releases containing bugfixes and security patches. This is very bad practice.

The correct specifier is probably `Django>=1.9,<1.10` but since this is kind of complicated I went with `<1.10`. Unfortunately then the shell needs escaping.

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -90,10 +90,10 @@ OK, we have all important dependencies in place. We can finally install Django!
 
 ## Installing Django
 
-Now that you have your `virtualenv` started, you can install Django using `pip`. In the console, run `pip install django~=1.9.0` (note that we use a double equal sign: `==`).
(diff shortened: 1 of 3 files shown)


dscloudlending:
We have to review the impact of this change on the tutorial steps since the tutorial specifically addresses all steps required for django 1.9 to work including hosting the app on pythonanywhere

ProProgrammer:
We have to review the impact of this change on the tutorial steps since the tutorial specifically addresses all steps required for django 1.9 to work including hosting the app on pythonanywhere

helenst:
See #704 where there is already some discussion on this.

remram44:
#704 didn't quite cut it, reopened :wink:

helenst:
closing, since #704 is fine.

PR #716, opened 2016-03-21
Remove python to models.py (error)

betdream:
Remove  `python`  to models.py  (error)

### DIFF
--- a/tr/django_models/README.md
+++ b/tr/django_models/README.md
@@ -110,6 +110,7 @@ INSTALLED_APPS = (
 Şimdi `blog/models.py` dosyasını açalım ve içindeki her şeyi silip şu kodu yazalım:
 
 ```
+python
 from django.db import models
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


tr/django_models/README.md
```diff
@@ -110,7 +110,6 @@ INSTALLED_APPS = (
 Şimdi `blog/models.py` dosyasını açalım ve içindeki her şeyi silip şu kodu yazalım:
 
 ```
-python
 from django.db import models
```
helenst: The word python is meant to be there but should be on the end of the previous line so the syntax highlighter knows this is python.

so you will have

```
```python
```

helenst:
Hi @betdream thank you for your contribution! If you can move that "python" up to the line with the backticks rather than removing it, then we can get this merged! :sparkling_heart:

aniav:
Hi @betdream, what Helen was suggesting is moving python to the same line as the ``` signs, now there are no changes to the original code :)

betdream:
Thanks, i was wrong :) no error

PR #718, opened 2016-03-23
warn at the earliest opportunity about issues caused by accented chars

abeld:
See mentions of accented characters in later part of the tutorial:
since having accented chars in the directory name will cause problems
later on on windows, we should make sure students won't need to
start again with a good directory name.

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -12,7 +12,7 @@ Before we install Django we will get you to install an extremely useful tool to 
 
 So, let's create a **virtual environment** (also called a *virtualenv*). Virtualenv will isolate your Python/Django setup on a per-project basis. This means that any changes you make to one website won't affect any others you're also developing. Neat, right?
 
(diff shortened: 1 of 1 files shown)


en/django_installation/instructions.md
```diff
@@ -12,7 +12,7 @@ Before we install Django we will get you to install an extremely useful tool to
 
 So, let's create a **virtual environment** (also called a *virtualenv*). Virtualenv will isolate your Python/Django setup on a per-project basis. This means that any changes you make to one website won't affect any others you're also developing. Neat, right?
 
-All you need to do is find a directory in which you want to create the `virtualenv`; your home directory, for example. On Windows it might look like `C:\Users\Name\` (where `Name` is the name of your login).
+All you need to do is find a directory in which you want to create the `virtualenv`; your home directory, for example. On Windows it might look like `C:\Users\Name\` (where `Name` is the name of your login). One Windows, make sure that this directory does not contain accented or special characters; if your username contains accented characters, use a different directory, for example `C:\djangogirls`.
```
helenst: "One Windows" -> "On Windows".

Maybe this bit could be a separate note?

PR #719, opened 2016-03-23
Remove advice to move virtualenv directory

abeld:
as renaming or moving a virtualenv directory will break it.
Reword warning to delete & create new virtualenv in new directory.
(Suggested order is to create new & then delete the old one, just in
case one needs to refer to the old one for any reason.)

(see "Normally environments are tied to a specific path. That means
that you cannot move an environment around or copy it to another computer."
in https://virtualenv.pypa.io/en/latest/userguide.html#making-environments-relocatable)


helenst:
Good call - thanks! :dolphin:

PR #720, opened 2016-03-24
Invalid syntax on line 123 (django_orm chapter - FR version)

matgerard:
Invalid syntax on line 123 in chapter django_orm (FR version)
Should be done on two lines to avoid "SyntaxError: invalid syntax"


helenst:
Thank you! :tulip:

PR #745, opened 2016-04-22
Add $ before commands in Virtual Environment installation

Geekfish:
People don't always understand they need to type things if the `$` sign is missing. This should make it slightly clearer and consistent with the rest.


helenst:
Thank you! :zap::panda_face:

Issue #712, opened 2016-03-14
Update directory tree for blog to show "apps.py" on Step 10: Django models


ProProgrammer:
Running django on a Windows machine. For me, after I run `python manage.py startapp blog`. The blog directory has an additional file `apps.py` the default contents for which are:
```
from django.apps import AppConfig


class BlogConfig(AppConfig):
    name = 'blog'
```

We need to update the directory structure on [Step 10: Django models](http://tutorial.djangogirls.org/en/django_models/index.html) after the paragraph:

> You will notice that a new blog directory is created and it contains a number of files now. Our directories and files in our project should look like this:

helenst:
:+1: I think this is a Django 1.9 thing

helenst:
This could be a good first pull request for somebody.

The file `apps.py` needs adding to the tree in the "Creating an application" section.

https://github.com/DjangoGirls/tutorial/blob/master/en/django_models/README.md

PR #746, opened 2016-04-23
"apps.py" included in directory tree on Step 10. Django models

ProProgrammer:
Fix for #712 - Update directory tree for blog to show "apps.py" on Step 10: Django models.
I have also added `db.sqlite3` in the tree structure and ignored the `myvenv` directory


phalt:
👍 brilliant, thanks! ⚡

PR #749, opened 2016-04-23
Use git add --all instead of -A

ekohl:
Long options are more readable and easier for beginners. It's
self-documenting.

PR #752, opened 2016-04-23
Remove outdated mention of -A

ekohl:
5c1d220a7e00a4ecf10539124bd56130fecd0983 changed -A to --all but the
note wasn't updated. This only fixes it for English, but should be done
for all languages.

See https://github.com/DjangoGirls/tutorial/pull/749 as well.


phalt:
⚡ nice spot

ekohl:
What is the process of getting this translated?

PR #753, opened 2016-04-23
Remove confusing (myvenv) from pythonanywhere prompt

Geekfish:
When you navigate into the project directory at that point of the tutorial in pythonanywhere you actually see something along the lines of:

```
(myvenv) somestuffhere (master)$
```
I think it's simpler to not have the `(myvenv)` part at all (since it's not needed/relevant in that step) and only have what people need to type in.
Otherwise it can get confusing as the previous command doesn't have the same prefix.

### DIFF
--- a/en/html/README.md
+++ b/en/html/README.md
@@ -179,7 +179,7 @@ Once we've done that, we upload (push) our changes up to Github:
 
 ```
 $ cd ~/my-first-blog
-(myvenv)$ git pull
+$ git pull
 [...]
 ```
 


phalt:
Yep, I had a question about this today too! Nice find. 💃

ProProgrammer:
Hi, i beg to differ here. I think `(myvenv)` should be there since it allows people at all time to understand that virtualenv is active and it is supposed to remain thay way.

phalt:
@ProProgrammer in this particular step myvenv isn't required, it's just a `git pull`

ProProgrammer:
Yeah but removing it on one step and having it on others would make it inconsistent, people would get confused why all of a sudden virtualenv is deactivated here. Isnt it?

helenst:
I think it could be more confusing to activate the virtualenv when it's not needed - it makes the purpose more clear if we're only using it when necessary.

There was a PR recently that removed quite a lot of these unnecessary activations, :+1: for consistency

PR #760, opened 2016-05-01
PythonAnywhere app instructions fix

olgakogan:
When filling the venv path in PythonAnywhere, it says `/home/<your-username>/my-first-blog/myvenv/`.
This <your-username> is easy to confuse with previous places where `<your-username>` referred to the github or computer username, so I suggest to make it more explicit


olasitarska:
Thanks! 🌴✨☀️

PR #766, opened 2016-05-04
it: fix "la forma" vs "il form"

xrmx:
form is translated as "il form" in all other occurences, also "la forma" does not make sense in this context :)


patjouk:
Thanks! :cake:

xrmx:
Merci @patjouk !

PR #768, opened 2016-05-07
Added Python Execution in Windows

shante66:
Currently coaching and I realized Many people following the tutorial, and using windows OS are running into the same mistake. I was thinking putting a reminder of 'python' instead of 'python3' could make things easier


aniav:
Thanks! 👏

PR #770, opened 2016-05-21
Fix broken image and add python highlight to a block

echavezNS:
Hi there, I was couch at the DjangogirlsCUU and while the girls were following the tutorial I notice that there was a broken image, so I fixed it.


helenst:
Thanks! :turtle::rose:

PR #775, opened 2016-05-28
In Installation, tell Windows folk how to open cmd

Lingnik:
During installation night, one of our Windows users (Windows 10) ran
into the case where the tutorial was telling her to run some command
line tools before telling her how to open Command Prompt. The OSX
instructions tell the user to open Terminal (although don't explain
how), so I added a section on how to launch Command Prompt and hint that
it will be explained in a forthcoming section.


aniav:
Thank you! 💃

PR #777, opened 2016-05-28
Replaced the `del` statement with an explicit `pop()` method

MarkusH:
Turns out, some attendees get confused by the differences between global functions like `len()` and `print()`, statements like `del` and methods like `reverse()` and `append()`.

I hope I didn't mess up any of the translations :grin:


aniav:
Looks good! ⚡

PR #779, opened 2016-05-28
Added the `p` tag to the list of tags

MarkusH:


helenst:
Good catch, thanks! :sunflower:

PR #780, opened 2016-05-28
Removed www from pythonanwhere admin url

jennxf:
URL specified does not work. 
Should be https://yourname.pythonanywhere.com/admin


helenst:
Thank you! :tulip:

PR #787, opened 2016-06-08
EN: fix bad url

MaxymVlasov:


helenst:
Well spotted, thanks! :cake::blue_heart:

PR #803, opened 2016-06-24
Remove duplicate images in Spanish translation

logological:
The Spanish translation contains an unused folder with duplicate images.  This commit removes it.


aniav:
🎉

PR #804, opened 2016-06-25
Line order instruction about Lobster font

thatgirlcodes:
Adding a hint as to the order of the addition of the lobster font to the CSS.

If the new line is inserted after the link to `blog/static/css/blog.css` then the font doesn't change. We (with @maxwell-k) noticed this at the second Belfast DjangoGirls event.


aniav:
I had no idea this has to be in any specific order. Thanks! ✨

PR #806, opened 2016-07-02
Remind people that the server needs be killed.

smitec:
remind people to stop the server before creating a superuser


tomtomau:
Change to include Mac/ Windows:
```
CTRL+C or CTRL+Break
```

smitec:
Yeah it is a tough one, we ran into a few problems related to either the server running when it shouldn't be or not running when it should.

I agree a second window is a good idea, the server being unable to detect new files (as opposed to file changes) may complicate some things there though.

helenst:
Has been fixed by #811

PR #809, opened 2016-07-06
Correction python3 to python3.5 at prompt in Tutorial

riseriyo:
Minor correction: Need at the prompt 'python3.5' instead of 'python3' when creating 'myvenv' virtual environment in the tutorial


helenst:
Thank you! :sunflower:

Issue #491, opened 2015-10-14
"python managy.py runserver" needs its own console window


xyzisinus:
As of 10/10/2015, the tutorial in chapter 9 starts a server in a command line terminal.  In the following chapters it doesn't mention where to type the subsequent commands.  As the result, my students simply typed the new commands into the server's window when the server is still running.  The tutorial should say "before you try something using url localhost:8000 (by the way, "localhost" is much better than "127.0.0.1") start the server.  To type the new commands, stop the server using control-c".  This should be said AFTER the "it worked" page is shown.  Otherwise, the information doesn't register.

A better approach:

1. Leave the server window alone to let the server run -- that's what a server does anyway.  Ask the student to observe the server output after each action on the page, such as "reload".  My students were in awe when seeing their actions cause the server to spill out something.

2. Start another window, cd to the right place and re-run the environment script prior to typing new commands.  Be sure to point out when the server must be restarted after certain file changes.

domenkozar:
+1, it was very confusing to explain `ctrl+c` and switching back and forth with other terminal work during the workshop.

aniav:
This is now fixed :)

PR #811, opened 2016-07-13
Add reminders to open new window for new commands while server is running

jdelbello:


helenst:
Thank you! :tulip:

aniav:
I believe this fixes #491 :)

PR #812, opened 2016-07-15
Add instructions to upgrade pip

### LINKED ISSUE #774: Get them to upgrade pip
Older versions of pip have issues with the version specifier when we install Django. (see #704)

Attendees sometimes also get distracted by pip's messages suggesting they upgrade pip.

I suggest we get them to run `pip install --upgrade pip` after creating the virtualenv, so everybody's starting from the same point.

helenst:
Fixes #774


aniav:
Woohoo! ✨

Issue #300, opened 2015-03-28
Add gitbook-plugin-codeblocks-filename


olasitarska:
It's sometimes confusing which file you should edit, so gitbook plugin like [gitbook-plugin-codeblock-filename](http://plugins.gitbook.com/plugin/codeblock-filename) could be very helpful. However, that's gonna be very time consuming to implement in various translations :( 

It looks like that on gitbook: 

![](http://cl.ly/image/0f3D071E3i1H/Screen%20Shot%202015-03-28%20at%2017.33.17.png)

aniav:
https://github.com/litmon/gitbook-plugin-codeblock-filename/pull/1 bam!

PR #817, opened 2016-07-22
Filenames over code blocks

aniav:
Now this has been hanging for quite some time. We are adding descriptions over the code blocks to show if a person should type the commands into the console window or in a specific file. The plugin is now rewritten and shouldn't produce errors.


aniav:
This finally closed #300

PR #820, opened 2016-07-23
Provide link to skip Chromebook instructions

charlesdwesley:
This is probably the wrong way to do this, but when reading the installation instructions I wasn't sure how far to skip down the page to skip the Chromebook instructions. So I added a link that would skip down the page for you.


aniav:
Thank you! ✨

PR #821, opened 2016-07-23
Include security change on OS X before installing a pkg

charlesdwesley:
While following the installation instructions, we got a message that said the Python installation pkg "is damaged and can't be opened." Googling the error indicated we needed to change the setting to allow apps downloaded from identified developers, so I added that as a callout in the docs.


aniav:
Thank you again! 🍝

PR #822, opened 2016-07-23
Change localhost to 127.0.0.1

charlesdwesley:
This matches both with the image below and standardizes, since the tutorial uses "127.0.0.1" later on in http://tutorial.djangogirls.org/en/django_admin/#django-admin.


aniav:
Thank you! 👾

PR #851, opened 2016-09-03
Update README.md

gkhaburzaniya:
Reading this, the definition of a punishment for children was brought to my mind more readily than the other definitions time-out, so it sounds a little funny. http://www.merriam-webster.com/dictionary/time%E2%80%93out I think this is because time-outs aren't usually used as breaks. they're used as planning sessions. So for the other definition, people don't generally "deserve" a time-out.


helenst:
Great point! We don't want to punish anybody for completing the chapter :laughing: Thank you!

PR #852, opened 2016-09-03
Update django-models text

gkhaburzaniya:
This is a list using square brackets in my settings.py file


helenst:
Thanks, this changed with Django 1.9 and went unnoticed :sparkling_heart:

PR #856, opened 2016-09-08
instructions.md: --version without ndash

che0:
hyphen-minus instead of ndash in `--version`


helenst:
Thanks! :cherry_blossom:

PR #864, opened 2016-09-19
Fix path highlight hiding username placeholder

thibaudcolas:
In the french and portuguese translations, a particular file path in the "Deploy!" step is higlighted with bold (*) instead of backticks (`). This is causing issues with the username placeholder because of the way the Markdown is interpreted:

*/home/<your-PythonAnywhere-username>/my-first-blog/myvenv/* vs `/home/<your-PythonAnywhere-username>/my-first-blog/myvenv/`.

Edit: I spotted this issue in the french translation first, then did a global search and haven't found it anywhere else but the Portuguese translation.


patjouk:
Thanks! :rose:

PR #865, opened 2016-09-19
Change verb in french translation

thibaudcolas:
The original english text says:

> In the "Virtualenv" section, click the red text that says "**Enter** the path to a virtualenv", and **enter** `/home/<your-PythonAnywhere-username>/my-first-blog/myvenv/`.

The french translation says:

> Dans la section "Virtualenv", cliquez sur le texte en rouge qui indique "**Enter** the path to a virtualenv" (**ajouter** le chemin d'accès de votre environnement virtuel), et **ajouter** ceci : `/home/<your-username>/my-first-blog/myvenv/`.

Enter shouldn't be translated as "ajouter", whose most common translation would be "add". Here we really don't want people to add that path at the end of whatever is in that field. "Entrer" is the appropriate word here – a common translation for "Enter" in the IT context, eg. "Enter your password" – "Entrer votre mot de passe".

### DIFF
--- a/fr/deploy/README.md
+++ b/fr/deploy/README.md
@@ -226,7 +226,7 @@ Une fois l'assistant fermé, vous serez automatiquement conduite sur la page de 
 
  [7]: images/pythonanywhere_web_tab_virtualenv.png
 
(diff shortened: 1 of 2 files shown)


patjouk:
Thanks! :sparkles:

PR #868, opened 2016-09-24
Revised wording on URL page to emphasize need to add 'Include' to pyt…

kbenzel:
…hon import statement.


olasitarska:
Hi @kbenzel! Thanks for this contribution. We do prefer more friendly wording in the original version, but agree that maybe some emphasis would be good. Do you mind submitting another PR that would take that into account? Have a lovely day!

Issue #862, opened 2016-09-18
Installation page is long


helenst:
Some attendees find the installation chapter too long and a bit off-putting (especially as the first thing they do). Perhaps it could be split into sub-chapters?

(Thanks to Emma Delescolle for raising this in her talk at PyCon UK :)

chozabu:
Totally agreed, what would be the best way to split it up? It could be good to have a page for each OS, though there would be some repetition for each system, it'd be great to be able to follow a linear set of instructions

helenst:
I think each section would still be really long if split by OS (though having a section for Chromebook might still make sense since it's quite different)

I think it would be best to split it up by steps, e.g.:
Chromebook setup
Install Python
Set up virtualenv
Install django
Install a code editor
Install git
Create accounts
Background reading

Some of those might still be a little on the long side but some kind of tabbing system for OS switching would also help (see https://github.com/DjangoGirls/tutorial/issues/188 if you are interested in looking at this).

aniav:
As for the OS splitting we have another issue to use a plugin, which would add tabs on every code block instruction, this way, a person would choose an OS once and all the code blocks would be suitable for their OS. So I would definitely not split the chapters by OS, this will be done by a plugin.

Other than that, it is worth noting that the installation chapter is created from instructions in further chapters and they are already split, this one chapter is just a short summary. Maybe instead of splitting it we should create a more through introduction to the chapter for the attendees as well as the coaches, so everyone would know that they can in fact, go through the chapters slowly instead of storming through the installation chapter.

The installation chapter was created for the purpose of the installation party, where you just need to install stuff, and next you should go through all the nice informative stuff in further chapters.

Does this sound sane? 🌹

aniav:
Also a ping to @nanuxbe so she would also know that a discussion is being handled here :)

helenst:
@aniav ah thanks, I didn't know her github handle :)

chozabu:
pinging @lukestanley - who has been into at this too - info from https://github.com/DjangoGirls/tutorial/issues/188 (linked above) may be very handy

Another option to make the section seem shorter could be collapsible or hover-able items? To provide more detailed information when needed

But really not sure what is best, all of the suggestions above sound good - would be great to have some kind of A-B testing going on, but that is probably beyond the scope of this ticket/issue

delenamalan:
I opened PR #874 to address this issue using the `sectionx` extension.

I started with the Python installation instructions to show how it might behave.

[...]

PR #874, opened 2016-10-09
Add sectionx and use in installation instructions

delenamalan:
Use the `sectionx` extension to put instructions for different operating systems in separate sections to address #862 .

Started with the Python installation instructions to show what it might look like.

Example of Python installation instructions in separate sections.

Add `sectionx` to:
- [x] Chromebook setup.
- [x] Python installation.
- [x] Virtualenv usage.
- [x] Django installation.
- [x] Git installation.

### DIFF
--- a/book.json
+++ b/book.json
@@ -12,7 +12,8 @@
         "github@2.0.0",
         "language-picker",
         "sidebar-ad",
-        "codeblock-label"
+        "codeblock-label",
+        "sectionx"
     ],
     "pluginsConfig": {
         "ga": {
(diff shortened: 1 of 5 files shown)


aniav:
Awesome job! Do you want this to be merged now or would you like to make some more changes using this extension before merging? :)

delenamalan:
Thanks! I will make some more changes before merging. I'll use the extension for the other installation instructions as well. :)

delenamalan:
All right, @aniav , I think this can be merged.

Good to know, the pdf generated from the book is still fine after using `sectionx`:
Screenshot of pdf generated after sectionx was used.

patjouk:
Thanks a lot for this PR! Your changes are now live and they look awesome! :sparkles:

PR #875, opened 2016-10-11
Update README.md

ricoms:
I've added my name to introduction as volunteer. I've translated more than 300 sentences and voted for all good translating done before me. :) Can I?


patjouk:
Ok :)

PR #877, opened 2016-10-12
clarify where primary key comes from & how to reference it

hancush:
When I walked through the tutorial, I got hung up on where post.pk came from and, later, bugs introduced by renaming the pk variable in the URL pattern without also changing the variable name in the template link and view function – which isn't made explicit in the tutorial. These edits clarify/rectify where I went wrong!


olasitarska:
Thanks for this pull request @hancush and @souravsingh for the review! 🍭

PR #891, opened 2016-10-27
Fix backslash -> slash

rhcarvalho:
Came across this when going through the tutorial with my team :smile:


patjouk:
Thanks! :sparkles:

PR #898, opened 2016-11-05
Fix #894 order of OS instructions in installation chapter

### LINKED ISSUE #894: Keep the same OS instructions' order in the installation chapter
Instructions in the installation chapter should follow this order:
1. Windows
2. MacOS X
3. Linux

It needs to be fixed in [this file](https://github.com/DjangoGirls/tutorial/blob/master/en/python_installation/instructions.md).

ubracklow:


patjouk:
Thanks! :sparkles:

PR #899, opened 2016-11-05
Gitter badge: update linke and remove tracking

bittner:
- Update Gitter badge URL format -- [_Beautiful is better than ugly._](https://www.python.org/dev/peps/pep-0020/#id3) :heart_eyes: 
- Remove tracking data (query string) from Gitter badge link -- [_Now is better than never._](https://www.python.org/dev/peps/pep-0020/#id3) _... to remove tracking._  :smirk: 
- Make word wrapping a bit nicer in licensing paragraph. -- [_Readability counts._](https://www.python.org/dev/peps/pep-0020/#id3) :relieved:


patjouk:
Thanks! :sparkles:

PR #900, opened 2016-11-06
Fix for issue #188 

ubracklow:


patjouk:
Thanks! :chocolate_bar:

PR #901, opened 2016-11-07
Fixed bug with allowed hosts.

im-auld:
This was an issue several people ran in to during the last conference. There is probably a better fix for this but this one is simple and can be done quickly without distracting from the rest of the tutorial.


patjouk:
Thanks! :sparkles:

PR #903, opened 2016-11-08
Remove duplicate images for Russian (ru) and Korean (ko)

bittner:
In the `css/images/` folder for Korean and Russian there is the `images` folder again, with the identical set of images as at the level above. This PR removes those files again.


olasitarska:
Thanks @bittner! :)

PR #904, opened 2016-11-12
Use ~ rather than /home/<your_name>

ekohl:
This allows copy-pasting and doesn't require knowing the username.

Note that I don't know if there's a Windows equivalent. There may also be a better wording of the paragraph.


ekohl:
Oh, and I also didn't test the Mac version but I assume that works. Can't hurt to verify this.

olasitarska:
Thanks @ekohl! 🌷

ekohl:
For windows there's [%homepath%](http://environmentvariables.org/HomePath). I don't have a windows install to verify this so I'm hesitant to come up with a patch.

PR #905, opened 2016-11-14
Correct sample code in If...elif...else section

kmaria:
Correct sample code in If...elif...else section


patjouk:
Thanks! :sparkles:

PR #906, opened 2016-11-18
ALLOWED_HOSTS = ['*']

fajpunk:
Hi!  I'm going to be a mentor at the Chicago workshop on Saturday, December 3rd.  I think I found a stumbling block in the tutorial.

Starting in Django 1.8.16, the ALLOWED_HOSTS setting is checked even if `DEBUG` is `True`, so folks going through this tutorial will need to modify it.  I debated whether or not to set it to `['*']` in the `settings.py` section, or to go back and edit it during the deployment section.  This seemed like the easiest thing to do in the interest of getting started, but I'm glad to submit a PR for the other way if you folks think that's better.

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -77,6 +77,15 @@ In `settings.py`, find the line that contains `TIME_ZONE` and modify it to choos
 TIME_ZONE = 'Europe/Berlin'
 ```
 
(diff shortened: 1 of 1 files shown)


fajpunk:
...and it looks like someone already submitted a PR for this, so I'll close mine :)
https://github.com/DjangoGirls/tutorial/pull/901/files

PR #908, opened 2016-11-18
Inform attendees what's a command and what's command output

MarkusH:


en/django_installation/instructions.md
```diff
@@ -165,6 +165,8 @@ In the console, run `pip install --upgrade pip`.
 
 Then run `pip install django~=1.9.0` (note that we use a tilde followed by an equal sign: `~=`) to install Django.
 
+**NOTE:** When you see command-line examples like the following you only need to type what is written after the `$` character (or `>` on Windows). The other lines represents the output of the command you entered.
```
patjouk: It's a great idea but you should probably put this the first time they encounter the command line. It's in [python instruction file](https://github.com/DjangoGirls/tutorial/blob/master/en/python_installation/instructions.md).

das-g: > [...] you should probably put this the first time they encounter the command line. It's in python instruction file.

Actually, the first time is even before that: in the "Introduction to the command-line interface" chapter ([book](https://tutorial.djangogirls.org/en/intro_to_command_line/#your-first-command-yay) / [source](https://github.com/DjangoGirls/tutorial/tree/a26bcdaf536f144ec6d17d0808c3d2bf64137db9/en/intro_to_command_line#your-first-command-yay))

das-g:
I wonder whether we could visually discriminate between input and output, e.g. having input in black font (and maybe bold typeface) and the output (including the prompt) in a gray that is dark enough to be well legible on the light gray code block background, but light enough to look different from black.

Dunno what MarkDown/GitBook Markup and CSS would be needed to achieve that, though, or whether it's feasible at all.

aniav:
I think this has been fixed in a different change and the information about ignoring the prompt is in the tutorial when we introduce command line. So we are good, I am closing that, thank you! 🏋️‍♀️

PR #919, opened 2016-11-26
Move explanation of Git to the 'install' file

helenst:
This means the explain of what git actually is will be included
in the install instructions page as well.


patjouk:
:ok_hand:

PR #920, opened 2016-11-26
Matching welcoming texts

PaolaTosca:
Changed: The default page for your site should say "Welcome to Django!"
To: The default page for your site should say "It worked!"


patjouk:
Thanks! :sparkles:

PR #922, opened 2016-11-26
Make pip upgrade instruction clearer

helenst:
Putting it inline makes it less clear where to type it, and also makes people include the '.' when they shouldn't.


patjouk:
Thanks! :sparkles:

Issue #196, opened 2014-11-16
Disillusionment from getting errors


freakboy3742:
One of my students got really disillusioned with the process of using errors as a teaching tool. "I typed everything right, but then I got an error, and I kept feeling stupid". It wasn't clear to her that getting an error was actually the point of the exercise, as part of the process of teaching what _can_ go wrong, how to recover. 

She suggested using icons/coloring to identify the "Even if you do this right, it won't work" steps in the tutorial, so that when you hit one of those steps, the student isn't left with the impression that the problem is them.

vrutkovs:
I guess rephrasing some parts (e.g. incorrect function name in http://tutorial.djangogirls.org/en/python_introduction/README.html#your-own-functions) would help. 

People usually read the chapter line by line and get confused that things break (on such an early stage). The coach has to know this situation in beforehand and warn the student (I almost filed a bug about incorrect syntax in the example)

jacebrowning:
Could I get some feedback on this proposal? https://github.com/DjangoGirls/tutorial/pull/1225

<a href="https://cl.ly/2y0L3O3l0F0b" target="_blank"><img src="https://s3.amazonaws.com/zapier-cloudapp/items/2x2T3x2G2x053e312040/Screen%20Shot%202018-02-19%20at%208.04.02%20PM.png" style="display: block;height: auto;width: 100%;"/></a>

If we want to go down this route we might considering changing or modifying the `filename` plugin.

jacebrowning:
This change is now live for English: https://tutorial.djangogirls.org/en/python_introduction/#errors

ekohl:
Awesome. This looks like a big improvement! Thanks @jacebrowning for closing this very old issue.

PR #933, opened 2016-12-02
Mention some possible errors in the python intro

MatMoore:
Hello, this is an attempt to help with https://github.com/DjangoGirls/tutorial/issues/196

Specifically, @vrutkovs's suggestion:
> People usually read the chapter line by line and get confused that
> things break (on such an early stage). The coach has to know this
> situation in beforehand and warn the student (I almost filed a bug about
> incorrect syntax in the example)

This PR just mentions some things that can go wrong when defining a function, instead of saying "that was easy".

I think adding some styling to display these kinds of things in a different way would help a lot, but I wanted to try and see if rephrasing some bits would help. Happy to extend this if it's helpful, but wanted to keep it simple to start with as I haven't read the full tutorial myself.

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -793,7 +793,11 @@ Hi there!
 How are you?
 ```
 
-That was easy! Let's build our first function with parameters. We will use the previous example – a function that says 'hi' to the person running it – with a name:
+If you see an error instead, don't panic! Python is trying to help you find the source of the problem.
+If you get a **NameError**, that probably means you typed something wrong, so you should check that you used the same name when creating the function with `def hi():` and when calling it with `hi()`.
+If you get an **IndentationError**, check you don't have any extra whitespace at the start of the line: python wants all the code inside in your function to be neatly aligned.
+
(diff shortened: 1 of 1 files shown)


en/python_introduction/README.md
```diff
@@ -793,7 +793,11 @@ Hi there!
 How are you?
 ```
 
-That was easy! Let's build our first function with parameters. We will use the previous example – a function that says 'hi' to the person running it – with a name:
+If you see an error instead, don't panic! Python is trying to help you find the source of the problem.
+If you get a **NameError**, that probably means you typed something wrong, so you should check that you used the same name when creating the function with `def hi():` and when calling it with `hi()`.
+If you get an **IndentationError**, check you don't have any extra whitespace at the start of the line: python wants all the code inside in your function to be neatly aligned.
+
```
MatMoore: Yeah, that makes sense. I've reformatted it as a note with a list. What do you think of the latest commit?

helenst:
This is awesome, thank you! :whale::tulip::star2:

PR #935, opened 2016-12-03
Fix grammar and tyops

SEJeff:
While going through the tutorial with a few mentees, I spotted a few minor issues. Lets fix them up.


stantonk:
👍

SEJeff:
Note that the final three commits were directly from things multiple people got caught up on. I'm totally open to removing or rewording things if preferred.

stantonk:
👍

helenst:
Thanks! :whale:

PR #942, opened 2016-12-12
Replace another instance of /home/<your_name> with ~

ekohl:
In addition to https://github.com/DjangoGirls/tutorial/pull/904


helenst:
Thanks! :whale2:

PR #943, opened 2016-12-12
Use %HomePath% instead of C:\Users\<your_name>

ekohl:
This is the windows version of a7f5e56c03168f2e68164b95961c9d15d6d108be. By using a variable users can copy paste and don't have to figure out their username.

As stated in https://github.com/DjangoGirls/tutorial/pull/904#issuecomment-266379543 I don't have a Windows install to verify this works. This Thursday I'm going to a follow up meetup where I might have the chance to test this.


helenst:
Thanks! :tulip:

PR #947, opened 2016-12-18
Fix `\*` typo

mmcgahan:
`\*` will throw an error if used as an operator - it should just be `*` as in `2 * 3`. It looks correct in the other tutorial translations.


helenst:
Thanks! :)

PR #950, opened 2016-12-20
Correct path to `startproject` command on Windows.

heldinz:
The current instruction to start the project fails silently under Windows:

```
(myvenv) C:\Users\Name\djangogirls> django-admin.py startproject mysite .
```

This PR provides the path to `django-admin.py` within the virtual environment to ensure that the project can be started correctly. For some reason, the command is not being picked up otherwise, even though the virtual environment is activated.

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -38,7 +38,7 @@ On Windows you should run the following command. **(Don't forget to add the peri
 
 {% filename %}command-line{% endfilename %}
 ```
-(myvenv) C:\Users\Name\djangogirls> django-admin.py startproject mysite .
+(myvenv) C:\Users\Name\djangogirls> myvenv\Scripts\django-admin.py startproject mysite .
 ```
 > The period `.` is crucial because it tells the script to install Django in your current directory (for which the period `.` is a short-hand reference).
 


olasitarska:
I'd love if someone with Windows could verify this.

patjouk:
I've tried it on Windows: it works without adding `myvenv\Scripts\`. Can you give me more details on how you've encountered this bug?

heldinz:
Two participants at the Leipzig workshop were unable to create the project via the startproject command. There was no error. We reinstalled Python, ensuring that the PATH checkbox was checked, and recreated the virtual environment but had the same issue. Running the command with the full path was the only way to create the project for these two participants.

heldinz:
If this issue is only affecting a few users, maybe we could add this as a troubleshooting note to the tutorial instead, since I don't have the means to diagnose the problem further. It took some time to find a solution and was extremely frustrating for the participants involved, so I'd like to spare others that frustration.

das-g:
> We reinstalled [...] recreated the virtual environment but had the same issue.

Did you/they also [activate](https://tutorial.djangogirls.org/en/django_installation/#working-with-virtualenv) the virtual environment after re-creating it? Did trying to do so result in any error messages in the terminal?

A not (or not properly) activated virtualenv might be an explanation for the behavior they observed.

heldinz:
Yes, they did, and there were no errors.

das-g:
Huh. Then a troubleshooting note might indeed be warranted.

heldinz:
Yes, it was definitely strange. I wasn't present for the installation evening where she initially installed everything, but we redid all the steps starting with installing Python up to activating the (newly-created) virtual environment, and the `startproject` command still had no effect.

Perhaps:

> Note for Windows users: If you don't see the directory structure after running the `startproject` command, first verify that you have followed all steps correctly. If the command still does not create the expected directory structure, try running the command like this: 

```
(myvenv) C:\Users\Name\djangogirls> myvenv\Scripts\django-admin.py startproject mysite .
```

aniav:
This is now deprecated as `django-admin.py` is deprecated and now there is just `django-admin` or `django-admin.exe`. Just checked on Windows :)

PR #952, opened 2016-12-22
Mcr1

lwyso:
altered explanation to match correct operating system to stop confusions for reader.
also added a missing label 'command line' in python introduction page


patjouk:
Thanks! :tulip:

PR #953, opened 2016-12-22
gitbook.io no longer exists

kevbo:
also https


helenst:
Well spotted, thank you :-)

PR #958, opened 2017-01-21
[Portuguese] Fixed django version to avoid version conflict

seocam:


ayr-ton:
The PR #1050 provides an alternative solution to this step, via requirements.txt.

aniav:
This has been fixed already on another PR. Thank you! :)

PR #960, opened 2017-01-25
A more positive feel for the "Introduction"

bauerv:
The version "feel somehow left behind" has a strong negative feeling to it. To me (as a coach) this feels problematic/counterproductive when starting off the day with super motivated participants (essentially, I'm "reminding" them about the "fact" that they might be "left behind" instead of, e.g., "courageous" or "engaged"). 

I'm proposing a small correction to soften this feel. However, ideally, I would prefer the tutorial to start off with a friendly and warm welcome (and potentially congratulations that the participant worked up the courage to get herself into this in the first place). Maybe add a dedicated "Welcome" section?

### DIFF
--- a/en/README.md
+++ b/en/README.md
@@ -6,7 +6,7 @@
 
 ## Introduction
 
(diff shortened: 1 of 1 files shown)


patjouk:
Hi! This is great :heart: 
If you feel like writing it, I would to read your welcome paragraph and add it to the tutorial!

PR #961, opened 2017-01-26
Added clarification about directory content

bauerv:
On several occasions, we had confusion about the contents of the directory after running the django-admin.py script. The directory tree suggests to show all the contents, but in fact, when following the installation guideline, venv will also sit in the djangogirls directory. This is missing in the pic, so participants sometimes think it is an error and might try to delete it. I suggest a note to fix this.


patjouk:
Thanks! :sparkles:

PR #963, opened 2017-01-27
merge from djangogirls (#2)

bauerv:
* inclusos for incluso

* Fixed, code seperated line from 123. Line

* A more positive feel for the "Introduction" 

The version "feel somehow left behind" has a strong negative feeling to it. To me (as a coach) this feels problematic/counterproductive when starting off the day with super motivated participants (essentially, I'm "reminding" them about the "fact" that they might be "left behind" instead of, e.g., "courageous" or "engaged"). 

I'm proposing a small correction to soften this feel. However, ideally, I would prefer the tutorial to start off with a friendly and warm welcome (and potentially congratulations that the participant worked up the courage to get herself into this in the first place). Maybe add a dedicated "Welcome" section?

* Added clarification about directory content

On several occasions, we had confusion about the contents of the directory after running the django-admin.py script. The directory tree suggests to show all the contents, but in fact, when following the installation guideline, venv will also sit in the djangogirls directory. This is missing in the pic, so participants sometimes think it is an error and might try to delete it. I suggest a note to fix this.

### DIFF
--- a/en/README.md
+++ b/en/README.md
@@ -4,6 +4,10 @@
 > This work is licensed under the Creative Commons Attribution-ShareAlike 4.0 International License.
 > To view a copy of this license, visit https://creativecommons.org/licenses/by-sa/4.0/
 
+## Welcome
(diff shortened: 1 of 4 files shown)


bauerv:
Please delete this merge.. I mixed something up and can't find how to fix it. Sorry about that!

PR #964, opened 2017-01-27
Consistency between text and code example

bauerv:
`blog.views.post_detail` is not in the source code example so the explanation in the following paragraph was not consistent with it. From what I understand, it is the URL created by Django, so I added this to the text.

### DIFF
--- a/en/extend_your_application/README.md
+++ b/en/extend_your_application/README.md
@@ -37,11 +37,11 @@ We will start with adding a link inside `blog/templates/blog/post_list.html` fil
 
 {% raw %}Time to explain the mysterious `{% url 'post_detail' pk=post.pk %}`. As you might suspect, the `{% %}` notation means that we are using Django template tags. This time we will use one that will create a URL for us!{% endraw %}
 
-`blog.views.post_detail` is a path to a `post_detail` *view* we want to create. Please note: `blog` is the name of our application (the directory `blog`), `views` is from the name of the `views.py` file and the last bit – `post_detail` – is the name of the *view*.
+The URL that is created resolves to `blog.views.post_detail` and is a path to a `post_detail` *view* we want to create. Please note: `blog` is the name of our application (the directory `blog`), `views` is from the name of the `views.py` file and the last bit – `post_detail` – is the name of the *view*.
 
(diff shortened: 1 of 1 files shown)


en/extend_your_application/README.md
```diff
@@ -37,11 +37,11 @@ We will start with adding a link inside `blog/templates/blog/post_list.html` fil
 
 {% raw %}Time to explain the mysterious `{% url 'post_detail' pk=post.pk %}`. As you might suspect, the `{% %}` notation means that we are using Django template tags. This time we will use one that will create a URL for us!{% endraw %}
 
-`blog.views.post_detail` is a path to a `post_detail` *view* we want to create. Please note: `blog` is the name of our application (the directory `blog`), `views` is from the name of the `views.py` file and the last bit – `post_detail` – is the name of the *view*.
+The URL that is created resolves to `blog.views.post_detail` and is a path to a `post_detail` *view* we want to create. Please note: `blog` is the name of our application (the directory `blog`), `views` is from the name of the `views.py` file and the last bit – `post_detail` – is the name of the *view*.
```
helenst: I'm not sure about this explanation - I think the part we want to get across here is that the `post_detail` must match a URL with that name. I think we could drop the bit that explains about what the different parts mean, and including something equivalent further down (i.e. when the url rule is added, link that to the location of the view). Does that make sense? :sparkles:

bauerv: Yes, sounds good with me, do you have a concrete suggestion what to write?
I think dropping the current description is a good idea: We were looking at this with a few people and could not agree what it really meant.

helenst: How about this to replace it:

> The `post_detail` part means that Django will be expecting a URL in `blog/urls.py` with `name=post_detail`

And later on (in the part where we add a view) something like:

> In `blog/urls.py we created a URL rule named `post_detail` that refers to a view called `views.post_detail`. This means that Django will be expecting a view function called `post_detail` inside `blog/views.py`.

I am not sure about the exact wording of that last bit, feel free to adjust to make it clearer :-)

helenst: Hi @bauerv would you like to update the pull request with the suggested text (or something similar) so we can merge this?

bauerv:
Hi Helen,

Thanks for getting back to me and the reminder. I’m super-busy until next week, but I could do it then. Is that fine with you?

Cheers!
Veronika

helenst:
Of course - we will be super happy for your help with this whenever you have time, there is no rush. I just wanted to check that you are still happy to do it! :cherry_blossom:

aniav:
I have updated the text to match the code, thanks! :tada:

PR #967, opened 2017-02-01
Reminder to save files

bauerv:
Coming from the command line, participants keep forgetting to save their edits and keep getting lots of errors. Just a small reminder.


helenst:
Thanks! :fireworks:

PR #968, opened 2017-02-01
Short note how/why to open files in code editor

bauerv:
Caused confusion in workshop, participants suggested adding a note

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -73,7 +73,7 @@ Let's ignore the other files for now as we won't change them. The only thing to 
 
 Let's make some changes in `mysite/settings.py`. Open the file using the code editor you installed earlier.
 
-**Note**: Keep in mind that `settings.py` is a regular file, like any other. You can open it from inside the code editor, using the "open -> select file". Alternatively, you can open the file by navigating to the djangogirls folder on your desktop and right-clicking on it. Then, select your code editor from the list. Selecting the editor is important as you might have other programs installed that can open the file but will not let you edit it.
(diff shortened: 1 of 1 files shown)


en/django_start_project/README.md
```diff
@@ -73,6 +73,8 @@ Let's ignore the other files for now as we won't change them. The only thing to
 
 Let's make some changes in `mysite/settings.py`. Open the file using the code editor you installed earlier.
 
+**Note**: Keep in mind that `settings.py` is a regular file, like any other. You can open it from inside the code editor, using the "open -> select file". Alternatively, you can open the file by navigating to the djangogirls folder on your desktop and right-clicking on it. Then, select your code editor from the list. Selecting the editor is important as you might have other programs installed that can open the file but will not let you edit it.
```
helenst: Is "open -> select file" specific to a particular editor? In Atom at least, it is something different.

bauerv: I checked again (using sublime) where it is actually "file -> open" which gets you the usual explorer/finder window that allows you to navigate to and select the file you're interested in. Is this the same on atom?

bauerv: I updated the file, please let me know if it's more clear now.

helenst: I still think that instructions for any one editor are going to cause confusion for others, or be unnecessarily specific. Similarly I think desktop / right click instruction is not necessarily applicable to all systems either.

I am not sure what the best approach is here. We could start giving instructions for each editor but that would add more bulk to the tutorial. We could make the wider decision to just pick one editor rather than presenting a choice of three, which would make it easier to provide editor instructions. I know attendees do often struggle with the editor as it's a new tool for them and not all are familiar with navigating their way around, so there is probably some need to provide extra help here. I don't know what is best, maybe other contributors have opinions about this?

aniav:
This was long time waiting. I am merging it, let's fix problems as they appear :)

PR #971, opened 2017-02-02
Windows10 error

dns2utf8:
We just made the tutorial on windows an linux in parallel and noticed the command did not work.

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -38,7 +38,7 @@ On Windows you should run the following command. **(Don't forget to add the peri
 
 {% filename %}command-line{% endfilename %}
 ```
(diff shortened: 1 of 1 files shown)


brondsem:
Hm, the original `django-admin.py` works fine for me in Windows 10.  (.exe works too)

dns2utf8:
Interesting, we worked with python 3.6, maybe this is a new bug?

aniav:
`django-admin.py` is deprecated so I am accepting that! :)

dns2utf8:
ah, thank you for searching for the reason.

EllekeB:
For me django-admin.exe does not work, I get the message "access denied" - changing to admin doesn't help, but replacing .exe with .py does work.

PR #972, opened 2017-02-02
Explain the LANGUAGE_CODE

dns2utf8:
Adding some text for the language code variable.


helenst:
Thank you for your contribution. I think this would be great to have, but with a couple of improvements:

1. A little more explanation of why they would want to do this and/or how it affects the site they are building
2. A link to the Django docs for this setting (https://docs.djangoproject.com/en/1.10/ref/settings/#language-code) so attendees can find the correct value for their own location.

Thanks! :sunflower:

aniav:
I have added the required fixed and merged! 😇

dns2utf8:
cool

PR #973, opened 2017-02-03
Windows command prompt updates

brondsem:
* Update for Windows 10
* Update venv command to not use hardcoded path to `python`


aniav:
Thank you! 🎉

PR #974, opened 2017-02-03
Simpler, clearer way to check for 32/64-bit windows

brondsem:
The linked support page does not cover Windows 10, and has complex
different steps for each other version of Windows.  The Win+Break
shortcut should work on many versions of Windows and is simple.


aniav:
Thank you and sorry for the long wait! :)

PR #975, opened 2017-02-03
Git installer now has this checkbox in step 3 not 5

brondsem:


brondsem:
Nevermind, sure seems like step 5 still, not sure what I was seeing yesterday :)

PR #976, opened 2017-02-03
PythonAnywhere sometimes gets weird without www

astronouth7303:
Right now, https://pythonanywhere.com/ is timing out but https://www.pythonanywhere.com/ is working.


helenst:
Thanks! :tada:

PR #980, opened 2017-02-07
start server on port 8000

das-g:
Start server on port 8000, as that is the port we'll access with the browser [further below in these instructions](https://github.com/das-g/django-girls-tutorial/blame/ea7948eb5a6c62197635cb39f570a149a02c4035/en/django_start_project/README.md#L173).


aniav:
Thank you! :)

PR #981, opened 2017-02-10
Fix:  del to  pop in the python_introduction_ko file

xho95:
I guess `pop` method is correct than `del` in the Dictionary part of `ko` version of `python introduction` file.


aniav:
This has already been fixed in another PR. Thank you! 🎉

PR #984, opened 2017-02-15
use "key" rather than "button"

das-g:
when referring to the thingies to press on the **key**board.


uk/template_extending/README.md
```diff
@@ -131,4 +131,4 @@
 
 Ось і все! Перевірте, чи ваш сайт все ще коректно працює :)
 
-> Якщо з'явилася помилка `TemplateDoesNotExists`, то це говорить про те, що файлу `blog/base.html` не існує і ви маєте в консолі запущений `runserver`, спробуйте зупинити його (натиснувши Ctrl+C - кнопки Control і C buttons разом) і перезавантажити командою `python manage.py runserver`.
```
helenst: Did you intend to update the Ukrainian version too?

das-g: Yes, as I could (without actually knowing the language) conclude that the English word "button" was used in the meaning of "key" there, too. Though someone who actually knows Ukrainian should probably review that. Although unlikely, maybe in Ukrainian the foreign word "button" is used to mean what would be "key" in English.

helenst: Hi @das-g It can take some time for translation reviews to happen - rather than waiting for a Ukrainian reviewer to come along and check this. I am also not sure that the word 'buttons' being used originally necessarily means 'keys' is a correct replacement.

If you take the Ukrainian file modification out of this, I'll merge it and we can get your improvements to the English version in place! :sparkles:

das-g: done: 8b18e4c

helenst:
Thank you! :sunflower:

das-g:
You're welcome. :smiley:

For the Ukrainian text, should I file a new separate pull request or will that be handled via Crowdin or ...?

helenst:
Hi @das-g I think the best thing would be to let it go through crowdin. Thanks!

PR #989, opened 2017-02-21
Add missing command line results

Tichau:
Hi,
Great tutorial! thanks for your work :)

There were just some missing command line result. This can be strange for someone which doesn't know the command lines to have a different result than the tutorial.


helenst:
Thank you - I'm sure this will be really helpful for attendees encountering Python for the first time: :sparkling_heart::dolphin:

PR #1007, opened 2017-03-23
[Korean] Add explanation about 'django==1.8'

repositivator:
Added explanation why we type "==1.8" for notifying users to install django of  appropriate version when they need it.


aniav:
This is no longer valid as the tutorial now uses Django 1.11, so I am closing that. Sorry for not merging earlier! :)

PR #1012, opened 2017-04-04
Update QuerySet representations to match Django 1.10

koterpillar:
In Django 1.10, QuerySet repr() was changed to distinguish it from an ordinary
list (https://docs.djangoproject.com/en/1.10/releases/1.10/#miscellaneous).
Update the command output in the tutorial to reflect this.


helenst:
Hi @koterpillar - thank you so much for this! :tada:

PR #1015, opened 2017-04-12
Fixed error on redirect function

eduingles:
redirect function has to call post_detail function without ' '.


helenst:
Hi @eduingles I am going to close this because removing the quote marks would make the code incorrect. Is there a particular error you are seeing with the current code?

eduingles:
I know that code should be correct, but when I follow the Django tutorial in djangogirls.org I need to do that without ' '. In other way the project doesn't work.

Maybe it's a bug I've had in another part of the code.

PR #1021, opened 2017-05-02
[Korean] Typo Corrected

trent-92:
Name of field should be 'text', not 'test'.


aniav:
It seems this has been fixed in another commit. Thank you! :)

PR #1022, opened 2017-05-02
Corrected Cloud 9 port

amdevine:
Corrected the port the server is run on when using Cloud 9 - should be 8080, not the standard 8000.


aniav:
Thanks! :)

PR #1023, opened 2017-05-05
Django official link

tfevan:
edited django official link  with new **django  1.11** version

### DIFF
--- a/cs/whats_next/README.md
+++ b/cs/whats_next/README.md
@@ -30,7 +30,7 @@ Později můžeš zkusit, některý ze zdrojů uvedených níže. Všechny můž
 - [Getting Started With Django video lessons][10]
 - [Two Scoops of Django: Best Practices for Django 1.8 book][11]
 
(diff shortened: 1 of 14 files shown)


aniav:
I think we can't update the link to Django 1.11 in languages which use Django 1.8. If I remember there might have been bigger changes between these versions? Have you checked that?

ekohl:
Even English still explicitly installs 1.10 so IMHO it should really point to the version which ```pip install``` will give.

aniav:
Different languages are translated to different versions of Django because the instructions between versions change a bit and we also change the english version upon pull requests and/or issues on the repo.

The tutorial gives directions on installing Django with a specific version number, currently this is `pip install django~=1.10.0` so if we are changing the version number in the tutorial we need to
a) update all instructions for the specific language
b) check if the instructions shouldn't be changed in terms of deprecation warnings and functional changes in Django itself.

ekohl:
@aniav said what I meant :)

PR #1031, opened 2017-05-12
Altered ALLOWED_HOSTS to include '<your_username>.pythonanywhere.com'…

gsimore:
Updated settings for ALLOWED_HOSTS to include '<your_username>.pythonanywhere.com' to fix the DisallowedHost error from recent Django Security updates.


gsimore:
[English]

aniav:
Thank you 👍

PR #1037, opened 2017-05-16
Proper definition of Crowdin assets

magul:
I have asked a question on slack regarding sync of tutorial between Github and Crowdin (https://djangogirls.slack.com/archives/C03PLJLM8/p1494944212777047).

This pull request resolves issue with syncing source language (`en`) from Github to Crowdin and define, where Crowding should commit translated assets. It also removes few unnecessary options.

Somebody (maybe @olasitarska  as an owner on Crowdin) should also check if API key to project on Crowdin has been regenerate as it cause potential vulnarebality (`This key gives full access to all your Crowdin project data. Treat this just like a password!` see: https://crowdin.com/project/django-girls-tutorial/settings#api)

### DIFF
--- a/crowdin.yaml
+++ b/crowdin.yaml
@@ -1,19 +1,3 @@
-project_identifier: django-girls-tutorial
-api_key: 2be4b1707d0745bc96e9a5c9ffb4ca9b
-base_path: /Users/olasitarska/Desktop/Events/Django Girls/tutorial
-
 files:
-  -
-    source: "/en/**/*.md"
-    translation: "/%two_letters_code%/%original_path%/%original_file_name%"
-    ignore:
-      - /_book
-      - /node_modules
-      - /CONTRIBUTING.md
-  -
-    source: "/en/*.md"
-    translation: "/%two_letters_code%/%original_file_name%"
-    ignore:
-      - /_book
-      - /node_modules
-      - /CONTRIBUTING.md
+  - source: /en/**/*.md
+    translation: /%two_letters_code%/**/%original_file_name%


aniav:
Are you sure we should remove all these lines? Can you add some comments on why did you remove them and what was the problem with each of them?

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
-project_identifier: django-girls-tutorial
```
magul: `project_identifier` is used for API integration, actually We could leave it here, as it does no harm. Do You want me to leave it here (as I believe, we don't use API integration with Crowdin)?

See: https://support.crowdin.com/api/api-integration-setup/

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
...
-api_key: 2be4b1707d0745bc96e9a5c9ffb4ca9b
```
magul: Need to be removed and if not changed before, `api_key` need to be revoked/changed.

See: https://support.crowdin.com/api/api-integration-setup/

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
...
-base_path: /Users/olasitarska/Desktop/Events/Django Girls/tutorial
```
magul: I believe that is correct path on @olasitarska environment and was used with crowdin-cli (actually I believe, that whole content of this file was generated using `crowding generate` command) and doesn't make sense beside her environment.

See: https://support.crowdin.com/cli-tool/#configuration

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
...
 files:
-  -
-    source: "/en/**/*.md"
-    translation: "/%two_letters_code%/%original_path%/%original_file_name%"
-    ignore:
-      - /_book
-      - /node_modules
-      - /CONTRIBUTING.md
-  -
-    source: "/en/*.md"
```
magul: this part of configuration is also covered in `/en/**/*.md`

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
-project_identifier: django-girls-tutorial
-api_key: 2be4b1707d0745bc96e9a5c9ffb4ca9b
-base_path: /Users/olasitarska/Desktop/Events/Django Girls/tutorial
-
 files:
-  -
-    source: "/en/**/*.md"
-    translation: "/%two_letters_code%/%original_path%/%original_file_name%"
-    ignore:
```
magul: There's no so much sense to ignore files, that are not included in `/en/**/*.md` (and `/_book`, `/node_modules` and `/CONTRIBUTING.md` are not).

crowdin.yaml
```diff
@@ -1,19 +1,3 @@
-project_identifier: django-girls-tutorial
-api_key: 2be4b1707d0745bc96e9a5c9ffb4ca9b
-base_path: /Users/olasitarska/Desktop/Events/Django Girls/tutorial
-
 files:
-  -
-    source: "/en/**/*.md"
-    translation: "/%two_letters_code%/%original_path%/%original_file_name%"
```
magul: Assuming that we will leave this line here translation of `/en/beginning/start.md` will be saved in `/pl/en/beginning/start.md`.

My change will save translation in `/pl/beginning/start.md` (which is compatible with structure of Github repo).

aniav:
The topic of translations and building source from crowdin is a bit of a pain for us for a long time but we didn't find anything that would be better than the current process. Thanks for the changes, I am merging them :)

PR #1041, opened 2017-05-20
Use Django 1.10 queryset repr

ekohl:
The above screenshot (images/step1.png) still needs to be updated, but I don't have an env right now.


aniav:
Thanks! :)

PR #1043, opened 2017-05-21
[English] Fix: Templates path

rofi93:
Defined path for templates in settings.py

### DIFF
--- a/en/html/README.md
+++ b/en/html/README.md
@@ -28,6 +28,16 @@ blog
 
 And now create a `post_list.html` file (just leave it blank for now) inside the `blog/templates/blog` directory.
 
+Now open `mysite/settings.py` and find the line that contains `TEMPLATES`. You need to define path in `DIRS: []` to find `templates`.
+For example:
+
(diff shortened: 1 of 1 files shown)


aniav:
I am not sure do we actually need to add this? If you have the default setting
```
'APP_DIRS': True,
```
in `TEMPLATES` you don't need to add any paths here.

rofi93:
@aniav It gives `TemplateDoesNotExist` exception if we don't add the path.

ekohl:
Where are you creating the template? The tutorial suggests ```blog/templates/blog/<file>.html``` so the whole ```BASE_DIR/templates``` should not be used if the instructions are followed.

rofi93:
@ekohl I was creating in project directory - `blog/templates/blog/<file.html>`.

aniav:
The tutorial shows the whole path of where to put the templates, they should be in the corresponding app directory, not the whole project directory. This works without adding any paths and we would like to keep it this way and not introduce more complications and additional steps 👣

PR #1046, opened 2017-05-25
Remove instructions for old Fedora versions

ekohl:
According to https://fedoraproject.org/wiki/End_of_life Fedora 21 has been EOL since 2015-12-01. We can expect that users have upgraded by now or know to use yum instead of dnf.

Some languages still have instructions for older python versions and they're not likely to work anymore but I decided to leave that in for now since I don't speak those languages.

### DIFF
--- a/cs/python_installation/instructions.md
+++ b/cs/python_installation/instructions.md
@@ -29,15 +29,7 @@ Použij tento příkaz v konzoli:
 sudo apt-get install python3.4
 ```
 
-#### Fedora (do verze 21)
-
-Použij tento příkaz v konzoli:
-
-```
-sudo yum install python3.4
-```
-
-#### Fedora (22 +)
+#### Fedora
 
 Použij tento příkaz v konzoli:
 
--- a/en/deploy/install_git.md
+++ b/en/deploy/install_git.md
@@ -28,17 +28,7 @@ $ sudo apt-get install git
 
 <!--endsec-->
 
-<!--sec data-title="Fedora (up to 21)" data-id="git_install_fedora_21"
-data-collapse=true ces-->
-
-{% filename %}command-line{% endfilename %}
-```bash
-$ sudo yum install git
-```
-
-<!--endsec-->
-
-<!--sec data-title="Fedora 22+" data-id="git_install_fedora_22"
+<!--sec data-title="Fedora" data-id="git_install_fedora"
 data-collapse=true ces-->
 
 {% filename %}command-line{% endfilename %}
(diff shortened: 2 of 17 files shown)


aniav:
We can't imply people with older Fedora versions know how to do things. From my experience it's exactly the other way round. Could You add a comment about older versions of Fedora using `yum` instead of `dnf` at the firs occurance of Fedora instructions, just so we have it covered?

ekohl:
That sounds like a fair compromise. I'll rebase it as well to fix the merge conflict.

ekohl:
I can mention it in English, but what should I do with other languages?

aniav:
Not sure, I would say: let's leave the instructions as they were if we aren't able to fix them properly.

ekohl:
Still, I really wonder if we should be giving instructions for versions that old. I think it's unlikely there are Django Girls attendees with such an old Fedora version. As a Fedora user I will admit that it's usage isn't that high save, especially among beginners. It does however take up screen real estate (like https://github.com/DjangoGirls/tutorial/issues/1058 mentions).

Right now I've rebased, fixed the merge conflict and added a commit with some warning for EN. Do you want me to roll back the changes for other languages?

aniav:
I would like to agree but I have seen people with computers borrowed from someone / not used frequently / with Windows XP... You have to remember the tutorial is used all around the globe and there might be cases where this one line helps. In perfect world we could just remove this, but I would still not vote on complete cleanup without even a note. We should also remember that the tutorial is used by coaches, who frequently need to help people using different OS than they use on a daily basis. I would have no idea what package manager does Fedora use and the instructions would help :)

ekohl:
In Fedora 22 dnf replaced yum. https://www.linux.com/learn/what-you-need-know-about-fedoras-switch-yum-dnf tells you why in case you're interested. /usr/bin/yum is now provided by ```dnf-yum``` and is a redirect to dnf:
```
$ sudo yum install git
[sudo] password for ekohl: 
Redirecting to '/usr/bin/dnf install git' (see 'man yum2dnf')

Last metadata expiration check: 2:12:31 ago on Wed Jun  7 22:09:03 2017.
Package git-2.9.4-1.fc25.x86_64 is already installed, skipping.
Dependencies resolved.
Nothing to do.
Complete!
```

If you want I can roll back the changes for translations and only fix EN, leaving it up to the translators to do the same for their language.

aniav:
Maybe someone else could also say something about this topic? :)

aniav:
Merging this because of lack of other opinions. Thank you! 🎉

PR #1050, opened 2017-05-28
Fixed Django version on PythonAnywhere and reviewed some typos

ayr-ton:
On the PythonAnywhere step on "pt" version, there's a `pip install` install step that doesn't mention the same version recommended on the django install step. 
So at the django install step I've added a `pip freeze` and on deploying I've fixed this issue with a `pip install -r requirements.txt`.

Also, I have fixed a small typo on the forms step.

### DIFF
--- a/pt/deploy/README.md
+++ b/pt/deploy/README.md
@@ -59,6 +59,7 @@ Git irá controlar as alterações para todos os arquivos e pastas neste diretó
     __pycache__
     myvenv
     db.sqlite3
+    static/
     .DS_Store
 
 
@@ -193,13 +194,11 @@ Assim como fez em seu próprio computador, você pode criar um virtualenv na Pyt
 
     20:20 ~ $ source myvenv/bin/activate
 
-    (mvenv)20:20 ~ $  pip install django whitenoise
+    (mvenv)20:20 ~ $  pip install -r requirements.txt
     Collecting django
     [...]
     Successfully installed django-1.8.5 whitenoise-2.0
 
-
-<!--TODO: think about using requirements.txt instead of pip install.-->
 
 ### Coleta de arquivos estáticos.
 
--- a/pt/django_installation/README.md
+++ b/pt/django_installation/README.md
@@ -93,13 +93,15 @@ Ok, nós temos todas as dependências importantes no lugar. Finalmente podemos i
 
 ## Instalando o Django
 
-Agora que você tem a sua `virtualenv` iniciado, você pode instalar Django usando `pip`. No console, execute `pip install django==1.8.5` (Perceba que usamos um duplo sinal de igual: `==`).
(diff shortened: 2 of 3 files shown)


ayr-ton:
@ekohl There is at https://github.com/ayr-ton/tutorial-djangogirls/blob/712efcb27a0a565baf733efb09b971228e305764/pt/django_installation/README.md 

I could provide a new pull request with updated steps to 1.10 like the English version, but I would like to test it first. Is that possible? For now I would like to ensure the 1.8 is fully working first. What do you think? x)

pt/deploy/README.md
```diff
@@ -193,14 +194,12 @@ Assim como fez em seu próprio computador, você pode criar um virtualenv na Pyt
 
     20:20 ~ $ source myvenv/bin/activate
 
-    (mvenv)20:20 ~ $  pip install django whitenoise
+    (mvenv)20:20 ~ $  pip install -r requirements.txt
```
scorphus: I might be missing something. But I'm willing to ask, where is the `requirements.txt` file? I believe that, at this point, there ain't none. It's yet to be created at some further step described below. What am I missing?

pt/deploy/README.md
```diff
@@ -59,6 +59,7 @@ Git irá controlar as alterações para todos os arquivos e pastas neste diretó
     __pycache__
     myvenv
     db.sqlite3
+    static/
```
scorphus: Is it always the case to completely ignore the `static` directory?

pt/deploy/README.md
ekohl: There's a pip freeze in this translation, but that's completely missing in the EN version. In fact, in the EN version whitenoise isn't even used. Most likely because it's updated to 1.10.

ayr-ton: @scorphus There's a freeze on this translation. On the next pull requests, we should start the translation to fit the en version using the 1.10 version.

pt/deploy/README.md
ayr-ton: @scorphus Yep. As it is autogenerated on the PythonAnywhere step, there was some users sending the static files accidentally while using git on the same time they where trying to send the css. What do you think?

ayr-ton: Any thoughts?

pt/django_installation/README.md
```diff
@@ -93,13 +93,15 @@ Ok, nós temos todas as dependências importantes no lugar. Finalmente podemos i
...
-    Downloading/unpacking django==1.8.5
-    Installing collected packages: django
-    Successfully installed django
+    (myvenv) ~$ pip install django==1.8.5 whitenoise==2.0
+    Downloading/unpacking django==1.8.5 whitenoise==2.0
+    Collecting 
+    Installing collected packages: django, whitenoise
     Cleaning up...
+
+    (myvenv) ~$ pip freeze > requirements.txt
```
ayr-ton: The freeze I had talked.

aniav:
Thanks! 👍

PR #1051, opened 2017-05-31
Add sudo apt-get update command

charlesdwesley:
On the latest Cloud 9 VM, an update is required before `python3.5-venv` will install.


charlesdwesley:
Probably these should both be `apt` commands, not `apt-get`.

PR #1063, opened 2017-06-06
Update hungarian translation due to changes in english tutorial

abeld:
update up to 'Deploy!' chapter, except python and django installation chapters

(note: updates of other chapters still forthcoming)


RachellCalhoun:
I checked it out in gitbook locally and it looks good!

PR #1069, opened 2017-06-11
[English] Allow copy paste WSGI

ekohl:
By using os.path.expanduser we automatically find the correct directory. This means the user doesn't have to modify anything making it easier. There is also no added value in doing this manually.


aniav:
Thanks! :)

PR #1079, opened 2017-06-22
Update README.md

camelidus:
When we add later the urls in templates we use the url name and that makes a noreversematch error because it´s not defined in urls.py. Thxs!


aniav:
Thanks! :)

PR #1080, opened 2017-06-24
Add Python path change to Windows install instructions

charlesdwesley:
Because we use the "C:\Python35\python" absolute path when making the virtual environment, we should make sure that this is the path where Python is installed for Windows. (On Windows 10 the absolute path is buried deep inside "AppData" in the user's home directory.)


aniav:
This has already been fixed in another PR and we actually don't want tu set path like `C:/Python3x`, we want to use `python` or `python3` after install, so putting it in the python PATH is enough! :)

PR #1082, opened 2017-06-24
[Turkish] PostForm() type correction

ozgurturkiye:
PostForm() is a class and we don't need to say anything about PostForm() type.


aniav:
Thankss! :))

PR #1086, opened 2017-07-09
Updated python and django versions

pauloxnet:


en/python_installation/instructions.md
```diff
@@ -2,24 +2,24 @@
...
 
 ![Type "cmd" in the "Run" window](../python_installation/images/windows-plus-r.png)
 
-Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python 3.5.x installer fails with an error, you can try either:
-1. install all Windows Updates and try to install Python 3.5 again; or
-2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [3.4.4](https://www.python.org/downloads/release/python-344/).
+Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python 3.6.x installer fails with an error, you can try either:
+1. install all Windows Updates and try to install Python 3.6 again; or
+2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [3.6.1](https://www.python.org/downloads/release/python-361/).
```
helenst: This line shouldn't be updated - it should be left alone (or possibly updated to reference 3.5.x - I'm not sure of the state of things on Windows)

en/python_installation/instructions.md
```diff
@@ -2,24 +2,24 @@
...
 
 
 <!--sec data-title="Windows" data-id="python_windows" data-collapse=true ces-->
 
 First check whether your computer is running a 32-bit version or a 64-bit version of Windows, by pressing the Windows key + Pause/Break key which will open your System info, and look at the "System type" line. You can download Python for Windows from the website https://www.python.org/downloads/windows/. Click on the "Latest Python 3 Release - Python x.x.x" link. If your computer is running a **64-bit** version of Windows, download the **Windows x86-64 executable installer**. Otherwise, download the **Windows x86 executable installer**. After downloading the installer, you should run it (double-click on it) and follow the instructions there.
 
-One thing to watch out for: During the installation you will notice a window marked "Setup". Make sure you tick the "Add Python 3.5 to PATH" checkbox and click on "Install Now", as shown here:
+One thing to watch out for: During the installation you will notice a window marked "Setup". Make sure you tick the "Add Python 3.6 to PATH" checkbox and click on "Install Now", as shown here:
```
helenst: There are screenshots that go along with the Windows install instructions that prominently display the 3.5.1 version number - these should probably be updated to avoid confusion in this section. (Are the instructions exactly the same for 3.6 installer? Can we verify that?)

helenst:
Hi @pauloxnet - thank you for contributing! :sparkles: 

I ran through this on Ubuntu and it worked - nothing major in 1.11 that affects the tutorial! :+1: I also tested the python install on cloud9 and it looks like pythonanywhere supports python3.6 too.

However I have a couple of concerns about the Windows version of Python 3.6 - it will need updated screenshots. The "old Python" screenshot of 3.4.x could be kept if we still expect people to use that. If 3.5 and 3.6 installers also look different from each other, the 3.5 could be used as another "old" screenshot, otherwise it could be simply replaced. I'm not sure where things are at with windows, so would appreciate input from somebody who knows better than me (and has a windows box to check this stuff!)

pauloxnet:
Hi @helenst 

I've updated python version to version 3.5.3 for old version and I've replaced images for python installation as requested.

I hope it will be ok.

helenst:
Hi @pauloxnet thank you for making those updates!

I think it would be easier to leave the 3.4.3 installer screenshot in place, since the text beside it talks about the dialog box looking different, and scrolling down to the right option. If we have a 3.5.1 screenshot there it won't match the text (and changing the text will make it harder for us to update translations at the same time)

I think it's fine to keep the reference to 3.4 installer there as it's probably still valid to use that (I don't think there's any code that requires later versions) and that's the one that actually looks different. We don't need a different screenshot for 3.5.1 installer as it looks very similar to 3.6.1.

Does that make sense?

pauloxnet:
Hi @helenst thanks to helping me with this update!

I thinks it's better keep the reference to 3.4 installer as you suggested.

I reverted the image to 3.4 version and replaced python 3.5 downlaod link with 3.4

aniav:
This was a lot of work! Thank you! 💃

pauloxnet:
It was a pleasure to contribute to the DjangoGirls project.

PR #1088, opened 2017-07-22
[Portuguese] Improves the language and visual guides for each OS.

rsip22:
#1088 

Improves the language to a more inclusive one, assuming that the reader won't be a male and admitting that the mentor helping them might be either a male or a female (as opposed as it is now, where mentors are always identified as males).

I'm experimenting with a new tag, that's already used in the English tutorial, to collapse instructions for each OS.

I've also changed the Python installation instructions to get it more up to date. I've chosen to keep Python 3.4 as the version to be used because it's the latest one can get on Ubuntu as of now and I believe it wouldn't be ideal if Windows and Linux users used different Python versions during the workshop.

### DIFF
--- a/pt/README.md
+++ b/pt/README.md
@@ -49,25 +49,25 @@ Wow! <3 <3
 
 ## Introdução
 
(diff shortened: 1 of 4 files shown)


rsip22:
@annanda Would you mind reviewing these changes, since I was the one who made them?

Thanks!

aniav:
@rsip22 as you are one of the people behind reviewing translations and because this is long awaiting I am going to merge it now. This is massive! Thanks ;)

PR #1091, opened 2017-08-17
Add image for clarification in PythonAnywhere console instructions

lmdragun:
The PythonAnywhere directions for finding the bash console are not as clear as some other parts (as users would not necessarily know what Bash is), so this adds an image for sighted users.


aniav:
Thanks youu! 💯

Issue #986, opened 2017-02-17
Use {% endblock %} instead of {% endblock <blockname> %} everywhere


das-g:
### History

The `endblock` Django template tag takes an optional argument: The block name, which must match the corresponding block name passed to the `block` tag that started the block. See [Template inheritance documentation](https://docs.djangoproject.com/en/1.10/ref/templates/language/#template-inheritance):

> For extra readability, you can optionally give a name to your `{% endblock %}` tag. For example:
> 
> ```html+django
> {% block content %}
> ...
> {% endblock content %}
> ```
> 
> In larger templates, this technique helps you see which `{% block %}` tags are being closed.

This has been used inconsistently in the tutorial. https://github.com/DjangoGirls/tutorial/issues/500#issuecomment-151468004 decided to get rid of it, but corresponding commits e97da025409dee4a88f8169a5d1663481aad6a9d and eb02ca9f958168da46ccef98f59f0b6f0ba85596 only partially removed it.

If I understand https://github.com/DjangoGirls/tutorial/pull/534#issuecomment-156489924 and https://github.com/DjangoGirls/tutorial/pull/534#issuecomment-161599097 correctly, it was then decided to use the optional argument everywhere or at least consistently.

### Status quo

The current instructions are still inconsistent, though:
In https://tutorial.djangogirls.org/en/template_extending/#create-a-base-template, the [end of the empty block](https://github.com/DjangoGirls/tutorial/blame/36b2c48601dd2104da47d8c4a055e7432f12e4cb/en/template_extending/README.md#L70) in the new `base.html` lacks the block name argument. For the stripped-down `post_list.html`, the [instructions](https://github.com/DjangoGirls/tutorial/blame/36b2c48601dd2104da47d8c4a055e7432f12e4cb/en/template_extending/README.md#L105) tell to
> put everything between `{% block content %}` and `{% endblock content %}`

but the result of that is [presented](https://github.com/DjangoGirls/tutorial/blame/36b2c48601dd2104da47d8c4a055e7432f12e4cb/en/template_extending/README.md#L119) to be

> ```html+django
> {% block content %}
>     {% for post in posts %}
>         <div class="post">
>             <div class="date">
>                 {{ post.published_date }}
>             </div>
>             <h1><a href="">{{ post.title }}</a></h1>
>             <p>{{ post.text|linebreaksbr }}</p>
>         </div>
>     {% endfor %}
> {% endblock %}
> ```

[...]

PR #1092, opened 2017-08-19
removed <blockname> #986

tugce:


ekohl:
I don't think this is the correct implementation for https://github.com/DjangoGirls/tutorial/issues/986. You are now removing it from the opening block too but I think that should stay. So ```{% block content %}``` is still needed but ```{% endblock content %}``` can be ```{% endblock %}```.

ekohl:
There are still some unrelated changes that look like a too global search-and-replace. Could you have a look at those as well?

aniav:
This definitelly neeeds fixes as now it would break the build of gitbook :)

aniav:
I have resolved #986 in another commit as this PR is too broad and hanging for a long time. I am closing that now. Thank you! 👍

PR #1093, opened 2017-08-19
Add www to front of gettingstartedwithdjango url

bri616:
This link was broken without the www, this fixed the issue.


aniav:
Thank you! 🎉

PR #1094, opened 2017-08-23
[Portuguese] Replace 'Ola' with 'Ana' in the examples of name

jkastrupb:
'Ana' is a more common name than 'Ola' in Brazil, and it avoids confusion with the portuguese word for Hello, which is 'Olá'.


aniav:
Thanks! 👍

PR #1096, opened 2017-08-29
Update idiom

dangry:
Though the idiom is "Comparing apples and oranges" when two items cannot be practically compared in Spanish pears are used instead of oranges.


aniav:
Thanks! :)

Issue #1072, opened 2017-06-17
Chromebook setup link broken on "Code editor" step


no1joel:
If you click the Chromebook setup link on the [Code editor](https://tutorial.djangogirls.org/en/code_editor/) step, nothing happens, and if you open it in a new tab you get some template-looking code.

aniav:
Fixed! :)

Issue #1071, opened 2017-06-17
Chromebook setup link broken on "Python installation" step


no1joel:
If you click the "chromebook setup" link on [Python installation](https://tutorial.djangogirls.org/en/python_installation/) nothing happens - if you open it in a new window, you get some funky template looking code.

aniav:
Fixed! :)

PR #1101, opened 2017-09-01
Fix: links to Chromebook Setup

lisac:
As noted in Issues #1071 and #1072 , the markdown links to the Chromebook Setup page fail to load that page.

While the embedded instance works fine (visible within the tutorial's [Installation](https://tutorial.djangogirls.org/en/installation/#chromebook-setup-if-youre-using-one)), it may be that in order for the standalone page to be accessible, it needs to be accounted for in the table of contents. That's my understanding from https://github.com/GitbookIO/feedback/issues/467. 

This PR adds the chromebook setup to the table of contents for the EN and KO versions (the only ones with chromebook content).

### DIFF
--- a/en/SUMMARY.md
+++ b/en/SUMMARY.md
@@ -2,6 +2,7 @@
 
 * [Introduction](README.md)
 * [Installation](installation/README.md)
+* [Installation (chromebook)](chromebook_setup/README.md)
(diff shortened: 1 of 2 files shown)


aniav:
Thank you for tracking this down! 🎉

PR #1105, opened 2017-09-09
Pl win fixes

qbolec:
While preparing to workshops in Wrocław, Poland, I went trough whole tutorial on Windows 10, and found some places which could be phrased better to minimize friction.
The most important one is probably the issue with `ALLOWED_HOSTS` in settings.py, which perhaps should be reflected also in other language versions.
Other than that:
- I've fixed the keyboard shortcut for exiting from Python 3.4 on Windows 8+ according to 
https://stackoverflow.com/a/41524896/575699 and my experience.
- added some rough explanation of why we use `virtualenv` on PythonAnywhere but `venv` on localhost, based on https://www.reddit.com/r/learnpython/comments/4hsudz/pyvenv_vs_virtualenv/
- replaced some of uses of word "method" with "function" to match the distinction between the two introduced in the Introduction to Python and common consensus https://stackoverflow.com/questions/155609/difference-between-a-method-and-a-function (there are still many occurences of "method" in other places of tutorial, but I've changed only these which were in the paragrahps I had to change anyway for other reasons)
- explained that the warning/promt from collectstatic will probably not occur on the first call (or rather: explained that the promt will be shown on subsequent calls)
- fixed the part which said that "the post_details returns render" to "the post_details returns the result of render"
- changed `<QuerySet [<Post: My second post>, <Post: My first post>]>` to the actually shown `[<Post: My second post>, <Post: My first post>]` 
- fixed explanation for the `url` Django tag, moving it a bit further, so we have required background to actually explain what all of the parts of it mean

### DIFF
--- a/pl/deploy/README.md
+++ b/pl/deploy/README.md
@@ -197,7 +197,7 @@ Na tę chwilę musimy tylko uruchomić polecenie `collectstatic` na serwerze. Po
 
     (myvenv) $ python manage.py collectstatic
 
-Będziemy z powżyszego polecenia korzystać za każdym razem gdy dodamy nowe pliki statyczne. Przy kolejnych uruchomieniach tego polecenia, najprawdopodobniej zobaczysz poniższe ostrzeżenie:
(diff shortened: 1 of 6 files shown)


aniav:
Thank you! Should be auto-deployed in a few minutes 🎉

PR #1109, opened 2017-09-16
[Polish] Fix apostrophe in commit message

Agalin:
Using apostrophe as commit message enclosing symbol doesn't work on Windows. Replaced with ".


aniav:
Thanks! 🎉

PR #1113, opened 2017-09-19
Change python3 --version to python --version

phoenixfangor:
The command to find version on windows is python --version NOT python3 --version


aniav:
@phoenixfangor this will be different depending if you have python 2 installed in the system and it will also depend on linux distribution, also `python3` should work in case you only have Python 3 or both Python 2 and 3. I can't accept that solution and I am closing that now 🙏

ekohl:
@aniav is correct and I shouldn't have approved.

phoenixfangor:
I'm on windows, and I didn't see your windows note. Thank you!

Issue #1058, opened 2017-06-06
Dropdown titles occupy too much space


pts:
The vertical whitespace occupied by the system-specific dropdown (e.g. *Windows*) headers is too much. On a small screen (1280x768), sometimes the entire useful area of the web page is occupied by these dropdowns, taking up valuable space from the useful information above and below.

My recommendation: Remove most of the vertical whitespace, and make the (closed) drop-downs vertically much smaller: make them 20% of their current height, removing 80%.

aniav:
This is now fixed :)

PR #1124, opened 2017-09-30
changed sectionx setting to make dropdowns neater

Frimkron:
The `sectionx` plugin now has a `tag` setting that defines the tag to use for the dropdown headers. Changing it to `b` tags means they take up less space and look a bit neater.

Addresses #1058


aniav:
This is a neat improvement! Thank you! :)

Issue #1089, opened 2017-07-27
Duplicate IDs in Installation page


funkybob:
Two headers have the same ID, and this the same fragment link:

https://tutorial.djangogirls.org/en/installation/#linux-and-os-x

This makes it impossible to give someone a link to the drop-down part about _activating_ their venv.

aniav:
This is now fixed! :)

PR #1126, opened 2017-09-30
Added unique prefixes to each os-depedent section so that heading lin…

Frimkron:
Quick fix for #1089 . Heading anchor ids for dropdown sections are based on the titles, which weren't unique, so linking to them didn't work. You can reference the `div` id defined by the `data-id` property, but the permalinks that appear next to the headings unfortunately don't use this.

Prefixing each dropdown title with the page section title was the least confusing way I could think of to fix this. I've only updated English so far, but if approved I'll update the other languages.


aniav:
Thanks! :)

PR #1131, opened 2017-10-04
[French] Remove references to collecting static files for deploy

creynaud:
The French version of the deploy step was still mentionning the fact that static files should be collected, which is not the case anymore since django.contrib.staticfiles.handlers.StaticFilesHandler is used in wsgi.py.


aniav:
I allowed myself to approve without French reviews :)

creynaud:
@aniav thanks for the review ;)!

PR #1134, opened 2017-10-09
Update portuguese typos and fixed version

priromagnoli:
Changes in this pull request:

-
-
-


pt/deploy/README.md
```diff
@@ -59,7 +59,7 @@ SQLite. De qualquer forma, ignorar seu banco de dados SQLite na cópia no GitHub
 criou até agora vão ficar disponíceis apenas localmente, e você terá que adicioná-las novamente em produção. Você deve pensar no seu banco de dados local como um bom "playground", onde você pode testar coisas diferentes e não ter medo de apagar as
```
GabrielCzar: Corrigir a palavra "disponíceis"

pt/deploy/README.md
```diff
@@ -188,8 +188,8 @@ Assim como fez em seu próprio computador, você pode criar um virtualenv na Pyt
     {% filename %}PythonAnywhere command-line{% endfilename %}
     20:20 ~ $ cd my-first-blog
 
-    20:20 ~ $ virtualenv --python=python3.6 myvenv
-    Running virtualenv with interpreter /usr/bin/python3.6
+    20:20 ~ $ virtualenv --python=python3.5 myvenv
+    Running virtualenv with interpreter /usr/bin/python3.5
```
aniav: We probably don't want this change

pt/deploy/README.md
rsip22: It was already wrong!

pt/deploy/README.md
rsip22: @priromagnoli Indeed, please change it back.

rsip22: There is another PR that corrects this, so I'll merge this one and the other afterwards so it will be fixed.

rsip22: Actually, @aniav, now I get it.

The install instructions in portuguese are in Python 3.5, so this makes sense (unless we change that instead).

priromagnoli: @rsip22 Yes... This was the reason why I've changed it, Python version is defined as 3.5 in PythonAnywhere. I'm sorry for not explaining this before!

PR #1137, opened 2017-10-09
Change Python version while creating venv on PythonAnywhere

biancarosa:
Changes in this pull request:

- Change instructions to install python3.5 on pythonanywhere instead of python 3.6 to make it coherent with the install instructions.

### DIFF
--- a/pt/deploy/README.md
+++ b/pt/deploy/README.md
@@ -188,8 +188,8 @@ Assim como fez em seu próprio computador, você pode criar um virtualenv na Pyt
     {% filename %}PythonAnywhere command-line{% endfilename %}
     20:20 ~ $ cd my-first-blog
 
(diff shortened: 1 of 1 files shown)


aniav:
is this actually ok? why are we moving to an older python version?

biancarosa:
Hi @aniav, actually it was just a PR to quickly fix a problem that was ocurring during Django Girls Belo Horizonte.

The best solution is to move to Python 3.6, but we have to review everything. The tutorial asks to install Python 3.5, then in the middle it asks to create a virtualenv on PythonAnywhere with Python 3.6 and it made everyone make mistakes during the workshop.

aniav:
I can understasnd that but yesterday we asked in another PR for this language to change to python 3.6 in install chapters instead of python 3.5. So I am a but confused, how many versions of python do you have in the tutorial now?

PR #1139, opened 2017-10-11
Added VS Code to the Code Editor list

izabelacborges:
Changes in this pull request:

- Added VS Code to Code Editor list
- Fixed bold formatting on 'texto sem formatação'


rsip22:
Can you say if do we want yet another editor, @aniav ? In particular one with Microsoft behind it?

If so, this PR is good to be merged.

aniav:
We have nothing agains Microsoft, and I do use VSCode ona daily basis, it's a very good IDE :)

PR #1150, opened 2017-10-15
Clarify simplest view example

emazzotta:
It's not true that the "simplest" *view* can look like this, because the import will still be needed. I am aware that it is mentioned before but one of the girls in my group removed it, since that's apparently the simplest the *view* can look like :).

Changes in this pull request:

- Add required import to view example


emazzotta:
@ekohl thanks, two very fair points. I've changed it accordingly

aniav:
Makes sense! :)

Issue #1149, opened 2017-10-15
[German] Fix bash commands in 'Einführung in HTML'


KatuBeltz:
### Issue description
In the end of the chapter 'Einführung in HTML' there are several bash commands displayed in one row:

'bash $ cd ~/my-first-blog $ source myvenv/bin/activate (myvenv)$ git pull [...] (myvenv)$ python manage.py collectstatic [...]bash'

### Language
German

### Operating system
All

PR #1151, opened 2017-10-15
fix the code lines at the end

joannakl:
should fix the issue https://github.com/DjangoGirls/tutorial/issues/1149

Changes in this pull request:

this should change replaces the broken lines with the matching lines from the English version.


aniav:
👍

PR #1152, opened 2017-10-16
fixed error message of TypeError for unorderable types

pts:
Old error message is by Python 3.5, new by Python 3.6.

The tutorial now instructs to install Python 3.6, so it should also show the corresponding error messages.


aniav:
Good catch! Thanks :)

PR #1143, opened 2017-10-13
Migrate Chromebook instructions to CodeAnywhere

vilmibm:
Changes in this pull request:

- Switch Chromebook setup from Cloud 9 to CodeAnywhere
- Removes Cloud 9 specific workarounds (SessionStorage, runserver port)
- Adds a few additional details about Chromebooks

This addresses issues #1081 #1143 and #1075 and #1073 .

PR #1162, opened 2017-11-04
Simplify ALLOWED_HOSTS configuration

ekohl:
998120bbbdbdc8907f3e1178aadf416b022de5b3 moved from .pythonanywhere.com to <your_username>.pythonanywhere.com to fix a DisallowedHost error. I can't reproduce that and this is much easier because it allows copy-pasting.

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -107,13 +107,13 @@ match our hostname on PythonAnywhere once we deploy our application so we will c
 
 {% filename %}mysite/settings.py{% endfilename %}
 ```python
-ALLOWED_HOSTS = ['127.0.0.1', '<your_username>.pythonanywhere.com']
+ALLOWED_HOSTS = ['127.0.0.1', '.pythonanywhere.com']
 ```
 
 > **Note**: If you're using a Chromebook, add this line at the bottom of your settings.py file:
 > `MESSAGE_STORAGE = 'django.contrib.messages.storage.session.SessionStorage'`
 
-> Also add `django-girls-<your_username>.c9users.io` to the ALLOWED_HOSTS if you are using cloud9
+> Also add `.c9users.io` to the `ALLOWED_HOSTS` if you are using cloud9
 
 ## Set up a database
 


ekohl:
https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts also still documents this should work.

@gsimore can you have a look?

thibaudcolas:
I've seen issues with this multiple times – attendees either copy paste and forget to replace the placeholder, or handwrite the `127.0.0.1` IP address and mistype it.

Wouldn't it be even simpler to use `ALLOWED_HOSTS = ['*']` instead? I know it's not a good practice in terms of security but it would definitely save some time.

en/django_start_project/README.md
```diff
@@ -107,13 +107,13 @@ match our hostname on PythonAnywhere once we deploy our application so we will c
...
 ```python
-ALLOWED_HOSTS = ['127.0.0.1', '<your_username>.pythonanywhere.com']
+ALLOWED_HOSTS = ['127.0.0.1', '.pythonanywhere.com']
 ```
 
 > **Note**: If you're using a Chromebook, add this line at the bottom of your settings.py file:
 > `MESSAGE_STORAGE = 'django.contrib.messages.storage.session.SessionStorage'`
 
-> Also add `django-girls-<your_username>.c9users.io` to the ALLOWED_HOSTS if you are using cloud9
+> Also add `.c9users.io` to the `ALLOWED_HOSTS` if you are using cloud9
```
thibaudcolas: Just tested this with CodeAnywhere (#1143), works like a charm.

ekohl:
> Wouldn't it be even simpler to use ALLOWED_HOSTS = ['*'] instead? I know it's not a good practice in terms of security but it would definitely save some time.

The current example can be copy-pasted. We could even include ```.c9users.io``` to the example so there's no difference between the two. Fewer exceptions and less manual changes is definitely the way to go.

aniav:
Thank you! Also @ekohl I have added you write rights on this repo, as you have already done massive amounts of work and you definitely know the repo well. You can now merge pull requests and I encourage you to do so whenever you have no doubts about changes 🌟

aniav:
I can also invite you to slack, I don't see you there. I would need your e-mail, you can write it to me on ania at djangogirls.org :)

ekohl:
Thanks @aniav! I'm not on slack, already on plenty of IRC channels and I prefer not to use walled gardens but I'll keep an eye on this repo.

aniav:
Sure! 💯

PR #1171, opened 2017-11-13
Added 'on_delete' to 'ForeignKey' in Django models:

hivickylai:
Not providing `on_delete` will give deprecation warning in Django 1.11 - will be required in Django 2.0.
See https://docs.djangoproject.com/en/1.11/ref/models/fields/#foreignkey

Changes in this pull request:

- Added `on_delete` positional argument to ForeignKey
- May currently alleviate some confusion, and required in upcoming Django 2.0.

### DIFF
--- a/en/django_models/README.md
+++ b/en/django_models/README.md
@@ -121,7 +121,7 @@ from django.utils import timezone
 
 
 class Post(models.Model):
-    author = models.ForeignKey('auth.User')
+    author = models.ForeignKey('auth.User',on_delete=models.CASCADE)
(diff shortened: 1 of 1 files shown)


ekohl:
Is that new screenshot from Django 2.0? Doesn't look familiar to me and the current tutorial is using 1.11 so that may confuse students.

hivickylai:
Right, my bad. Reverted new screenshot addition.

aniav:
Thank you! 🎉

PR #1172, opened 2017-11-17
Suggest Ctrl+Alt+T to open a terminal on Linux

mgedmin:
I'm afraid I don't have a list of Linux distributions where this works, and my impression that it's *all of them* might be mistaken.  Still, it can save a few minutes of googling.

### DIFF
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -38,6 +38,8 @@ Go to Applications → Utilities → Terminal.
 
 It's probably under Applications → Accessories → Terminal, but that may depend on your system. If it's not there, just Google it. :)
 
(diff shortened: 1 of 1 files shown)


mgedmin:
It does work in Gnome on Ubuntu 17.10, but this may be a Ubuntu patch.  :/

ekohl:
This is Fedora 26 with Gnome 3.24. An upgrade to Fedora 27 is planned and I can report back when I've upgraded what Gnome 3.26 does there.

ekohl:
On Fedora 27 it doesn't work either so I'm guessing it's an Ubuntu patch.

aniav:
I wouldn't add it if we are not sure this will work 🙏

Issue #902, opened 2016-11-08
Automate publishing the tutorial


bittner:
Last Saturday at the [Django Under the Hood](https://djangounderthehood.com/) 2016 conference sprint I've started to look into automating the [publishing of this tutorial](https://translate.djangogirls.org/when_its_ready.html). When translations are made and successfully reviewed they should be published online without manual intervention (e.g. no manual downloading from Crowdin, no manual corrections of broken generated code, no copying images, no pull requests required, etc.).

Instead translators should focus on translating (on Crowdin) and translations should be automatically deployed as soon as they are approved (a.k.a. continuous translation). Technically, this would translate into an automatic build process triggered by Crowdin and executed (probably) on Travis-CI.

Current Challenges
----------------------

- Crowdin seems to restructure the Markdown code breaking some of the resulting articles.
- A description of what happens to deploy the tutorial after a PR is accepted is missing from the [publishing guide](https://translate.djangogirls.org/when_its_ready.html).
- A Crowdin project identifier is [under version control](https://github.com/DjangoGirls/tutorial/blob/master/crowdin.yaml), but it doesn't seem to work ("Requested project does not exist or API key is not valid" when I try with `crowdin-cli-py download`). 
- I have no appropriate access to the [Crowdin project](https://crowdin.com/project/django-girls-tutorial) to verify the current project identifier.
- I have no appropriate access to the [GitBook organization](https://www.gitbook.com/@djangogirls) to verify the current deployment settings.

Observations, Side Notes
----------------------------

- According to @patjouk Crowdin is used, because it supports Markdown (whereas Transifex doesn't).
- GitBook gives the impression to work similar to ReadTheDocs, but I've not yet managed to find out how exactly.
- ReadTheDocs and reStructuredText allow for an established automatic process for publishing documentation including their translations (e.g. using Transifex). That could be an alternative if we can't get the Markdown/Crowdin issues under control. 

Help and Feedback Needed
--------------------------

- I'd be interested in details about why it is important to use Markdown, GitBook and Crowdin.
- I'd need the appropriate details of the Crowdin and GitBook setup, or privileges to look those up myself. (@patjouk, I got in touch with you via Crowdin Sunday; I guess you're busy doing the wrap-up of the conference.)

Any support greatly appreciated!

bittner:
A newer, partially related issue: #1271

das-g:
> I'd be interested in details about why it is important to use Markdown, GitBook and Crowdin.

[...]

PR #1173, opened 2017-11-18
Integrated the tabs plugin

woeye:
Also updated en/python_installation/instructions.md in to show how to use the plugin.

Changes in this pull request:

-
-
-

### DIFF
--- a/book.json
+++ b/book.json
@@ -16,7 +16,8 @@
         "language-picker",
         "sidebar-ad",
         "codeblock-label",
-        "sectionx"
+        "sectionx",
+        "simpletabs"
     ],
     "pluginsConfig": {
         "ga": {
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -5,7 +5,9 @@
 Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install Python 3.6, so if you have any earlier version, you will need to upgrade it.
 
 
-<!--sec data-title="Install Python: Windows" data-id="python_windows" data-collapse=true ces-->
+{% tabs windows="Windows", osx="OS X", linux="Linux", debian="Debian or Ubuntu", fedora="Fedora", suse="OpenSUSE" %}
+
+{% content "windows" %}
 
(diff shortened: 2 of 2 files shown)


aniav:
This is great! I can't merge it for now, because we need to change all the files at once to have the simpletabs instead of collapsible panels, so we will not confuse the users :)

ekohl:
Could we revisit this and integrate it on every page?

aniav:
I think I need to close this. We are now running gitbook on the legacy version of the platform and the new version of gitbook is incompatible with plugins, so effectively in order to upgrade to new version of gitbook we would need to remove the code related to this.

Sorry it never got merged 😢

das-g:
> I think I need to close this. We are now running gitbook on the legacy version of the platform and the new version of gitbook is incompatible with plugins, so effectively in order to upgrade to new version of gitbook we would need to remove the code related to this.

Well, we are already using a plugin, albeit a different one. Are there any plans to switch to the new GitBook version in the foreseeable future?

aniav:
We are using the other plugin, but merging this PR would require additional work in the whole repository, so the plugin is updated in all places. Quite a big refactor itself.

As for plans to switch to a new version, I don't think there are any before someone decides to actually implement it 🤷‍♀

magul:
@das-g @aniav two things here:

 * That plugin can potentially solve the problem with Crowdin inability to properly render  sections using `<!--sec` syntax (see: https://github.com/DjangoGirls/tutorial/pull/1539#issuecomment-515680553 and #1541)
 * We have at least to issues, that try to describe how we're publishing newest tutorial versions: #1271 and #902. My understanding is that deliverable of #1271 should be a description available in  https://github.com/DjangoGirls/translation-guide that will probably increase translators engagement in the flow that we tend to use. The second one (#902) is for me a little too broad in scope. We should split it in my opinion into 3 separate issues: GitBook, Crowdin and Markdown. If we want to migrate from legacy to GitBook (or any other solution) I believe that need to be carefully planned to minimalize the impact in what we already have in translations.

Issue #1179, opened 2017-11-28
post_list view code doesn't match up between view lessons in English translation


tothebeat:
### Issue description

In the [first introduction of creating the post_list view](https://github.com/DjangoGirls/tutorial/blame/master/en/django_views/README.md#L28-L29), the code is as follows: 

```python
def post_list(request):
    return render(request, 'blog/post_list.html')
```

In the follow-up chapter on "Dynamic data in templates" that continues to develop this view, it incorrectly states the existing view code as being:

```python
def post_list(request):
    return render(request, 'blog/post_list.html', {})
```

The difference being that there's an empty dictionary supposedly passed as an argument here, which wasn't present in the first chapter. [Here's the location of that line](https://github.com/DjangoGirls/tutorial/blame/master/en/dynamic_data_in_templates/README.md#L16
).

### Language

This only affects the English translation. The original view code in [Spanish](https://github.com/DjangoGirls/tutorial/blame/master/es/django_views/README.md#L23), [Italian](https://github.com/DjangoGirls/tutorial/blame/master/it/django_views/README.md#L23), and [French](https://github.com/DjangoGirls/tutorial/blame/master/fr/django_views/README.md#L27) does have the dictionary. I checked other languages by cloning this repository and searching for `render(request, 'blog/post_list.html')` and finding it only in the English translation, while `render(request, 'blog/post_list.html', {})` is present in the same file for all other translations.

### Operating system

This is code related to all operating systems.

ekohl:
With #1180 merged I think this has been fixed.

PR #1180, opened 2017-11-28
[English]: Issue #1179 Match view code

tothebeat:
Changes in this pull request:

Add an argument of `{}` to the `render()` call in the `django_views` portion of the tutorial, to match up to the code referenced in the `dynamic_data_in_templates` portion.

### DIFF
--- a/en/django_views/README.md
+++ b/en/django_views/README.md
@@ -26,7 +26,7 @@ Let's create a *view* as the comment suggests. Add the following minimal view be
 {% filename %}blog/views.py{% endfilename %}
 ```python
 def post_list(request):
-    return render(request, 'blog/post_list.html')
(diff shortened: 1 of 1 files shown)


ekohl:
Shouldn't we simplify ```dynamic_data_in_templates``` instead and only pass in the context when we have context?

tothebeat:
@ekohl That sounds good, that just represents a larger change in all language versions rather than a change in one file for the English version.

ekohl:
@tothebeat it does, but I think the simplicity is worth it. Might be good to get one of the repo owners to state their opinion(s).

aniav:
I think it's ok having the `{}` added as we do fill the context later on in the code. If it would be jus this render used in the whole tutorial I would also suggest removing the empty context.

Thank you! 👍

Issue #729, opened 2016-04-01
Replace uses of "easy", "simple", and "fast"


phalt:
At DjangoCon Europ 2016, [@oboechick_](https://twitter.com/oboechick_) did a lightning talk on steps to make beginner tutorials easy. The steps are summarised in this blog post: 

https://www.wordfugue.com/how-make-beginner-workshops-more-beginner-friendly/

I think we could analyse the steps and see if we can improve the tutorial! This should be a friendly task for beginners.

asendecka:
I totalny agree! It is an amazing point.

ChristinaG86:
Hi, I'm going to work on this for 24 Pull Requests :)

aniav:
Done!

PR #1182, opened 2017-12-02
removed instances of the words 'easy', 'simple' and 'fast'

ChristinaG86:
Changes in this pull request:

-Addressing this issue - https://github.com/DjangoGirls/tutorial/issues/729

### DIFF
--- a/en/README.md
+++ b/en/README.md
@@ -21,7 +21,7 @@ We hope that we'll be able to make you love technology as much as we do!
 
 ## What will you learn during the tutorial?
 
-Once you've finished the tutorial, you will have a simple, working web application: your own blog. We will show you how to put it online, so others will see your work!
+Once you've finished the tutorial, you will have a working web application: your own blog. We will show you how to put it online, so others will see your work!
 
 It will (more or less) look like this:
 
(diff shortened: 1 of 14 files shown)


en/README.md
```diff
@@ -21,7 +21,7 @@ We hope that we'll be able to make you love technology as much as we do!
 
 ## What will you learn during the tutorial?
 
-Once you've finished the tutorial, you will have a simple, working web application: your own blog. We will show you how to put it online, so others will see your work!
+Once you've finished the tutorial, you will have a working web application: your own blog. We will show you how to put it online, so others will see your work!
```
aniav: I think in this case simple makes sense. We don't want simple when we want to say doing something is simple. While indicating that the website is simple is true.

en/django_start_project/README.md
```diff
@@ -7,7 +7,7 @@ tutorial](http://django-marcador.keimlink.de/) licensed under the Creative Commo
 Attribution-ShareAlike 4.0 International License. The django-marcador tutorial
 is copyrighted by Markus Zapke-Gründemann et al.
 
-We're going to create a simple blog!
+We're going to create a blog!
```
aniav: That's the same with this one.

en/README.md
ekohl: basic, minimal or small might be better? less implications about complexity.

aniav: Makes sense! Thank you :)

Issue #1165, opened 2017-11-04
Students miss the include import in URLs


ekohl:
### Issue description

In https://github.com/DjangoGirls/tutorial/blob/96176a0e17f44512dc60298c370b635a5d92f928/en/django_urls/README.md#your-first-django-url it does state the import is needed:

> Note that we are using the include function here so you will need to add that to the import on the first line of the file.

Yet this is a problem because today all 3 of my students forgot this include. I'd submit a PR if I knew how to reword this so it can't be missed.

### Language

All 3 students were using the English version.

### Operating system

Irrelevant, but Mac and Windows.

thibaudcolas:
I've seen this problem quite often as well. They miss the `include`, and generally only realise this at the "views" step where they are expected to get a `TemplateDoesNotExist` error, but get `NameError: name 'include' is not defined` instead.

It might help to put the import on a separate line to make it more visible?

```python
from django.conf.urls import url
from django.contrib import admin
from django.conf.urls import include

urlpatterns = [
    url(r'^admin/', admin.site.urls),
    url(r'', include('blog.urls')),
]
```

ekohl:
That's exactly where my students ran into it as well. I like your suggestion, the include in the middle of the line is easy to miss.

PR #1184, opened 2017-12-02
Make the import include clearer

ekohl:
A lot of my students miss this import. It's very subtle. This isn't as pretty but more obvious (as suggested by @thibaudcolas).

Closes GH-1165


aniav:
I reordered the imports so they are in alphabetical order. Merging when CI passes. Thank you! ⚡️

PR #1194, opened 2018-01-20
%locale% will properly handle crowdin-generated assets

magul:
We have setup dialect-grade translations on Crowdin, so we are not able to cover all of them using two letter language code - we need to use full locales.

### DIFF
--- a/crowdin.yaml
+++ b/crowdin.yaml
@@ -1,3 +1,3 @@
 files:
   - source: /en/**/*.md
-    translation: '**/%original_file_name%'
+    translation: '/%locale%/**/%original_file_name%'


aniav:
Shouldn't this be changed on crowdin? It will automatically update the file on the repo as well.

magul:
I'm closing this one. `**` did the trick 8 months ago, but now Crowdin is not accepting `**` for file destination pattern. Nor in web interface neither in Github integration (there are issues from yesterday sync: https://crowdin.com/project/django-girls-tutorial/settings#integration).

I would try to figure out other solution here.

PR #1202, opened 2018-01-27
Update PythonAnywhere screenshots to match current UI

zumwalt:
Hi team! This PR updates screenshots inside the "Deploy!" section of each language (`/{lang}/deploy/README.md`) to match PythonAnywhere's current UI. It also introduces a new image, `pythonanywhere_web_tab.png`, to account for the difficulty of including both PythonAnywhere's navigation *and* the Virtualenv settings within the same screenshot. 

As screenshots are managed individually by language, this PR touches a lot of files.

### `pythonanywhere_bash_console.png`
| Before  | After  |
|:-:|---|
| image  | pythonanywhere_bash_console |

### `pythonanywhere_web_tab_virtualenv.png`
| Before  | After  |
|:-:|---|
| image | image |

### `pythonanywhere_web_tab.png`
| Before  | After  |
|:-:|---|
| None | image |

Thanks for your time!


ekohl:
First of all: thanks for such a detailed PR and all the effort!

The consoles part is unchanged. You made a screenshot from the dashboard which is another way of getting there but doesn't match the text. If you remove the ```or "Consoles" page``` it will match but it might be good to mention that Consoles page since that's in the navigation.

Other than that it looks good to me.

zumwalt:
> The consoles part is unchanged. You made a screenshot from the dashboard which is another way of getting there but doesn't match the text. If you remove the or "Consoles" page it will match but it might be good to mention that Consoles page since that's in the navigation.

Ah. Great catch @ekohl! I've replaced that screenshot with an updated version of the original.

aniav:
Sorr you had to wait so long. Is this PR still valid after we changed how deployment on PA is done?

ekohl:
Looks like we waited for a while and now we have conflicts. I'm going to close this now but we'd welcome an updated PR if it's still needed.

PR #1208, opened 2018-02-07
specify on_delete=CASCADE in models.py for Django 2.0 compatibility

cf0af844a67a12276984d45b5fa92cbd:

PR #1212, opened 2018-02-11
Added 'on_delete' to 'ForeignKey' in Django models

anezkamll:
It caused error when deploying to PythonAnywhere


aniav:
This is a duplicate of #1208, closing 💃

ekohl:
This is another language so we should take care of changing it for every language.

aniav:
Yeah, ideally it should be in all languages at once

PR #1217, opened 2018-02-12
Add British flag to LANGS.md alongside the US flag

Kwpolska:
Rationale: if the North Korean flag is listed for Korean, even though residents of said country can’t access the tutorial anyway, then there is no reason why the British flag should not be listed for English, especially considering the country from which the English language originates.

### DIFF
--- a/LANGS.md
+++ b/LANGS.md
@@ -1,4 +1,4 @@
-* [🇺🇸 English](en/)
+* [🇬🇧/🇺🇸 English](en/)
 * [🇨🇿 Czech (beta)](cs/)
 * [🇩🇪 Deutsch (beta)](de/)
 * [🇪🇸 Español (beta)](es/)


ekohl:
I'm not against this, but there is a merge conflict now. Would you be willing to update it?

das-g:
I'm not against this either, but would like to ask whether the following aspects have been considered:

* The tutorial AFAIK uses US English orthography (e.g. "color" instead of "colour", and not just when it's a CSS keyword)
* There are more countries than the USA and the UK where English is the official language, e.g. Australia. Should their flags be added, too?
* There are even more countries where English is _one_ of the official languages and in practice often the main official language. (Canada, India, Singapore and many more)
* There are yet even more countries where many people speak English natively.

ekohl:
> The tutorial AFAIK uses US English orthography (e.g. "color" instead of "colour", and not just when it's a CSS keyword)

This if for me a very strong argument. UK speakers might consider it typos if they see a UK flag while accepting it if they see a US flag.

Kwpolska:
> I'm not against this, but there is a merge conflict now. Would you be willing to update it?

I can’t, since I deleted my original repository.  And I honestly don’t care enough to do it again.

PR #1225, opened 2018-02-20
Include a warning icon before blocks that return errors

jacebrowning:
TODO:

- [x] Add tooltip
- [x] Apply to every page
- [ ] Apply to every language?

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -1,3 +1,5 @@
+{% set warning_icon = '<span class="glyphicon glyphicon-exclamation-sign" aria-hidden="true" style="color: red;"></span>' %}
+
 # Introduction to Python
 
 > Part of this chapter is based on tutorials by Geek Girls Carrots (https://github.com/ggcarrots/django-carrots).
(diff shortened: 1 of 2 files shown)


brondsem:
This looks like a good step forward to me.  Would it be possible to include title/hover text on the icon?  That might be nice if someone forgets what it means.

jacebrowning:
@brondsem Yeah, I can add that. Should this change be applied to every language before merging?

en/python_introduction/README.md
```diff
@@ -498,7 +502,7 @@ You can give Python as many numbers to compare as you want, and it will give you
 
 Have you heard of the expression "comparing apples to oranges"? Let's try the Python equivalent:
 
-{% filename %}command-line{% endfilename %}
+{% filename %}{{ warning_icon }} ccommand-line{% endfilename %}
```
brondsem: Typo "ccommand"

jacebrowning:
I fixed the typo.

Initially I included this icon on a few of the later sections, but I took it out because:

- the icon was only needed in one place (less to keep updated)
- the students might be comfortable with errors at that point

jacebrowning:
@rsip22 @ekohl @aniav What do you think about merging this? We have a workshop coming up in just over a week and I think this will really help people! 😄

RachellCalhoun:
@jacebrowning do you have a gitbook example up that you can share so we can see these changes?

jacebrowning:
This is what it looks like:

<a href="https://cl.ly/2y0L3O3l0F0b" target="_blank"></a>

PR #1227, opened 2018-02-21
Crowdin support fix for spanning one XPath into multiple translatable instances

magul:
This changes should fix an issue that occurs in Crowdin. Because of this space multiline code block is wronlgy parsed and span from https://crowdin.com/translate/django-girls-tutorial/296/en-pl#8808 up to https://crowdin.com/translate/django-girls-tutorial/296/en-pl#8816

Whole selection on screenshot is composed by one XPath selector - `/pre[4]/code`, I spoke with Crowdin support and they suggest, that this change will fix it.
multiple-parts-xpath


magul:
I'm merging it, as already translated Crowding assets are wrongly formated here.

PR #1228, opened 2018-02-21
fixed command line sections

luizsanches:
Changes in this pull request:

Fixes:

- section heading that was without dividing line
- two sections of linux commands were outside the area


luizsanches:
@rsip22, can you review this?

Issue #1224, opened 2018-02-20
Git on Windows Install Changes


brousch:
### Issue description

The installer for Git on Windows now has a step (5) where you have to pick a default editor for Git. We should have the user pick Nano as the editor (default is Vim, very unfriendly).

Adjusting the PATH is now step 6 in the wizard, so that should also be addressed.

https://github.com/DjangoGirls/tutorial/blob/master/en/deploy/install_git.md

### Operating system

Windows

PR #1236, opened 2018-03-07
Omit the step number in windows git installation wizard

brondsem:
The step name is sufficient, and more future-proof to changes.

This addresses part of #1224


ekohl:
Thanks @brondsem!

PR #1238, opened 2018-03-08
switch from apt-get command to apt

sevu:
How about changing apt-get to apt. Given that fedora uses one three letter command for installation it looks better.
It's included in Ubuntu since 14.04 and in Debian since Jessie.


ekohl:
Thanks!

PR #1273, opened 2018-05-30
fix typo in Chromebook Setup instructions

das-g:
"or" &rarr; "for"


magul:
... and as You can see it already landed on Crowdin dropin coverage: https://crowdin.com/translate/django-girls-tutorial/292/en-pl

PR #1274, opened 2018-06-01
Update pip command for all OS

anapaulagomes:
Changes in this pull request:

- Update `pip install` command to support all OS (it had a problem with the previous command on Windows).


ekohl:
Confirmed to work on my Fedora 27 with python 3.6.5. Thanks!

PR #1276, opened 2018-06-02
Make the indentation instructions more clear

### LINKED ISSUE #1257: Make the instructions about indentention more clear
### Issue description
In more than one workshop, we have noticed the participants not getting that the indentation _really matters_ and that **it should be consistent in all the code**. That the four spaces mentioned aren't just a suggestion, that if they don't use the same amount of spaces or tabs, the code will output an error.

I suggested we make this a bit more clear in the "Introduction to Python" section. 

### Language

This has been observed particulary with the Portuguese version, but I reviewed it and it's a direct translation from the original version, so I believe it could be improved in all the languages.

ktalik:
Fixes #1257

Changes in this pull request:

- Make the indentation instructions more clear
- Add note about text editor setup (it might be misleading to say Tab = 4 spaces on every text editor)

What do you guys think?

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -683,7 +683,7 @@ if 3 > 2:
     print('It works!')
 ```
 
-Notice how we've indented the next line of code by 4 spaces? We need to do this so Python knows what code to run if the result is true. You can do one space, but nearly all Python programmers do 4 to make things look neat. A single `tab` will also count as 4 spaces.
+Notice how we've indented the next line of code by 4 spaces? We need to do this so Python knows what code to run if the result is true. You can do one space, but nearly all Python programmers do 4 to make things look neat. A single Tab will also count as 4 spaces as long as your text editor is set to do so. When you made your choice, don't change it! If you already indented with 4 spaces, make any future indentation with 4 spaces, too - otherwise you may run into problems.
 
 Save it and give it another run:
 


en/python_introduction/README.md
```diff
@@ -683,7 +683,7 @@ if 3 > 2:
     print('It works!')
 ```
 
-Notice how we've indented the next line of code by 4 spaces? We need to do this so Python knows what code to run if the result is true. You can do one space, but nearly all Python programmers do 4 to make things look neat. A single `tab` will also count as 4 spaces.
+Notice how we've indented the next line of code by 4 spaces? We need to do this so Python knows what code to run if the result is true. You can do one space, but nearly all Python programmers do 4 to make things look neat. A single Tab will also count as 4 spaces as long as your text editor is set to do so. When you made your choice, don't change it! If you already indented with 4 spaces, make any future indentation with 4 spaces, too - otherwise you may run into problems.
```
ekohl: Not sure the 'as long long as your editor` part makes sense, but I think this does improve it.

das-g: > Not sure the 'as long long as your editor` part makes sense,

There's at least 3 aspects here:

1. Will the tab **key** insert a tab **character** or a number of space characters?
   - If the latter, how many?
2. If a tab character is inserted, how long will it appear in the editor?
3. If tabs and spaces are mixed, how will Python interpret them?

Dunno how much of that the instructions here should cover, though.

ekohl: That was my consideration as well. Too many technical details can be confusing and actually worse than the full story. Most important part is not mixing spaces and tabs and that was added now.

PR #1279, opened 2018-06-03
Move note about searching for "Command Prompt" to start-menu-based instructions

das-g:
Move note about searching for "Command Prompt" to between the instructions for getting it from the start menu and the Win-R "cmd"-instructions, as the searching probably refers to within the start menu (which is searchable in newer Windows versions and might offer some entries only through search).

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -13,7 +13,7 @@ One thing to watch out for: During the installation, you will notice a window ma
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
-In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu → Windows System → Command Prompt. (On newer versions of Windows, you might have to search for "Command Prompt" since it's sometimes hidden.) You can also hold in the Windows key and press the "R"-key until the "Run" window pops up. To open the Command Line, type "cmd" and press enter in the "Run" window.
+In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu and enter "Command Prompt" into the search field there. (On older versions of Windows, you can start the Command Line with Start menu → Windows System → Command Prompt.) You can also hold in the Windows key and press the "R"-key until the "Run" window pops up. To open the Command Line, type "cmd" and press enter in the "Run" window.
 
 ![Type "cmd" in the "Run" window](../python_installation/images/windows-plus-r.png)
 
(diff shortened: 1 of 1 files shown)


das-g:
closed by accident, re-opening

en/python_installation/instructions.md
```diff
@@ -13,7 +13,7 @@ One thing to watch out for: During the installation, you will notice a window ma
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
-In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu → Windows System → Command Prompt. You can also hold in the Windows key and press the "R"-key until the "Run" window pops up. To open the Command Line, type "cmd" and press enter in the "Run" window. (On newer versions of Windows, you might have to search for "Command Prompt" since it's sometimes hidden.)
+In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu → Windows System → Command Prompt. (On newer versions of Windows, you might have to search for "Command Prompt" since it's sometimes hidden.) You can also hold in the Windows key and press the "R"-key until the "Run" window pops up. To open the Command Line, type "cmd" and press enter in the "Run" window.
```
ekohl: Should we remove the parenthesis now?

das-g: I guess you mean remove the round braces around the sentence, not removing the parenthesized sentence itself, right? (Colloquially, "parenthesis" can refer to either.)

I don't know which Windows versions are affected and how common they are nowadays. I'd leave the sentence parenthesized for now, but if you'd like to change it, feel free.

ekohl: Yes, basically make it a normal sentence. Because I haven't used Windows I can't exactly give advice but the flow of the paragraph might be nicer.

das-g: Reading the paragraph again, if Windows versions where CMD has to be found by the search field to be started are now prevalent, then I think we should **switch** the parenthesized instruction and the preceding one:

> In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu **and enter "Command Prompt" into the search field there.** (On **older** versions of Windows, **you can start the Command Line with Start menu → Windows System → Command Prompt**.)

(emphasis only in this comment to highlight changed parts)

Though, I think this is a decision to be made independently of whether we move the current parenthesized sentence to where it belongs, so if this PR here is OK, please merge it as-is. (And open an issue or another PR for further changes.)

das-g: > [...] this is a decision to be made independently of whether we move the current parenthesized sentence to where it belongs, so if this PR here is OK, please merge it as-is. (And open an issue or another PR for further changes.)

Any objections, @ekohl ?

ekohl: I'm not sure we need this intermediate step of this PR. Your suggestion above sounds good.

After that: do we know which version is "older" and how much it's used? If search always works and we can provide the CMD+r -> cmd as a fallback that always works, we can drop the `Windows System` version. Note this is probably a separate PR.

ekohl:
Thanks!

PR #1285, opened 2018-06-19
Replace {% load staticfiles %} with {% load static %}

bulv1ne:
Changes in this pull request:

- Replace `{% load staticfiles %}` with `{% load static %}`

The usage of `{% load static %}` has been available since [Django 1.10](https://docs.djangoproject.com/en/dev/releases/1.10/#django-contrib-staticfiles).

The usage of `{% load staticfiles %}` has become deprecated in [Django 2.1](https://docs.djangoproject.com/en/dev/releases/2.1/#id2).


bulv1ne:
I restored the code in the languages not supporting django 1.10 or above. I double checked which languages are using which version of django just to be sure.

bulv1ne:
@magul This PR is ready for review.

PR #1287, opened 2018-06-23
Add minor nit to deploy section

alyobabi:
During the tutorial I had a problem with the authentication on pythonanywhere.com and I've created a superuser in pythonanywhere.com Bush Console.

Changes in this pull request:

-
-
-


magul:
Hi there,

I spoted that English (source) version contains:
```
The main thing to notice right now is that your database on PythonAnywhere is actually totally separate from your database on your own PC—that means it can have different posts and admin accounts.
```
Do You feel that's enough or should it be better described there?

I know that deployment chapter was totally rewritten and it would be great to make Russian translation up-to-date with English one, have You considered also help with that on Crowdin?

You can login there using your Github account and Russian translation is stored here: https://crowdin.com/project/django-girls-tutorial/ru

alyobabi:
Yeah, that's enough in eng version.

PR #1293, opened 2018-06-30
flipped below and above

DeepGreenThumbs:
two elements were listed backwards


DeepGreenThumbs:
oh, disregard! Misunderstood the intent of the paragraph.

PR #1296, opened 2018-07-06
[en] Code Editor - add Visual Studio Code

ondrejsika:
Add Visual Studio Code as a text editor. I think, it is the most familiar code editor for beginners.


graup:
Agree, have been recommending it to my students for some time.

KayO-GH:
So I added a small section for users of the VS Code terminal to be able to be able to activate their virtual envs. It's the second commit in the referenced pull request. Do I have to do anything else or someone will handle it from here on?
Forgive my childish question... I don't work on a lot of collaborative projects.

PR #1300, opened 2018-07-14
Adding django-admin parameters

CsatariGergely:
When a new project is started in django-admin the startproject and the name of the project are mandatory parameters.

Changes in this pull request:
- Adding django-admin parameters

### DIFF
--- a/en/django_start_project/README.md
+++ b/en/django_start_project/README.md
@@ -47,7 +47,7 @@ The `(myvenv) C:\Users\Name\djangogirls>` part shown here is just example of the
 
 <!--endsec-->
 
-`django-admin.py` is a script that will create the directories and files for you. You should now have a directory structure which looks like this:
+`django-admin.py` is a script that will create the directories and files for you. When you execute it you should indicate, that you would like to start a new project and tell the project name. For example `django-admin.py startproject myfirstproject`. As a result a new directory is created with the same name as your project. You should now have a directory structure which looks like this:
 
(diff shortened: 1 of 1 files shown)


en/django_start_project/README.md
```diff
@@ -47,7 +47,7 @@ The `(myvenv) C:\Users\Name\djangogirls>` part shown here is just example of the
 
 <!--endsec-->
 
-`django-admin.py` is a script that will create the directories and files for you. You should now have a directory structure which looks like this:
+`django-admin.py` is a script that will create the directories and files for you. When you execute it you should indicate, that you would like to start a new project and tell the project name. For example `django-admin.py startproject myfirstproject`. As a result a new directory is created with the same name as your project. You should now have a directory structure which looks like this:
```
das-g: The full command (i.e. `django-admin.exe startproject mysite .`&mdash;note the `.` at the end) is already given above in the operating system specific sections. I don't think this additional information and example bring an improvement. Especially using `myfirstproject` as the project directory name, when the actual instructions use `.` to have the current directory as project dir is probably confusing. But I also don't think we need to go into that detail here.

If we want to provide more information about the `django-admin.py` command, I'd prefer a link to the respective part of the official Django documentation over adding more explanations to the Django Girls tutorial.

CsatariGergely: Hah, okay. I've ignored the operating system specific sections totally :).

PR #1304, opened 2018-07-21
Fix manage.py migrate output text

graup:
It seems the output of this command has been changed recently. I deleted a line that no longer appears.


ekohl:
Thanks!

PR #1307, opened 2018-07-23
Allow for higher Bugfix/Patch version of Django

das-g:
I.e. Django 2.0.x, where x ≥ 6

Changes in this pull request:

- change `==` to `~=` in `requirements.txt`, thereby making the smallest specified version component a minimum instead of a hard-coded exact value. This is equivelent to `>=2.0.6,<2.1`, but easier to adapt (and shorter).

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -183,7 +183,7 @@ In your `djangogirls/requirements.txt` file you should add the following text:
 
 {% filename %}djangogirls/requirements.txt{% endfilename %}
 ```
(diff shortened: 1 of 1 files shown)


das-g:
> We used to have this and I thought the 2.0 upgrade also included it. Guess it got lost in the conversion to requirements.txt

Indeed, that happened in 49705deb44581b31d815ad6a2974aee79ef1345f and I didn't notice when re-reviewing.

Thanks for merging to quickly!

PR #1310, opened 2018-07-27
[Russian] Fix Django version in doc link

izomeraza:
As the tutorial uses 1.11 version, it seems reasonable to share doc links for the same version: 

1.8 -> 1.11


ekohl:
Thanks!

PR #1311, opened 2018-07-28
fix combining QueySets documentation

graingert:
Changes in this pull request:

- QuerySets cannot be combined by chaining, only union (or intersection) can
- Document how to combine querysets
- Document how to chain queryset refinements

### DIFF
--- a/en/django_orm/README.md
+++ b/en/django_orm/README.md
@@ -190,9 +190,22 @@ We can also reverse the ordering by adding `-` at the beginning:
 ```
 
 
-### Chaining QuerySets 
+### Combining QuerySets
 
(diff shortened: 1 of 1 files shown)


das-g:
I think the tutorial shouldn't be expanded too much and can never cover all features of the Django ORM, anyway. Thus I've suggested an alternative change to fix the wrong semantics that @graingert noticed: #1466

graingert:
I think it would be better to use the terminology that Django uses. Eg 'refinements'

aniav:
I don't think beginners need to learn about the union query at this level of tutorial. We throw qiute a lot of knowledge at them, and I wouldn't call `union` as a base knowledge.

PR #1315, opened 2018-08-01
Add VS code powershell activation code

KayO-GH:
Corrected name of function being imported from 'url' to 'path' under the blog.urls section

The line I changed indicated that we were importing the url function from django.urls, when we are in fact importing the path function from django.urls.
This was confusing the first time I read it

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -150,6 +150,13 @@ Remember to replace `myvenv` with your chosen `virtualenv` name!
 
 <!--endsec-->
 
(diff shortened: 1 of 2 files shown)


ekohl:
I have merged your first commit. The VS Code change perhaps belong in https://github.com/DjangoGirls/tutorial/pull/1296?

KayO-GH:
Oh OK. Thank you.

I'm sorry for not looking into the issues first.

I had very little time yesterday.

aniav:
I manually merged this as it's adding an improvement and has been waiting for quite some time 💯

Issue #1259, opened 2018-05-07
Possible incorrect installation tutorial for Linux


ghost:
### Issue description

Installation instructions for Linux may be incorrect. Most of the Linux distributions have Python pre-installed. It is a one of the Linux development languages. So, it is important, that only the current version is installed globally. 

Your installation instructions [here](https://tutorial.djangogirls.org/en/installation/) proposes to install a version 3.6 of Python globally. The problem here is that if the system has already another Python version installed, it may break the normal OS work.

I would suggest to change that part of the tutorial to something described in [this post](https://www.caseylabs.com/how-to-create-a-python-3-6-virtual-environment-on-ubuntu-16-04/).

I tested it myself on Ubuntu 16.04 and it works great.

### Language

Is this related to a specific language of the tutorial? **No**

### Operating system

What operating system does this issue relate to? **Linux**

ekohl:
That particular blog post doesn't load for me, but I think it makes sense to change it to `apt install python3`. Should we consider this a duplicate of https://github.com/DjangoGirls/tutorial/issues/1222? A PR would be very welcome.

alevikpes:
It seems to be currently down. Please, visit it later.

The idea is that any other version of Python must be run in a virtual environment. There is a description in that article about how to do it.

I found also [this description](https://askubuntu.com/a/865569), which explains the same.

jhodgdon-drp:
It would also be helpful to say which versions of Python would work for the Linux installation.

In the Windows install instructions, it says that if you have an older version of Windows, you can install an older version of Python. If this is OK to do, then probably an older version of Python would work for Linux as well? For instance, my machine already had Python 3.5.x -- would that be OK? That is what I currently have (from the python3 package).

Incidentally, the instructions for Ubuntu saying
````
sudo apt install python3.6
````
are no good on Ubuntu 16.04.5 LTS.  I got:
````
E: Unable to locate package python3.6
````
So... I'm in favor of changing it to use package python3. This package exists. But for me, it's currently on python 3.5.1 and it says that is the latest version for my current Ubuntu installation, so that needs to be dealt with.

jhodgdon-drp:
Oh, the python3.6 package not available bug is #1222 sorry for the noise. But the same PR would fix both I think.

jhodgdon-drp:
That pull request should take care of the issue, plus #1222 as well.

jhodgdon-drp:
[...]

Issue #1222, opened 2018-02-19
Python 3.6 unavailable for Ubuntu 16.04 LTS by default


ThomasWo:
### Issue description
On https://tutorial.djangogirls.org/en/python_installation/ , it asks to install Python 3.6, but Python 3.6 isn't available in the default apt-get repos for Ubuntu 16.04 LTS without adding a PPA.

Instructions to add said PPA is here: https://www.rosehosting.com/blog/how-to-install-python-3-6-on-ubuntu-16-04/

Might I recommend add some language around detecting the version of Ubuntu and branching off accordingly?

Note: Python 3.6 _is_ available for installation in apt-get for Ubuntu 17.10, so if they're using the latest version, that will work out of the box.

### Language
Found in English, haven't looked elsewhere.

### Operating system
Ubuntu 16.04 LTS

sevu:
I guess the tutorial works with python 3.5 as well. The command should mention `python3` instead `python3.6`.

ekohl:
@sevu I think it's a good thing to use ```python3``` since we can pretty much assume it works in the various python minor versions.

saschalalala:
The same issue exists for the latest Debian. Debian 9.4 ships Python 3.5.3. So when letting users install their system's python one has to be aware that e.g. f-strings where only introduced with Python 3.6 :/

jhodgdon-drp:
That pull request should take care of this issue, plus #1259.

jhodgdon-drp:
I closed the earlier pull request and substituted this new one, which should take care of 3 related issues.

Issue #1322, opened 2018-08-20
Installation : Python version confusion -- install 3.6 specifically, or latest (3.7)?


jhodgdon-drp:
### Issue description

In the Windows installation instructions, it tells you to install Python version 3.6.x. But it also tells you to install the "latest version", which is currently 3.7.x. It is not clear which one should be used -- will Django work with 3.7.x? Is it better to get the latest version?

If the tutorial is updated to say to install 3.7.x, by the way, the instructions about the PATH setting are somewhat wrong, as the installation setup dialog now says something about "environment variables' and not "PATH" specifically.

### Operating system

This probably applies to all operating systems, actually.

jhodgdon-drp:
I went through the entire tutorial over the past few days with 3.7 and it seems to work fine.  Still not sure whether the tutorial should recommend 3.6 specifically or the latest version?

magul:
I believe it was recommending Python 3.6 as latest version, but it always looked a little bit too restrictive to me. Django 2 support Python 3.4 - 3.7 so we should recoment latest, but be ok with anything above or equal 3.4.

jhodgdon-drp:
OK, that is good information that should get into the tutorial. See also #1259 and #1222

jhodgdon-drp:
This pull request should fix all 3 related issues.

PR #1325, opened 2018-08-21
Fix package instructions for Python for Ubuntu

jhodgdon-drp:
Fix for #1259 and #1222 by updating package name for Ubuntu to python3, and clarifying which versions will be OK to use.

Changes in this pull request:

-
-
-


jhodgdon-drp:
I think we should disregard this pull request and start over so it can fix #1259, #1222, and #1322, so I'm closing this one.

PR #1329, opened 2018-08-21
Add more notes to top of installation page

### LINKED ISSUE #1014: Move the "Installation" step to the end of the table of contents
Hi all,

During out workshop in Melbourne today several of the participants navigated directly from the Introduction to the Installation step. This definitely slows down the progress as people end up installing everything up front instead of doing it as they move through the tutorial.

Is there a reason why this is listed as the second step? If not, I'd be happy to put together PR to move it to the bottom.

jhodgdon-drp:
Proposed fix for #1014 . Instead of moving the Installation page to the end, it adds more notes at the top, including a section for people who are at workshops.

Fixes #1014


jhodgdon-drp:
It looks like this PR was somewhat approved a while back... Do we still need a second reviewer?

PR #1336, opened 2018-08-22
Clarify instructions on URLs

### LINKED ISSUE #1335: Getting wrong error in urls.py
### Issue description

I'm attempting to run the tutorial. At the end of the django_urls page, I'm supposed to get an error saying that the post_list view doesn't exist.

But instead I'm getting an error saying 
````
NameError: name 'include' is not defined.
````

When I did a web search on this error, I found that there is a missing line in urls.py:
https://stackoverflow.com/questions/34471102/python-nameerror-name-include-is-not-defined

So I added this line to the top of my urls.py file:
````
from django.conf.urls import include
````

Now I get an error saying 
````
ModuleNotFoundError: No module named 'blog.urls'
````
I have no idea how to proceed now. The tutorial is just not working for me...

### Operating system

I'm running on Windows 8.1 if that makes a difference.

jhodgdon-drp:
Fixes #1335

### DIFF
--- a/en/django_urls/README.md
+++ b/en/django_urls/README.md
@@ -95,7 +95,8 @@ If you try to visit http://127.0.0.1:8000/ now, then you'll find some sort of 'w
 
 ![Error](images/error1.png)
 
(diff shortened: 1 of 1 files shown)


ekohl:
This is a longer standing issue. #1165 was about the same thing. In b97aa7cc524eee3863177bff192917a617648044 I changed it to not modify the existing line but explicitly add a new line because that's harder to miss (as suggested by someone else). Looks like this was lost when we migrated the instructions to Django 2.0. Perhaps worth reintroducing?

jhodgdon-drp:
I need to update this PR because we have similar instructions elsewhere for restarting. We should probably make a small .md file that gives the instructions for restarting the web server, and then include it in various places, rather than reinventing it every time it is needed. Anyway, I don't recommend merging just yet...

ekohl:
Thanks!

jhodgdon-drp:
You're welcome! I'm happy to be able to improve the tutorial a bit. :)

PR #1341, opened 2018-08-22
Clear up Python version confusion

### LINKED ISSUE #1222: Python 3.6 unavailable for Ubuntu 16.04 LTS by default

### LINKED ISSUE #1259: Possible incorrect installation tutorial for Linux

### LINKED ISSUE #1322: Installation : Python version confusion -- install 3.6 specifically, or latest (3.7)?

jhodgdon-drp:
We have 3 open issues about Python version and package confusion. This PR attempts to fix them all:
- Fixes #1222: says that the python3.6 package is unavailable for Ubuntu, so the instructions are changed to use package python3
- Fixes #1259: talks about the fact that Linux installs probably already have Python. The instructions in this PR are changed to tell you which versions are OK and which ones you need to upgrade.
- Fixes #1322: talks about the Windows instructions, which both say to get the latest Python and to get 3.6. The instructions are updated to say to get the latest.

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -2,14 +2,14 @@
 
 > This section is based on a tutorial by Geek Girls Carrots (https://github.com/ggcarrots/django-carrots)
 
-Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install Python 3.6, so if you have any earlier version, you will need to upgrade it.
+Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install the latest version of Python 3, so if you have any earlier version, you will need to upgrade it, but if you already have version 3.4 to 3.7 you should be fine.
 
 
 <!--sec data-title="Install Python: Windows" data-id="python_windows" data-collapse=true ces-->
 
 First check whether your computer is running a 32-bit version or a 64-bit version of Windows, by pressing the Windows key + Pause/Break key which will open your System info, and look at the "System type" line. You can download Python for Windows from the website https://www.python.org/downloads/windows/. Click on the "Latest Python 3 Release - Python x.x.x" link. If your computer is running a **64-bit** version of Windows, download the **Windows x86-64 executable installer**. Otherwise, download the **Windows x86 executable installer**. After downloading the installer, you should run it (double-click on it) and follow the instructions there.
 
-One thing to watch out for: During the installation, you will notice a window marked "Setup". Make sure you tick the "Add Python 3.6 to PATH" checkbox and click on "Install Now", as shown here:
+One thing to watch out for: During the installation, you will notice a window marked "Setup". Make sure you tick the "Add Python 3.6 to PATH" or 'Add Python to your environment variables" checkbox and click on "Install Now", as shown here (it may look a bit different if you are installing a different version):
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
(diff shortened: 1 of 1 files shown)


en/python_installation/instructions.md
```diff
@@ -2,14 +2,14 @@
 
 > This section is based on a tutorial by Geek Girls Carrots (https://github.com/ggcarrots/django-carrots)
 
-Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install Python 3.6, so if you have any earlier version, you will need to upgrade it.
+Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install the latest version of Python 3, so if you have any earlier version, you will need to upgrade it, but if you already have version 3.4 to 3.7 you should be fine.
```
ekohl: Not sure we should mention 3.7 here. How about 3.4 or newer? That's also what the requirements on PyPI are.

magul: @ekohl current English version installs Django 2.0.6. Since 2.0 Django supports Python 3.7, so we should be fine here.

jhodgdon-drp: The requirements for Django 2.x are here:
https://docs.djangoproject.com/en/2.1/releases/2.0/

It says specifically 3.4 to 3.7, at least currently. So I think we should leave it being specific, until Django changes its requirements?

jhodgdon-drp:
Thanks for the review! How's this version?

en/python_installation/instructions.md
```diff
@@ -50,7 +50,7 @@ $ python3 --version
 Python 3.6.1
 ```
 
-If you have a different 'micro version' of Python installed, e.g. 3.6.0, then you don't have to upgrade. If you don't have Python installed, or if you want a different version, you can install it as follows:
+If you have a different version of Python installed, at least 3.4.x (e.g. 3.6.0), then you don't have to upgrade. If you don't have Python installed, or if you want a newer version, you can install it as follows:
```
ekohl: I have doubts about the 3.4.x. It implies to me that you can only change the x. How about 3.4.0?

jhodgdon-drp: Good thought, fixed that in next commit.

en/python_installation/instructions.md
```diff
@@ -99,8 +99,9 @@ Verify the installation was successful by opening a command prompt and running t
 $ python3 --version
 Python 3.6.1
 ```
+The version shown may be different from 3.6.1 -- it should match the version you installed.
 
-**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) and check if it still might be a version of Python 3.6.
+**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) and check if it still might be a version of Python that is 3.4.x or higher.
```
ekohl: This should also be 3.4.0 then.

jhodgdon-drp: Yeah, sorry, missed that one.

PR #1342, opened 2018-08-23
modify STATIC_ROOT in settings.py

doyou89:
Cannot find 'css/blog.css'
If you want to create 'static' folder under 'blog', then STATIC_ROOT should be os.path.join(BASE_DIR, 'blog/static')

Changes in this pull request:

-
-
-


ekohl:
This shouldn't be needed. I can't read Korean but I am wondering when you exactly see this error. Can you give a bit more context what you did?

PR #1351, opened 2018-08-28
Make clear that view does not return a function

das-g:
Changes in this pull request:

- make clear that a regular value (the one returned by `render(...)` will be returned by the view
- switch that part of the sentence to future tense, to be somewhat more grammatical


ekohl:
Thanks!

PR #1355, opened 2018-08-29
Accessibility fixes

### LINKED ISSUE #1338: Instructions for HTML template and CSS are not accessible
### Issue description
We should probably not suggest that people make HTML templates and CSS that violate the W3C accessibility guidelines.
https://www.w3.org/WAI/standards-guidelines/wcag/

There are two problems I saw in the tutorial so far:

a) In post_list.html, it uses an H1 tag for the page title (which is good), but then an H1 tag for the post title as well. It shouldn't. Accessibility guidelines state there should only be one H1 tag on a page. The other should be H2.
https://www.w3.org/WAI/tutorials/page-structure/headings/

b) In blog.css, it is suggested to make the H1 titles orange #FCA205. This is insufficient contrast for accessibility, when displayed on a white background (or when white text is displayed on this color background):
https://www.w3.org/TR/WCAG21/#contrast-minimum
https://webaim.org/resources/contrastchecker/

jhodgdon-drp:
Partially fixes #1338 

Fixes the text in tutorial pages to use better contrast color in the CSS, and to only have one h1 header on a page (others are changed to h2).

Screenshots still need to be updated.

### DIFF
--- a/en/css/README.md
+++ b/en/css/README.md
@@ -77,15 +77,13 @@ In your `blog/static/css/blog.css` file you should add the following code:
 
 {% filename %}blog/static/css/blog.css{% endfilename %}
 ```css
-h1 a {
+h1 a, h2 a {
     color: #C25100;
 }
-h2 a {
-    color: #C25100;
-}
+
 ```
 
-`h1 a` is a CSS Selector. This means we're applying our styles to any `a` element inside of an `h1` element (the `h2 a` selector does the same thing for `h2` elements). So when we have something like `<h1><a href="">link</a></h1>`, the `h1 a` style will apply. In this case, we're telling it to change its color to `#C25100`, which is a dark orange. Or you can put your own color here, but make sure it has good contrast against a white background!
+`h1 a` is a CSS Selector. This means we're applying our styles to any `a` element inside of an `h1` element; the `h2 a` selector does the same thing for `h2` elements. So when we have something like `<h1><a href="">link</a></h1>`, the `h1 a` style will apply. In this case, we're telling it to change its color to `#C25100`, which is a dark orange. Or you can put your own color here, but make sure it has good contrast against a white background!
 
 In a CSS file we determine styles for elements in the HTML file. The first way we identify elements is with the element name. You might remember these as tags from the HTML section. Things like `a`, `h1`, and `body` are all examples of element names.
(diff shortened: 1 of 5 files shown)


en/css/README.md
```diff
@@ -70,19 +70,22 @@ Time to write some CSS! Open up the `blog/static/css/blog.css` file in your code
...
 
 In your `blog/static/css/blog.css` file you should add the following code:
 
 {% filename %}blog/static/css/blog.css{% endfilename %}
 ```css
 h1 a {
-    color: #FCA205;
+    color: #C25100;
+}
+h2 a {
```
ekohl: Should this be `h1 a, h2 a` perhaps? Not sure if the current explicitness makes it easier for beginners though.

en/css/README.md
```diff
@@ -70,19 +70,22 @@ Time to write some CSS! Open up the `blog/static/css/blog.css` file in your code
...
+    color: #C25100;
 }
 ```
 
-`h1 a` is a CSS Selector. This means we're applying our styles to any `a` element inside of an `h1` element. So when we have something like `<h1><a href="">link</a></h1>`, the `h1 a` style will apply. In this case, we're telling it to change its color to `#FCA205`, which is orange. Or you can put your own color here!
+`h1 a` is a CSS Selector. This means we're applying our styles to any `a` element inside of an `h1` element (the `h2 a` selector does the same thing for `h2` elements). So when we have something like `<h1><a href="">link</a></h1>`, the `h1 a` style will apply. In this case, we're telling it to change its color to `#C25100`, which is a dark orange. Or you can put your own color here, but make sure it has good contrast against a white background!
```
ekohl: IMHO the parenthesis aren't needed here. The contents are always relevant IMHO.

en/css/README.md
jhodgdon-drp: Yeah, I wondered about that. I wasn't sure if we wanted to get into explaining having two selectors for the same code or not?

jhodgdon-drp: Actually I see we have multiple selectors on the same line farther down in the file, so I made the change here.

en/css/README.md
jhodgdon-drp: Fixed in next commit.

ekohl:
Thanks!

PR #1359, opened 2018-09-02
Fix div tag in German translation

The-Compiler:
The English version says "replace everything from the second `<div>` to the third `</div>`" which seems correct.

### DIFF
--- a/de/django_templates/README.md
+++ b/de/django_templates/README.md
@@ -20,7 +20,7 @@ Um eine Variable in einem Django-Template darzustellen, nutzen wir doppelte, ges
 {{ posts }}
 ```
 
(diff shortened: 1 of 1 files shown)


de/django_templates/README.md
```diff
@@ -103,4 +103,4 @@ Glückwunsch! Nun kannst du dich daran machen, neue Posts in deinem Django Admin
 
 Funktioniert super? Wir sind so stolz auf dich! Steh kurz auf und geh ein Stück weg vom Computer. Du hast dir eine Pause verdient :)
 
-![Abbildung 13.4](images/donut.png)
\ No newline at end of file
+![Abbildung 13.4](images/donut.png)
```
The-Compiler: Looks like there wasn't a newline at the end of line before, so the GitHub web editor added one? Let me know if I should remove it again, but I suppose others might want to edit the tutorial via the GitHub editor as well.

ekohl: I'm in favor of newlines a the end so it's fine.

ekohl:
Thanks!

PR #1313, opened 2018-07-29
Portuguese, Brazilian translation from Crowdin

magul:
Beta version as it was not yet proofread.

Translation made by:
 * Adailton do nascimento (https://crowdin.com/profile/dhelbegor)
 * Adam Victor Nazareth Brandizzi (https://crowdin.com/profile/brandizzi)
 * Adjamilton Junior (https://crowdin.com/profile/jr)
 * Alessandra Cruz (https://crowdin.com/profile/tacigomess)
 * Álvaro Justen (https://crowdin.com/profile/turicas)
 * Ana Carolina Arcanjo (https://crowdin.com/profile/aarcanjo)
 * Ana Clara Lacerda (https://crowdin.com/profile/aninhalacerda)
 * Anna Warzecha (vi) (https://crowdin.com/profile/aniav)
 * Annanda Sousa (https://crowdin.com/profile/annanda)
 * Antonio Luis (https://crowdin.com/profile/toim_23)
 * Ariane Camilo (https://crowdin.com/profile/ArianeCamilo)
 * arturlicks (https://crowdin.com/profile/arturlicks)
 * Bernardo Fontes (https://crowdin.com/profile/bernardoxhc)
 * bksrocks (https://crowdin.com/profile/bksrocks)
 * Camila Achutti (https://crowdin.com/profile/camilaachutti)
 * Camilla Nobre (https://crowdin.com/profile/camillanobre)
 * classicalmu (https://crowdin.com/profile/zhanyanqiao)
 * Claudio Ferreira (https://crowdin.com/profile/filhocf)
 * Cleiton Lima (https://crowdin.com/profile/cleiton_lima)
 * clemente.jnr (https://crowdin.com/profile/clmntjr)
 * Cynthia Zanoni (https://crowdin.com/profile/cyz)
 * Daniel Teixeira dos Santos (https://crowdin.com/profile/danieltex)
 * Diogo Leal (https://crowdin.com/profile/diogoleal)
 * Eduardo Braga (https://crowdin.com/profile/eduardocbraga)
 * Eduardo Vital (https://crowdin.com/profile/vital-edu)
 * Eric Hideki (https://crowdin.com/profile/erichideki)
 * Erick Simões (https://crowdin.com/profile/erick_simoes)
 * Fábio (https://crowdin.com/profile/ffabiorj)
 * Fábio C. Barrionuevo da Luz (https://crowdin.com/profile/bnafta)
[...]

PR #1361, opened 2018-09-11
Fixed a littlet expression

taniaa:
Changes in this pull request:

- fixed a little expression from 'por acidentalmente' to 'acidentalmente'


magul:
Hi there, can You also add this fix to our Crowdin project here: https://crowdin.com/project/django-girls-tutorial/pt-BR#

We use Crowdin to build our translation and your change could be overwritten by Crowdin if You will not apply it there also. See for example #1313

PR #1362, opened 2018-09-13
Fix crashed markdown link

maxtortime:
Changes in this pull request:

- The url was enclosed in square brackets rather than parentheses.


ekohl:
Thanks!

PR #1363, opened 2018-09-13
Change Python 3.6 to 3.5

maxtortime:
I think `pythonanywhere` doesn't provide **3.6** version, but in this tutorial use **3.6**, So I change them to python **3.5**.

image

Changes in this pull request:

- Change all 3.6 text to 3.5 text.

### DIFF
--- a/ko/deploy/README.md
+++ b/ko/deploy/README.md
@@ -179,8 +179,8 @@ PythonAnywhere에서도 내 컴퓨터에 있는 것과 같이 작동할 수 있�
 ```
 $ cd my-first-blog
 
-$ virtualenv --python=python3.6 myvenv
-Running virtualenv with interpreter /usr/bin/python3.6
+$ virtualenv --python=python3.5 myvenv
+Running virtualenv with interpreter /usr/bin/python3.5
 [...]
 Installing setuptools, pip...done.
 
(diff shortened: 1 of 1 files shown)


magul:
It actually support Python 3.6. Did you make this screenshot today or it originate from current tutorial chapter?

magul:
I forgot about link https://blog.pythonanywhere.com/151/

maxtortime:
I captured above screenshot yesterday. 

image

But it is also same situation today. How can I use Python 3.6 for bash console?

das-g:
@hjwp Do the available Python version differ between PythonAnywhere accounts?

maxtortime:
@das-g I think so too. I guess this is the difference between the paid version and the free version.

hjwp:
if you have a really old pythonanywhere account it may still be on the old system image, which does not have python 3.6.  you can ask for it to be upgraded to the new image at any time by emailing support@pythonanywhere.com.  it's free.

hjwp:
actually the very latest system image even has python3.7 as well as 3.6!  https://blog.pythonanywhere.com/166/

maxtortime:
Thanks. I think it is a problem that is limited to me.

PR #1372, opened 2018-09-21
Changes to chapters after Your First Django Project -> where the manage.py file ends up.

BenPoser:
When I went through the tutorial myself (and also on another django tutorial this also happened so I'm fairly sure it is consistent) it created a mysite directory inside the Djangogirls directory this mysite directory contained the manage.py file and another mysite directory with __init__.py etc. So if you try to run any commands on the manage.py file whilst in the Djangogirls directory it wouldn't work. I updated as many of the command line visuals I could showing that you needed to be in the mysite directory containing the manage.py file. It might be worth adding in a cd mysite part also? Or if I'm wrong (I am only a beginner) and this happens to some people or it's something weird I did then still might be worth having as an optional troubleshooting bit. Working on the assumption that this is just how it works for everyone I also updated how the file directories looked.

Changes in this pull request:

- updated all the command line prompts I could see to show what subdirectory or the Djangogirls directory you need to be in
- updated the directory structure to reflect what happened to me

Hopefully this helps. Have really enjoyed the tutorial so far and keep up the good work! 👍

### DIFF
--- a/en/django_admin/README.md
+++ b/en/django_admin/README.md
@@ -24,12 +24,12 @@ To log in, you need to create a *superuser* - a user account that has control ov
 
 {% filename %}Mac OS X or Linux:{% endfilename %}
 ```
-(myvenv) ~/djangogirls$ python manage.py createsuperuser
+(myvenv) ~/djangogirls/mysite$ python manage.py createsuperuser
 ```
 
 {% filename %}Windows:{% endfilename %}
 ```
-(myvenv) C:\Users\Name\djangogirls> python manage.py createsuperuser
+(myvenv) C:\Users\Name\djangogirls\mysite> python manage.py createsuperuser
 ```
 
 When prompted, type your username (lowercase, no spaces), email address, and password. **Don't worry that you can't see the password you're typing in – that's how it's supposed to be.** Type it in and press `enter` to continue. The output should look like this (where the username and email should be your own ones):
--- a/en/django_models/README.md
+++ b/en/django_models/README.md
@@ -68,34 +68,35 @@ To keep everything tidy, we will create a separate application inside our projec
 
 {% filename %}Mac OS X and Linux:{% endfilename %}
 ```
-(myvenv) ~/djangogirls$ python manage.py startapp blog
+(myvenv) ~/djangogirls/mysite$ python manage.py startapp blog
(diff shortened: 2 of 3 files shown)


das-g:
That happens when you use
```bash
django-admin startproject mysite
```
instead of
```bash
django-admin startproject mysite .
```
in the ["Your first Django project!" chapter](https://tutorial.djangogirls.org/en/django_start_project/#your-first-django-project). We are aware the difference is subtle, thus why we emphasize **in bold**:

> **Don't forget to add the period (or dot) `.` at the end!**

([here](https://github.com/DjangoGirls/tutorial/blame/cec495335455055603afda5a7c6722cb887879bb/en/django_start_project/README.md#L21) for Mac and Linux and [here](https://github.com/DjangoGirls/tutorial/blame/cec495335455055603afda5a7c6722cb887879bb/en/django_start_project/README.md#L37) for Windows)

das-g:
I suggest we do keep the dot (so that we don't produce unnecessarily nested directory structures) and thus don't change the current working dir indicates (as this PR would) nor add `cd` commands (as the PR description suggests).

Though, if you have suggestions to make it even more clear that the dot is part of the command and should not be omitted, I'd be very interested to hear them. At every workshop I've coached so far, there have been some participants that missed it, so it's indeed not prominent enough as it is now.

BenPoser:
Thanks for the clarification, I definitely agree that keeping the dot is worth it as I will now be forced to go through this unnecessary directory myself! I think it's likely one of those small mistakes some users will inevitably make, especially as this is a tutorial aimed at beginners it's better not to change the route the tutorial takes but rather show a way out to people if they did make the mistake.

With that in mind maybe it would just be worth at that point getting users to check their file directories and make it clear that if they didn't include the dot then their manage.py file will be inside a mysite directory and not to worry, just they will have to change to the mysite directory whenever there is a command using manage.py. 

I would be happy to write something like that if it will help :)

ekohl:
In the instructions we have instructions for this. Not sure how we can make this clearer.
> The period . is crucial because it tells the script to install Django in your current directory (for which the period . is a short-hand reference).

For now I'm going to close this but I'd be happy to accept a PR if you can think of an improvement.

Issue #1368, opened 2018-09-17
Suggested addition to the pre-workshop requirements.txt instructions


JuXp:
### Issue description

I recently attended the DjangoGirls workshop in Cardiff.  Prior to the workshop we had to download software to our computer. The part of the installation regarding the requirements.txt file was unclear here:
 

![screen shot 2018-09-17 at 16 01 45](https://user-images.githubusercontent.com/42972531/45631578-14fec580-ba93-11e8-976d-8e0e4df82f12.png)

I would like to suggest this additional sentence:

"You do this by opening a new file in the code editor and then saving as requirements.txt it in the djangogirls/ folder.  Your directory will look like this:"

I'm open to other suggestions, I just think some more wording is needed to avoid any confusion.  Mainly because the folder structure convention was not familiar to me as a beginner.

### Language

It's related to the English language version of the tutorial

das-g:
These instructions are used at two places:
- [in the pre-event installation](https://tutorial.djangogirls.org/en/installation/#installing-packages-with-requirements)
- [in the "Django installation" chapter](https://tutorial.djangogirls.org/en/django_installation/#installing-packages-with-requirements)

In the former case, your suggested wording is probably better. In the latter case, it probably won't matter too much, as participants will have created some files that way by then, but some reminder probably won't hurt.

Though I wonder how much sense it makes to create the virtualenv and install Django in advance. Installing Python, Git and a code editor like Atom in advance makes sense to save some time during the workshop. Even if not everyone manages to do that at home, coaches can then focus on helping those that didn't succeed in installing all of that.

But creating a virtualenv and installing Python packages (like Django) in it is very much part of the software development process, and thus should iMHO be part of the Workshop proper. (So that a coach can also explain what's going on behind the commands, if participants wonder about that.)

ekohl:
https://github.com/DjangoGirls/tutorial/pull/1374 fixes this

PR #1374, opened 2018-09-22
add detail to instructions for requirements.txt 

lisac:
connects to https://github.com/DjangoGirls/tutorial/issues/1368


jhodgdon-drp:
I think the intention in this step of the tutorial was actually that you would go to your file explorer application, and create a new file first, then open it in the code editor?

PR #1377, opened 2018-09-22
Minor proposals to wording

lisac:
Proposals:

- ~reference "syntax highlighting" in the instructions about code editors~
- replace "your own PC" with "your own computer" (to be more generic)


jhodgdon-drp:
I reviewed this pull request, and the changes look good to me.

However, regarding the PC -> computer change, there is a second place in that same file that should also be changed. Where it says "... is actually totally separate from your database on your own PC..."

I used grep and verified that "PC" isn't used anywhere else in the tutorial except in these 2 places in the deploy topic, whereas "computer" is used many times, so this is definitely a good change to make for consistency.

jhodgdon-drp:
Another comment: The term "syntax highlighting" is actually introduced later, in the python_introduction topic, where the reader can actually see it in action. So... maybe it is not too necessary to introduce it here as well?

jhodgdon-drp:
PR looks good to me now. Thanks for the updates!

lisac:
Thank you for the feedback @jhodgdon-drp ! I apologize for taking so long to respond. 

I agree with your review and have added commits that address that additional instance of "PC" and reverts the quick definition of "syntax highlighting".

So, revised summary of this PR: 
- replace "your own PC" with "your own computer" (to be more generic)

PR #1381, opened 2018-09-29
Add note about end-of-install dialog

### LINKED ISSUE #1380: End of Windows Python install pops up confusing dialog
### Issue description

At the end of installing Python on Windows, the installer pops up a dialog asking if you want to read a Python tutorial, read release notes, etc. Two people going through the installation party tutorial this evening asked me what they should do with the dialog, since the tutorial doesn't say anything about it.

So, maybe we should add a quick note?

I'll propose an edit...

### Operating system

Windows

jhodgdon-drp:
Fixes #1380

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -13,6 +13,8 @@ One thing to watch out for: During the installation, you will notice a window ma
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
+When the installation completes, you may see a dialog box asking if you want to learn more about Python or about the version you installed. Close or cancel that dialog -- you'll be learning more in this tutorial!
+
 In upcoming steps, you'll be using the Windows Command Line (which we'll also tell you about). For now, if you need to type in some commands, go to Start menu and enter "Command Prompt" into the search field there. (On older versions of Windows, you can start the Command Line with Start menu → Windows System → Command Prompt.) You can also hold in the Windows key and press the "R"-key until the "Run" window pops up. To open the Command Line, type "cmd" and press enter in the "Run" window.
 
 ![Type "cmd" in the "Run" window](../python_installation/images/windows-plus-r.png)


en/python_installation/instructions.md
```diff
@@ -13,6 +13,8 @@ One thing to watch out for: During the installation, you will notice a window ma
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
+When the installation completes, you may see a dialog box asking if you want to learn more about Python or about the version you installed. Close or cancel that dialog -- you'll be learning more in this tutorial!
```
das-g: Maybe clearer with "whether" instead of "if".

jhodgdon-drp: Hm. I feel like "whether" would need an "or not" in there somewhere... So it would need to be:

... asking whether or not you want to learn...

or

... asking whether you want to learn more about Python or the ... or not

That second one is not good, to many "or" there. The first one... I don't like it better than "if". ???

das-g: I'm not a native speaker of English, but AFAIK, "whether" can be used just fine without being followed by "or not". In fact, to me, it already carries an implied "or not" in its meaning.

While phrasing yes-no questions in indirect speech with "whether" instead of "if" may be a bit more arcane, "if" has the problem that it carries two meanings: introducing conditionals (its IMHO predominant meaning) and introducing yes-no questions in indirect speech (IMO only its secondary meaning). So while a bit far-fetched, using "if" here risks the sentence being misinterpreted as "When the installation completes and if you want to learn more about Python or about the version you installed, then you may see a dialog box asking." Granted, this doesn't make a lot of sense and even less so with the direct object of "to ask" apparently omitted. But it might be enough to trip up readers (maybe specifically non-native readers of English like me), especially since the sentence is not that short.

I think
> When the installation completes, you may see a dialog box asking whether you want to learn more about Python or about the version you installed.

would solve that just fine. (And does so in other occurrences, too, thus why I have a general preference for "whether" as a drop-in replacement for "if" when used with this meaning, not just here in this specific sentence, at least in formal written language. But maybe that just because the English I learned as foreign language is not quite up-to-date with the native speakers' usage, or because I'm a computer scientist/software developer who prefers unambiguously parseable phrasing, or because my native language German has distinct words for conditional "if" ("falls" or "wenn") and "whether"/"if" ("ob").)

A problem that both "if" and "whether" have in this context alike is that it's unclear whether the dialog box asks you whether you want to do either (learn about python and/or learn about the version you installed [vs. doing none of them]) or whether it asks you which of those you want to do (i.e. a choice between learning about python vs. learning about the version you installed.)* However, that distinction probably matters less and when you see the dialog box and its question, it should be clear that the former was meant.

* I don't think I've ever used that many "whether" in a single sentence. And no, that was not intentional, it just came out that way.

jhodgdon-drp:
OK, how about this? Gets rid of both "if" and "whether". :)

PR #1391, opened 2018-09-29
Clarify Windows command line and $

### LINKED ISSUE #1390: Windows users confused by $ prompts
### Issue description

On the Intro to the Command Line page, a Windows user would probably not open the sections that are marked for Linux or Mac. So, they would see examples of what the Windows command prompt looks like.

Then, for the rest of the Tutorial, most of the command prompt examples show the $ prompt. People get confused by that, because they haven't seen this. They try to type the $ and stuff that's shown to the left of the $.

The confusion level is even higher when they get to PythonAnywhere, the command prompt there is more like Linux, and it's very unfamiliar.

Solution?? Maybe let people on Windows know to look at the Linux part because they'll need it later. And add a sentence explaining that the prompts shown in the tutorial will show $ (which is not what Windows users will see).

### Operating system

Windows

jhodgdon-drp:
Fixes #1390

### DIFF
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -47,7 +47,7 @@ You now should see a white or black window that is waiting for your commands.
 <!--sec data-title="Prompt: OS X and Linux" data-id="OSX_Linux_prompt" data-collapse=true ces-->
 
 
-If you're on Mac or Linux, you probably see `$`, like this:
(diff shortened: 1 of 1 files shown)


Forceflow:
Can confirm, the people who I was coaching also often accidentally copy-pasted the "$" into their (Windows) command lines as well.

PR #1394, opened 2018-09-29
Add info about command line to Installation page

### LINKED ISSUE #1379: One-step install page needs more info about command line
### Issue description

On the one-page install page in the tutorial, there are instructions for opening a command window if you are using Windows. But they are missing for other operating systems.

Also, sometimes it's called a "console", which could be confusing, since I don't think it's actually called "console" in any system menu in any operating system. 

This wouldn't be so confusing for people running the full tutorial, because they'd read the "Intro to the command line" topic and get explanations for how to open the command window, and learn about the "console" terminology.

jhodgdon-drp:
Also remove info about command line in python install page.

Fixes #1379

### DIFF
--- a/en/installation/README.md
+++ b/en/installation/README.md
@@ -16,6 +16,52 @@ data-id="chromebook_setup" data-collapse=true ces-->
 {% include "/chromebook_setup/instructions.md" %}
 <!--endsec-->
 
+# Brief intro to command line
+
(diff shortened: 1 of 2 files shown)


ekohl:
Also just noticed there's a merge conflict.

jhodgdon-drp:
Can I suggest that we leave this PR alone for the moment, and go look at #1350 first? I see there is a request for changes there too... I will go address that, and once that one is done, let's come back to this one.

PR #1396, opened 2018-09-29
Make django_orm consistent with django_admin

### LINKED ISSUE #1395: django_orm not consistent with django_admin
### Issue description

On django_admin, you are instructed to create a superuser. The example uses user name "admin".

Then on django_orm, you are instructed to retrieve that user from the database. The example uses user name "ola".

This confused all the attendees at the table I was at. The text said to substitute their user name, but they have created a bunch of user names during this workshop (PythonAnywhere, GitHub, superuser for their Django site, etc.) and they might also have a computer user name, so they all had to get help from coaches to figure this out.

PR coming your way...

jhodgdon-drp:
Fixes #1395

### DIFF
--- a/en/django_orm/README.md
+++ b/en/django_orm/README.md
@@ -83,17 +83,17 @@ What users do we have in our database? Try this:
 {% filename %}command-line{% endfilename %}
 ```python
 >>> User.objects.all()
-<QuerySet [<User: ola>]>
+<QuerySet [<User: admin>]>
 ```
 
-This is the superuser we created earlier! Let's get an instance of the user now (adjust this line to use your own username):
+This is the superuser we created earlier! Let's get an instance of the user now (adjust this line to use your own superuser's username):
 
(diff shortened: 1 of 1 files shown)


ekohl:
In django_admin it tells you `When prompted, type your username (lowercase, no spaces)` so it doesn't have to be admin. The rest of the tutorial tries to use ola when it can be customized so I'm not sure we should merge this. Thoughts?

jhodgdon-drp:
Check the issue report #1395... What happens is on the django_admin tutorial page, it says to create a user account, and the example uses "admin". Then when you get to the django_orm page, the example commands assume the user name is "ola". It should be consistent between the two pages.

I'm OK with changing django_admin instead of django_orm if that is what you prefer, but one of them definitely needs to be changed, because it confused several people in our tutorial in September.

jhodgdon-drp:
OK I see you prefer ola when I read your comment again. I'll close this PR and make another one that changes admin instead.

PR #1413, opened 2018-10-14
Add more resources to "What's next?"

eoinnoble:
After coaching a number of sessions over the last couple of years it's pretty clear that people come to the Django Girls tutorial for all kinds of reasons – maybe they want to create a website for their small business, maybe they want to use Python to help crunch data for their PhD – the audience is _really_ broad. With that in mind I wanted to flesh out and restructure the "What's next?" section of the tutorial to better help people find resources that are relevant to their interests so that they can better build on their experience of the tutorial.

The only resource I've removed from the original page is the book _Two Scoops of Django_ – I am a huge admirer of this book but I believe it is far too advanced to give to someone coming straight out of the Django Girls tutorial, and it costs money as well, unlike the other resources on the page.

I'm on the fence about adding links to Coursera and edX since there are paid-for components to their offerings, but I think there is enough free content there that they could still provide value to our audience.

Changes in this pull request:
- rewrote much of the text on the "What's next?" page
- added more free resources to the "What's next?" page
- grouped the resources into subgroups

### DIFF
--- a/en/whats_next/README.md
+++ b/en/whats_next/README.md
@@ -4,21 +4,36 @@ Congratulate yourself! __You're totally awesome__. We're proud! <3
 
 ### What to do now?
 
-Take a break and relax. You have just done something really huge.
+Take a break and relax! You have just done something really huge.
 
 After that, make sure to follow Django Girls on [Facebook](http://facebook.com/djangogirls) or [Twitter](https://twitter.com/djangogirls) to stay up to date.
 
 ### Can you recommend any further resources?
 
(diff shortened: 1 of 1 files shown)


eoinnoble:
@jhodgdon-drp thank you for the review! I agree with your points, the resources you mention were in the old version of the page and I left them in there in case people had strong feelings about them but my instinct was the same as yours.

I wonder if the data science resources are less relevant because data science itself seems less relevant to the DjangoGirls tutorial in the first place? I am always surprised when I come across an attendee who cites data science as their main interest, but it does happen.

I should say that if there are other areas of interest that coaches or attendees have encountered during their sessions I'd be really interested to hear about them!

jhodgdon-drp:
Looks good to me now! I think it should be merged.

PR #1417, opened 2018-10-22
Control + break for windows

evascholten:
It is introduced earlier in the tutorial that on windows, control + break can be used instead of control + c. From experience I know it is a good idea to repeat that here in the tutorial, as students may have forgotten. This will save them some confusion.

Changes in this pull request:

-
-
-


aniav:
I have updated the code to match the requested changes and have merged it from command line. Thanks! ✨

PR #1418, opened 2018-10-27
also name LibreOffice when comparing Git to MSWord "track changes" feature

das-g:
When comparing version control systems like Git to the "track changes" feature of word processor programs, not only the proprietary Microsoft Word (part of the Microsoft Office office suite) should be named, as the "track changes" feature is implemented in more than that one text processor / office suite. Also naming a FLOSS one (that is also available on FLOSS systems like GNU/Linux) seems required for completeness, and for the (probably still few) participants that never have used Microsoft software.

### DIFF
--- a/en/deploy/install_git.md
+++ b/en/deploy/install_git.md
@@ -1,4 +1,4 @@
-Git is a "version control system" used by a lot of programmers. This software can track changes to files over time so that you can recall specific versions later. A bit like the "track changes" feature in Microsoft Word, but much more powerful.
+Git is a "version control system" used by a lot of programmers. This software can track changes to files over time so that you can recall specific versions later. A bit like the "track changes" feature in word processor programs (e.g. Microsoft Word or LibreOffice Writer), but much more powerful.
 
 ## Installing Git
 
--- a/en/deploy/install_git.md
+++ b/en/deploy/install_git.md
@@ -1,4 +1,4 @@
-Git is a "version control system" used by a lot of programmers. This software can track changes to files over time so that you can recall specific versions later. A bit like the "track changes" feature in word processor programs (e.g. Microsoft Word or LibreOffice Writer), but much more powerful.
+Git is a "version control system" used by a lot of programmers. This software can track changes to files over time so that you can recall specific versions later. A bit like the "track changes" feature in word processor programs (e.g., Microsoft Word or LibreOffice Writer), but much more powerful.
 
 ## Installing Git
 


das-g:
> I would avoid using "e.g.". It is kind of problematic:
> - Many people confuse it with "i.e." and aren't really sure what it means

Indeed I, too, noticed many people (even native writers) using "e.g." instead of "i.e." and vice versa. I guess the problem is that they're both abbreviations of Latin phrases, the expanded version of which are never used in today's English neither written nor spoken, rather the English equivalents "for example" and "that is"/"that are"/"which means" would be used. Also, they can appear in the same contexts and switching them produces sentences that usually still make sense (at least syntactically) even though carrying a different meaning.

I have never though heard about the recommendation to avoid these abbreviations for these reasons.

In case we _do_ want to avoid them, here are the current occurrences:
```
$ git grep --count "[Ii]\.e\."
en/django_forms/README.md:1
en/django_urls/README.md:1
en/extend_your_application/README.md:1
en/html/README.md:1
es/django_urls/README.md:1
it/django_installation/instructions.md:1
it/django_models/README.md:3
it/html/README.md:1

$ git grep --count "[eE]\.g\."
en/django_start_project/README.md:2
en/python_installation/instructions.md:2
es/django_start_project/README.md:1
sk/python_installation/instructions.md:1
```

(I wasn't aware, these abbreviations are in use in other languages than Latin and English. Do they mean the same there?)

> - It really needs to be punctuated, in English, like one of these samples:
> 
> foo bar bang; e.g., blah blah blah
> foo bar bang (e.g., blah blah blah)
> foo bar bang. E.g., blah blah blah
> 
> In other words, you need a ; or ( before it, if it's in a sentence, and a comma afterwards.

Is that still so or has the use without comma become so widespread that it's nowadays considered correct, too? Anyway, I've added a comma, as it's certainly correct with one.

das-g:
For avoiding "e.g.", what would you suggest? Using "like"? Or expanding it to "for example"?

jhodgdon-drp:
In the Drupal project documentation, we have a policy of not using e.g. and i.e., and yes, we use "for example" or "like". But being an open-source project, sometimes i.e. and e.g. slip in there anyway. :)

Anyway, I'm fine with the latest PR. Or you could just say
... feature in word processors like Microsoft Word or LibreOffice Writer...

ekohl:
I'm neutral about the use of `e.g.` vs `such as` vs `like` though I do see the benefit of the simpler text.

das-g:
As I'm not aware of any such policy for Django Girls, please review and if appropriate merge this PR as-is, please.

If you'd like to establish such a policy for Django Girls, too, please open a separate issue for discussing that.

das-g:
> I'm approving now but not merging because we're in the middle of a workshop as I write this. Later I'll do a merge spree.

No hurry. :smiley: I just didn't want this to get blocked indefinitely due to the "e.g." debate.

ekohl:
Thanks!

PR #1422, opened 2018-11-03
Explicitly suggest to replace the content

ekohl:
A lot of students are unsure whether they should replace the content so they often just add to it. This change is intended to assure students that they can safely remove it.

I'm not sure this is the best phrasing. Suggestions welcome.


ekohl:
@jhodgdon-drp could you have a look?

PR #1431, opened 2018-11-06
Fix PythonAnywhere command in Turkish translation

Oedipus:
Changes in this pull request:

- Replace `git clone` command with the correct `pa_autoconfigure_django.py`


magul:
I saw that You already add this to Crowdin. Thanks!

PR #1434, opened 2018-11-17
correct expected code regarding include function

developingAlex:
The example code of what to expect in your mysite/urls.py file erroneously included the include import,
This change fixes that and updates the reference below to explicitly mention which line to add the include to.


developingAlex:
hey thanks for that you're right I've updated it now

ekohl:
Thanks!

PR #1444, opened 2018-11-24
Update Django Installation section on Windows

mignonnesaurus:
Changes in this pull request:
- Updated section of Virtual Environment in Django Installation section to include the installation and usage of `virtualenvwrapper-win` to ease the usage of `virtualenv` on Windows. 

Thank you!

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -34,11 +34,20 @@ $ python3 -m venv myvenv
 <!--sec data-title="Virtual environment: Windows" data-id="virtualenv_installation_windows"
 data-collapse=true ces-->
 
-To create a new `virtualenv`, you need to open the command prompt and run `python -m venv myvenv`. It will look like this:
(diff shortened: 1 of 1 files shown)


das-g:
AFAIK it was an explicit decision not to use virtualenvwrapper in the Django Girls tutorial for any of the systems, because it would be yet another tool to install and which needs to be explained.

If we want to change that, we should probably use it on all systems (i.e. Mac and Linux, too) and we should go through the whole tutorial to see if any other chapters need to be adapted due to that.

mignonnesaurus:
Hi @das-g ! 
Thank you very much for your review and input :) 

I tried to reproduce the issue we ran into at one of the workshops on a `Windows 10` machine but following the instructions worked in this version. The version of Windows at the workshop was different though and I did not take note of what version it was to be able to reproduce the problem. In that particular case `virtualenvwrapper-win` helped us move forward, however it was our latest resource. Given this, I understand it does not apply in general and hence it's not worth it to include it in the tutorial. 

I will close the PR. Thank you! :)

PR #1450, opened 2018-12-08
Added small note to use `python` instead of `python3` for Windows.

mignonnesaurus:
Changes in this pull request:

- Update command step to use `python` instead of `python3` since we are working within the virtual environment. 

Thank you! :)

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -167,6 +167,8 @@ Before we do that, we should make sure we have the latest version of `pip`, the 
 (myvenv) ~$ python3 -m pip install --upgrade pip
 ```
 
+**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) instead.
+
 ### Installing packages with requirements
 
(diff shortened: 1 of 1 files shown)


en/django_installation/instructions.md
```diff
@@ -167,6 +167,8 @@ Before we do that, we should make sure we have the latest version of `pip`, the
 (myvenv) ~$ python3 -m pip install --upgrade pip
 ```
 
+**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) instead.
+
```
das-g: As this is within the virtualenv, shouldn't we just use `python` on all operating systems? :thinking:

The tutorial says:
> When working within a virtual environment, `python` will automatically refer to the correct version so you can use `python` instead of `python3`.

just some paragraphs above

mignonnesaurus: Hello @das-g :) 
Yep, you are right. We can update this to use `python` instead of `python3` for all operating systems since we are working within a virtual env. 
Updated the PR. 
Thank you!

PR #1454, opened 2018-12-23
Add link to www.py4e.com

csev:
Hello, I am the teacher of the Python for Everybody Specialization on Coursera.  As you mention, the course is not exactly free because they put the assessments behind a paywall.   I have a 100% free and 100% open version of the entire specialization on my own www.py4e.com we site - you can do all the quizzes, the auto-graded assignments, and peer graded assignments for free on my site.

Changes in this pull request:

-
-
-


magul:
Thanks!

PR #1461, opened 2019-01-09
add slash to make code snippet consistent

SeeBenj:
Changes in this pull request:

- Code Fix:
added missing slash, so that code snippet is consistent with the immediately preceding one.

Other than claimed in the original description, skipping the slash doesn't cause any technical problems, but the inconsistency might confuse readers.

### original description:

Changes in this pull request:

- Code Fix:
added missing slash, since the original produced an error when implemented

### DIFF
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -91,7 +91,7 @@ We open `blog/urls.py` in the code editor and add a line:
 
 {% filename %}blog/urls.py{% endfilename %}
 ```python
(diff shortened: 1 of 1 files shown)


ekohl:
Which error did it produce? It is in line with the code below it where it does have a slash, but technically both should work I think.

SeeBenj:
I didn't note it down and I am unable to replicate it, so it was probably something else I did wrong.

PR #1462, opened 2019-01-14
fix MarkDown syntax in German translation

das-g:
(hopefully already fixed in Crowdin, although I'm not quite sure how that translates MarkDown to HTML and back again)

Changes in this pull request:

- fix MarkDown syntax in German translation
- add newline at end-of-file (GitHub's online editor did that automatically)

### DIFF
--- a/de/html/README.md
+++ b/de/html/README.md
@@ -91,17 +91,15 @@ Der Titel "Olas Blog" wird nun im Browser angezeigt. Hast du es bemerkt? Der Bro
 
 Wie du vielleicht bemerkt hast, hat jedes Element zu Beginn einen öffnenden Tag und einen zugehörigen *schließenden Tag* mit `/` und innerhalb davon sind Elemente *eingebettet*. Ein innerer Tag kann nicht außerhalb des umschließenden Tags geschlossen werden, die Reihenfolge muss immer stimmen.
 
-Es ist wie, wenn man Sachen in Kisten steckt. In die größte Box: `<html></html>` kommt eine kleinere Box: `<body></body>`, und dort wiederum packen wir viele kleine Kisten hinein: </code>`&lt;p&gt;&lt;/p&gt;</0>.</p>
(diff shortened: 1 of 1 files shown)


das-g:
In https://github.com/DjangoGirls/tutorial/issues/1468#issuecomment-457221686 @magul wrote:
> Just quick note here (as I don't want to hijack discussion here), I believe that what You want to resolve in #1462 is actually caused by Crowdin issue with rendering these few tags in destination language, so if You're figuring out this one You can ask them for support regarding this issue also.

Yeah, that's a good idea. I think we don't even have to wait until #1468 is figured out. **Can you (@magul) or some other "manager" of https://crowdin.com/project/django-girls-tutorial contact them, though?** I'm unfamiliar with our import/export procedure, so I can't give any information about that if it should be needed.

What you'll probably need to provide is a link to the affected string on Crowdin. You can get that in the translation view on Crowdin by clicking on <kbd><b>&vellip;</b></kbd> next to "Text for translation" when the affected string is selected in the left pane.

E.g. for German, it'd be https://crowdin.com/translate/django-girls-tutorial/313/en-de#9986

They may also want the source file, i.e. https://raw.githubusercontent.com/DjangoGirls/tutorial/master/en/html/README.md

das-g:
Though, may I suggest you merge this PR anyway, as a temporary fix?

It probably doesn't even need review from @DjangoGirls/de-pr, as it only changes (and fixes) markup and doesn't change any actual content.

magul:
@das-g I just merged it and agree that's only a temporary fix. We should work with Crowdin support if they are able to fix it or we should look for a workaround here (I will also ask them if they can provide one).

das-g:
Maybe Crowdin already noticed and fixed it. When I change the strings back to containing the appropriate placeholders instead of HTML-escaped gibberish, the translation I can get through the "download and upload" menu looks correct.

PR #1446, opened 2018-11-28
saner punctuation for parenthesized full sentences

das-g:
When a full (and without the parenthesized part already complete) sentence is followed by a parenthesized full complete sentence, then end the former with a period before the parenthesis and end the latter with a period within the parenthesis (before the closing brace). Also, start
the parenthesized sentence with a capital letter.

Same for stand-alone parenthesized sentences.

PR #1465, opened 2019-01-17
[en] mention possibly translated dir name for "Desktop" at first occurrence

das-g:
Changes in this pull request:

- Mention the possibility of the "Desktop" subfolder of Linux user home folder to be localized when that folder _first_ occurs in a command (in the command line intro chapter)
  - (For that, introduce separate expandable sections for Linux and OS X)
  - Give an example of a translated name. (I've chosen the German "Schreibtisch".)
- Adapt the existing remark at a later occurrence (in the python intro chapter) accordingly

This should ...
- ... help people following the English original of the tutorial despite having localized folder names. (E.g. because no (complete) translation of the tutorial is available for that language.)
- ... remind translators that they should mention how these folders are named on Linux accounts in the respective language.

### DIFF
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -202,11 +202,12 @@ data-collapse=true ces-->
 $ cd Desktop
 ```
 
-(The directory name "Desktop" might be translated
+Note that
+the directory name "Desktop" might be translated
 to the language of your Linux account.
 If that's the case, you'll need to replace `Desktop`
 with the translated name,
-for example `Schreibtisch` for German.)
+for example `Schreibtisch` for German.
 
 <!--endsec-->
 
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -185,7 +185,7 @@ Music
 
 Now, let's go to our Desktop directory:
 
-<!--sec data-title="Change directory: OS X" data-id="python_OSX" data-collapse=true ces-->
+<!--sec data-title="Change current directory: OS X" data-id="OSX_move_to" data-collapse=true ces-->
 
 {% filename %}command-line{% endfilename %}
 ```
@@ -193,7 +193,7 @@ $ cd Desktop
 ```
 <!--endsec-->
 
-<!--sec data-title="Change directory: Linux" data-id="python_linux" data-collapse=true ces-->
+<!--sec data-title="Change current directory: Linux" data-id="Linux_move_to" data-collapse=true ces-->
 
 {% filename %}command-line{% endfilename %}
 ```
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -205,7 +205,8 @@ $ cd Desktop
 (The directory name "Desktop" might be translated
 to the language of your Linux account.
 If that's the case, you'll need to replace `Desktop`
-with the translated name, for example `Schreibtisch` for German.)
+with the translated name,
+for example `Schreibtisch` for German.)
 
 <!--endsec-->
 
--- a/en/intro_to_command_line/README.md
+++ b/en/intro_to_command_line/README.md
@@ -185,8 +185,7 @@ Music
 
 Now, let's go to our Desktop directory:
 
-<!--sec data-title="Change directory: OS X" data-id="python_OSX"
-data-collapse=true ces-->
+<!--sec data-title="Change directory: OS X" data-id="python_OSX" data-collapse=true ces-->
 
 {% filename %}command-line{% endfilename %}
 ```
@@ -194,8 +193,7 @@ $ cd Desktop
 ```
 <!--endsec-->
 
-<!--sec data-title="Change directory: Linux" data-id="python_linux"
-data-collapse=true ces-->
+<!--sec data-title="Change directory: Linux" data-id="python_linux" data-collapse=true ces-->
 
 {% filename %}command-line{% endfilename %}
 ```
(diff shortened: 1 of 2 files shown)


en/intro_to_command_line/README.md
```diff
@@ -185,14 +185,30 @@ Music
 
 Now, let's go to our Desktop directory:
 
-<!--sec data-title="Change current directory: OS X and Linux" data-id="OSX_Linux_move_to" data-collapse=true ces-->
+<!--sec data-title="Change directory: OS X" data-id="python_OSX"
```
jhodgdon-drp: Was this separation into two lines intentional?

en/intro_to_command_line/README.md
```diff
@@ -185,14 +185,30 @@ Music
...
+
+{% filename %}command-line{% endfilename %}
+```
+$ cd Desktop
+```
+
+(The directory name "Desktop" might be translated
+to the language of your Linux account.
+If that's the case, you'll need to replace `Desktop`
+with the translated name, e.g. `Schreibtisch` for German.)
```
jhodgdon-drp: The Latin abbreviation "e.g." is problematic for many English speakers. They don't really know what it means, and it's often confused with "i.e.". Also, the punctuation here is incorrect -- it has to be punctuated as though it was "for example". So if you want to use "e.g." (which I recommend against), this line needs to be:

with the translated name; e.g., `Schreibtisch` for German.)

Also I am wondering why this whole paragraph is in parentheses? I think it would be better to take out the parens (didn't we have an issue/PR about that a month or so ago?) and start the paragraph with "Note that the directory...".

en/python_introduction/README.md
```diff
@@ -602,13 +602,15 @@ $ cd ~/Desktop
...
 
-On Linux, it will be like this (the word "Desktop" might be translated to your local language):
+On Linux, it will be like this:
 
 {% filename %}command-line{% endfilename %}
 ```
 $ cd ~/Desktop
 ```
 
+(Remember that the word "Desktop" might be translated to your local language.)
```
jhodgdon-drp: Again I think I would remove the () here.

jhodgdon-drp:
Here's the PR about parens: #1446

en/intro_to_command_line/README.md
das-g: Do you mean the separation in OS-X-specific and Linux-specific expandable sections instead of a common expandable section for OS X _and_ Linux? I've adopted that from [the original remark](https://github.com/DjangoGirls/tutorial/pull/1465/files#diff-0ce35d3afd321211a71112aa5f36d20aL591) which came in a later chapter and which is also being changed by this PR. And yes, adopting it was intentional, as I don't know whether OS X also has localized home default subfolder names.

Or do you mean the line break within the HTML comment (the GitBook pragma starting the expandable section)? That was also adopted from the other chapter, but unintentionally so. Does it matter? I can change it to a single line if wanted.

en/intro_to_command_line/README.md
das-g: > The Latin abbreviation "e.g." is problematic for many English speakers. They don't really know what it means, and it's often confused with "i.e.". Also, the punctuation here is incorrect -- it has to be punctuated as though it was "for example". So if you want to use "e.g." (which I recommend against), this line needs to be:
> 
> with the translated name; e.g., `Schreibtisch` for German.)

I've expanded the abbreviation to "for example". Is the current punctuation correct for that, or should it be "...<strong>;</strong> for example<strong>,</strong> ..." instead (or something yet else)?

das-g: > Also I am wondering why this whole paragraph is in parentheses?

The [original remark](https://github.com/DjangoGirls/tutorial/pull/1465/files#diff-0ce35d3afd321211a71112aa5f36d20aL605) was also in a parenthesis. (Even one of the dreaded full-sentence-within-another-full-sentence kind.) I adopted that here, because the remark is only relevant for part of the readership (those reading the English translation despite having their Linux account language set to something else), but now that you point it out I agree that "Note that ..." without parenthesis would be more appropriate. I'll change it accordingly.

das-g: > I agree that "Note that ..." without parenthesis would be more appropriate. I'll change it accordingly.

Done: 15723bfb0d23dd8897eba852733904fb48b78093

das-g:
> Here's the PR about parens: #1446

Yeah, I'll come back to that and take it back up eventually.

en/intro_to_command_line/README.md
jhodgdon-drp: I meant within the HTML comment. I don't care one way or the other, but it seems unrelated to the rest of this PR?

en/intro_to_command_line/README.md
jhodgdon-drp: ...; for example, ... 
is indeed the correct punctuation.

en/python_introduction/README.md
jhodgdon-drp: Well, we can come back to that on #1446 then.

jhodgdon-drp:
Aside from the "for example" punctuation, this looks fine to me now.

en/intro_to_command_line/README.md
das-g: fixed in ee1169a131f926439115bc98cdc3b9915a9bfecc

jhodgdon-drp:
I didn't notice that, oops! Does look correct now.

das-g:
> I didn't notice that, oops!

Me neither until I removed the line breaks in the pragmas. (After that, GitHub highlighted the changed words in the diff and it was more apparent. Dunno why it doesn't also (try to) do that when multiple lines have changed.)

PR #1466, opened 2019-01-17
correct language w.r.t. QuerySet method chaining

das-g:
Not QuerySets can be chained, but their method calls can.

Changes in this pull request:

- Mention that some QuerySet methods themselves return a QuerySet
- Correct language: It's the method( call)s that can be chained/combined, not several QuerySets.
- Change h3-section title to reflect this new content.

This is an alternative change to the one proposed at #1311. Other than #1311, this change **doesn't** try to explain how to combine several QuerySets with the `union` method, which isn't needed for the tutorial anyway.

### DIFF
--- a/en/django_orm/README.md
+++ b/en/django_orm/README.md
@@ -192,8 +192,11 @@ We can also reverse the ordering by adding `-` at the beginning:
 
 ### Complex queries through method-chaining
 
-As you saw, some QuerySet methods return a new QuerySet.
-You can combine their effect by **chaining** them together:
+As you saw, some methods on `Post.objects` return a QuerySet.
+The same methods can in turn also be called on a QuerySet,
+and will then return a new QuerySet.
+Thus,
+you can combine their effect by **chaining** them together:
 
 ```python
 >>> Post.objects.filter(published_date__lte=timezone.now()).order_by('published_date')
--- a/en/django_orm/README.md
+++ b/en/django_orm/README.md
@@ -190,9 +190,10 @@ We can also reverse the ordering by adding `-` at the beginning:
 ```
 
 
-### Chaining QuerySets 
+### Complex queries through method-chaining
 
-You can also combine QuerySets by **chaining** them together:
+As you saw, some QuerySet methods return a new QuerySet.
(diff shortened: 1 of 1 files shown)


das-g:
> I'm not sure that we really saw that "some QuerySet methods return a new QuerySet" in this topic. What we saw was that some methods on Post.objects return a QuerySet.

You're right. Thanks for catching that!

>But at least on this page of the tutorial, it isn't stated anywhere that Post.objects is itself a QuerySet (is it actually a QuerySet, or some kind of a collection object?)

`Post.objects` indeed isn't a QuerySet (thanks for reminding me, I had kinda forgotten about this subtlety), [it is](https://docs.djangoproject.com/en/2.0/ref/models/class/#objects) a ["manager"](https://docs.djangoproject.com/en/2.0/topics/db/managers/#django.db.models.Manager):
 
```python
>>> from django.db.models import QuerySet
>>> from django.db.models import Manager
>>> from blog.models import Post
>>>
>>> isinstance(Post.objects, QuerySet)
False
>>> isinstance(Post.objects, Manager)
True
```

> If it is in fact a QuerySet, and this is mentioned elsewhere in the tutorial, I think it should be stated at the top of this topic if this information is going to be used here.

Well, it isn't, although it often suspiciously behaves like one. That's because [it actually delegates many methods](https://github.com/django/django/blob/2.0.10/django/db/models/manager.py#L107) to a QuerySet. But as you suggest ...

> In any case... Maybe this would be easier to avoid that complication [...]

... I'd indeed rather avoid this complication in the tutorial.

> [...] and just say something like:
> 
> As you saw, when you run a query, the result is a QuerySet object. This object can also be queried! You could save it in a variable, or use what is called "method chaining", where you use the output of one query method as the input to the next query, all in one line:
> 
> Or something to that effect?

"run a query" would be misleading (or even wrong) here, as QuerySets are evaluated lazily, so the actual database query won't be evaluated until actually needed.

How about "construct a query"?

das-g:
Trying a different, more compact resolution: 39ff106

Please review again.

PR #1475, opened 2019-01-31
Fix command for installation on Debian and Ubuntu

andreagaredine:
This change fix the command line section for installing Python on Debian and it derivatives.

### DIFF
--- a/es/python_installation/instructions.md
+++ b/es/python_installation/instructions.md
@@ -60,7 +60,7 @@ Escribe este comando en tu consola:
 
 {% filename %}command-line{% endfilename %}
 
-    sudo yum install python3.6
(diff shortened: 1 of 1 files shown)


das-g:
Hi @andreagaredine

Good catch! Translations of the tutorial are maintained on Crowdin. Can you make the change, there, too?

andreagaredine:
> Hi @andreagaredine

Hi!

> Good catch! Translations of the tutorial are maintained [on Crowdin](https://crowdin.com/project/django-girls-tutorial). Can you make the change, [there](https://crowdin.com/translate/django-girls-tutorial/317/en-es#12402), too?

Done!

das-g:
Thank you very much!

PR #1477, opened 2019-02-05
Fix up PythonAnywhere cd instructions

### LINKED ISSUE #1474: "cd $USER.pythonanywhere.com" doesn't work for everyone
### Issue description

In [the re-deployment inststructions](https://tutorial.djangogirls.org/en/django_templates/#one-more-thing) in chapter "Django templates" we [currently let](https://github.com/DjangoGirls/tutorial/blame/51bd85d086567c3266dd4d215a8961f8d2c8051f/en/django_templates/README.md#L92) the user execute
```bash
cd $USER.pythonanywhere.com
```
on PythonAnywhere.

While this will change into the Git repository on PythonAnywhere for most users, it won't work for all. When a user has a PythonAnywhere username that isn't a valid domain name part (e.g. containing capital letters), the directory name will be different from the username + `.pythonanywhere.com`.

I'm not sure how to best resolve this. Should we change the instruction to
```bash
cd *.pythonanywhere.com
```
or try to imitate the username-to-subdomain conversion that `pa_autoconfigure_django.py` does?

Or is there some other environment variable or file that would tell us the correct repo name? (Maybe @hjwp knows?)

Or should we simply advice users to choose lowercase alphanumeric PythonAnywhere usernames?

### Language

(all)

### Operating system

n/a

jhodgdon-drp:
Fix #1474

### DIFF
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -415,12 +415,12 @@ $ git push
 
 {% filename %}PythonAnywhere command-line{% endfilename %}
 ```
-$ cd ~/<your-pythonanywhere-username>.pythonanywhere.com
+$ cd ~/<your-pythonanywhere-domain>.pythonanywhere.com
 $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-username>` with your actual PythonAnywhere username, without the angle-brackets).
+(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
 
 
 * Finally, hop on over to the ["Web" page](https://www.pythonanywhere.com/web_app_setup/) (use the menu button in the upper right of the console) and hit **Reload**. Refresh your https://yourname.pythonanywhere.com blog to see the changes.
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -423,7 +423,7 @@ $ git pull
 (Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
 
 
-* Finally, hop on over to the ["Web" page](https://www.pythonanywhere.com/web_app_setup/) (use the menu button in the upper right of the console) and hit **Reload**. Refresh your https://yourname.pythonanywhere.com blog to see the changes.
+* Finally, hop on over to the ["Web" page](https://www.pythonanywhere.com/web_app_setup/) (use the menu button in the upper right of the console) and hit **Reload**. Refresh your https://subdomain.pythonanywhere.com blog to see the changes.
 
 
 And that should be it! Congrats :)
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -420,7 +420,7 @@ $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
+(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere subdomain, without the angle-brackets.)
 
 
 * Finally, hop on over to the ["Web" page](https://www.pythonanywhere.com/web_app_setup/) (use the menu button in the upper right of the console) and hit **Reload**. Refresh your https://subdomain.pythonanywhere.com blog to see the changes.
--- a/en/extend_your_application/README.md
+++ b/en/extend_your_application/README.md
@@ -186,7 +186,7 @@ $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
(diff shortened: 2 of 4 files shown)


en/extend_your_application/README.md
```diff
@@ -181,12 +181,12 @@ Then, in a [PythonAnywhere Bash console](https://www.pythonanywhere.com/consoles
...
+$ cd ~/<your-pythonanywhere-domain>.pythonanywhere.com
 $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-username>` with your actual PythonAnywhere username, without the angle-brackets).
+(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
 
 
 ## Updating the static files on the server
```
das-g: There's another occurrence [further down](https://github.com/DjangoGirls/tutorial/blame/4aa58695a0e7b0124c682e90c6413d72906153a9/en/extend_your_application/README.md#L200) in this section.

en/html/README.md
```diff
@@ -201,12 +201,12 @@ $ git push
...
 {% filename %}PythonAnywhere command-line{% endfilename %}
 ```
-$ cd ~/<your-pythonanywhere-username>.pythonanywhere.com
+$ cd ~/<your-pythonanywhere-domain>.pythonanywhere.com
 $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-username>` with your actual PythonAnywhere username, without the angle-brackets).
+Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain name, without the angle-brackets. Your domain name is normally your PythonAnywhere user name, but in some cases it might be a bit different (such as if your user name contains capital letters). So if this command doesn't work, use the `ls` (list files) command to find your actual domain/folder name, and then `cd` to there.
```
das-g: In the order of the chapter in the tutorial, this (and not `en/django_templates/README.md`) seems indeed to be the first occurrence, so it makes sense to have the extended explanation here, as you suggest.

Though:
- Didn't you suggest to have the copy-pastable `$USER` solution at that first occurrence? (I'm fine with dropping that in favor of consistency. Just want to make sure whether it's a conscious decision.)
- Why "Remember to" if this substitution didn't occur in the tutorial before this point?

en/django_forms/README.md
```diff
@@ -415,12 +415,12 @@ $ git push
...
+$ cd ~/<your-pythonanywhere-domain>.pythonanywhere.com
 $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-username>` with your actual PythonAnywhere username, without the angle-brackets).
+(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
 
 
 * Finally, hop on over to the ["Web" page](https://www.pythonanywhere.com/web_app_setup/) (use the menu button in the upper right of the console) and hit **Reload**. Refresh your https://yourname.pythonanywhere.com blog to see the changes.
```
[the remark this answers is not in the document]
jhodgdon-drp: Yes indeed!

en/django_forms/README.md
```diff
@@ -415,12 +415,12 @@ $ git push
...
 {% filename %}PythonAnywhere command-line{% endfilename %}
 ```
-$ cd ~/<your-pythonanywhere-username>.pythonanywhere.com
+$ cd ~/<your-pythonanywhere-domain>.pythonanywhere.com
 $ git pull
 [...]
 ```
 
-(Remember to substitute `<your-pythonanywhere-username>` with your actual PythonAnywhere username, without the angle-brackets).
+(Remember to substitute `<your-pythonanywhere-domain>` with your actual PythonAnywhere domain, without the angle-brackets.)
```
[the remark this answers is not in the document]
jhodgdon-drp: Good idea. I'll fix that in all the occurrences.

en/extend_your_application/README.md
jhodgdon-drp: I think I got it...

en/html/README.md
jhodgdon-drp: Yeah, I thought it was better to be consistent. You're right about "remember to". That was in this page before and I left it, but you're right it should be taken out of this one.

jhodgdon-drp: Also, just to point out: the $USER was not actually in the first occurrence. It is in en/django_templates/README.md -- so I made that spot the same syntax as all the others already were (including this first occurrence, which is in the en/html file).

jhodgdon-drp:
OK, I think all the review changes have been addressed... take a look at the PR again now? Thanks for the great comments!

en/html/README.md
das-g: > Also, just to point out: the $USER was not actually in the first occurrence. It is in en/django_templates/README.md -- [...]

Yeah, that was what I was trying to point out. :+1:

Issue #1055, opened 2017-06-06
Navigation within the tutorial is confusing


pts:
It's quite confusing for many participants to reliably navigate to the next page of the tutorial. The `<` and `>` buttons are not prominent enough on screen, and they don't have a fixed location (they move around when the window width is changed).

My recommendation is adding explicit *Previous page* and *Next page* text links to the top and bottom of the page with a prominent color (e.g. blue).

My additional recommendation is adding this to the tutorial: Use the right arrow key to navigate to the next page of the tutorial, and the left arrow key for the previous page.

das-g:
#1495 didn't add additional links, but labeled them with text and improved their placement on small screens. (The color remained unchanged, though, as far as I can tell.)

Can you check, whether that satisfactorily resolves this issue?

pts:
Works like a charm, thanks!

Issue #1054, opened 2017-06-06
Screen real estate instructions missing


pts:
Many participants arrive to the Django Girls workshop with a laptop with very low screen resolution (e.g. 1280x768), and it's very inconvenient and frustrating for them to keep manually scrolling left-right and up-down for navigation.

One way to solve this problem is to reduce the amount of whitespace in the tutorial's CSS (or make it configurable). I'll file a separate issue for that.

My recommendation is extending the tutorial about using the screen efficiently, with recommendations like this:

1. Click on the gray `---` icon at the top left corner of the web page to hide the navigation menu.

2. Use the gray `A` icon and then the small `A` link at the top left corner of the page to make the font size smaller so that more text fits on your the screen.

3. Feel free to make your browser window showing the tutorial narrower, the tutorial will adjust to it.

das-g:
For the English version, can you check whether #1495 (now live on https://tutorial.djangogirls.org/en/) resolved this?

pts:
#1495 is a great improvement, but it doesn't directly address this issue: providing explicit hints on how to use small screens more efficiently.

PR #1495, opened 2019-02-25
Add text to prev/next and top menu icons, so they are easier to notice and understand

brondsem:
This addresses issues #1054 and #1055 although may not satisfy all the ideas/requests from the original reporter.  Further changes I think would involve gitbook core or theme customizations, which would be a much bigger undertaking.

I could also hide the social sharing icons in the upper right, to declutter that if we don't think they are useful.  I wasn't sure though.

### DIFF
--- a/en/styles/website.css
+++ b/en/styles/website.css
@@ -1,0 +1,57 @@
+
+/* add text to top menu items */
+
+.book-header .fa::after {
+    text-transform: none;
+    font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
+}
+.book-header .fa-align-justify::after {
+    content: ' Show Contents';
+    text-transform: none;
+    font-family: "Helvetica Neue",Helvetica,Arial,sans-serif;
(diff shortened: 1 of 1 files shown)


brondsem:
Anyone interested in this?  Here's how it will look on large screens:
screenshot 2019-04-02 22-31-21

And on small:
screenshot 2019-04-02 22-31-45

goanpeca:
Looks much better :-)

das-g:
The screenshots look good to me. :+1:

I don't know GitBook enough to review the CSS, but I'll just merge this now anyway. :shipit:

pts:
This is a great improvement! Some suggestions:

* For very narrow windows, Font Settings is rendered in a separate line (not a problem), and the main text becomes narrower (it's a problem) because of additional left margin.

* The label *Show Contents* is misleading, because the contents of the page is the main text. This should be *Show Outline* or *Show TOC*.

PR #1508, opened 2019-03-22
Installation subsection on menu

joopeed:
Today I have contributed to a Django Girls event in Wellington, NZ as a coach. It's a rather nice experience to see everyone sharing knowledge for this wonderful community. 
Also I got a chance to gather some feedback on this tutorial from some of my mentees. It's very complete, step by step, including all the installation steps, but one thing that kept happening among them when navigating back and forth on the menu was losing track of where they were. This mainly happens because there is one page for installing all components (Installation) and some of these have their own pages like Python. Sometimes, then, they would click on the Python Installation menu on the left thinking it would guide them to the same page, but this does not happen. 
To solve this small confusion a propose to include the subsections of installation in the menu as well so that they see where they are at installing.

Please let me know if I can give you any other details needed, thanks!


aniav:
This waited for a bit but I just merged it. I had to merge from command line as I have also added a plugin to (un)fold the menu as the newly introduced items were making everything hard to navigate. Thanks!

PR #1509, opened 2019-03-23
Django apps have their own urls.py file

sinistersnare:
Hey, I noticed this issue, and this could be fixed fairly easily. I did it only for English, I did not check any of the other languages. If you would like me to, I could.

Changes in this pull request:

- Add a line to django_models tutorial page


das-g:
> I did it only for English, I did not check any of the other languages.

That's OK.
> If you would like me to, I could.

We'll update translations through Crowdin, so no need for that.

Issue #1271, opened 2018-05-28
Document the translation proccess and translation maintenance process


das-g:
### Issue description

As far as I know, non-English versions of the tutorial are initially created on Crowdin by translating the English "original". However there seem to be several ways to change existing non-English versions:

- suggest changes on Crowdin (which probably makes sense for language-specific stuff, like phrasing, grammar, ...)
- pull requests on GitHub (which probably makes sense for changing a common thing (e.g. code) in several languages at once, so that they stay in sync)

What way is preferred for what kind of contributions and how do they interact? (When and how are changes from Crowdin integrated into the GitHub repo? When are new versions deployed to GitBook and tutorial.djangogirls.org?)

### Language

> Is this related to a specific language of the tutorial?

All except English

### Operating system

> What operating system does this issue relate to?

(not applicable)

ekohl:
For me the relation with Crowdin is unclear. I'm happy to merge typo changes that someone else has approved.

If we want to allow PRs on translations I'd suggest we start using labels for each language. In another project we run a bot to automatically assign labels based on the directory changes were made in. I could probably run that for Django Girls if there's interest/agreement on this.

The https://help.github.com/articles/about-codeowners/ can also be used to assign teams to each language though perhaps this can get in the way when you want to make a pure code change (like shell commands) in all languages. Then you'd need to get approval from every language maintainer.

magul:
I would share what I know regarding translation and Github <-> Crowdin relation:

From community perspective we have at least two documents, that try to describe, how to contribute translations here:
 * [README in this repository](https://github.com/DjangoGirls/tutorial#editing-basics)
 * [Django Girl's translation guide](https://github.com/DjangoGirls/translation-guide)
AFAIK none of them prohibit to contribute to translation using Github PR.

There's also at least two channels on Django Girls slack: #translations and #tutorial, where ad hoc discussion may occur (I rarely there, but I've just logged in and see e.g. discussion regardin Korean translation between @sujinleeme and @RachellCalhoun ).

For what I know, @RachellCalhoun is the primary person, that's responsible for translators community/translations policy. We spoke (also with @aniav ) around year ago about how to maintain translations. After initial hipotesis to abandon Crowdin at all, @RachellCalhoun returned with feedback, that major part of community want to use Crowdin, so it's still running.

From technical perspective we use two-way integration Github <-> Crowdin:
[...]

PR #1515, opened 2019-04-30
add GitHub CODEOWNERS using the existing PR review teams

das-g:
Add [GitHub CODEOWNERS](https://help.github.com/en/articles/about-code-owners) using the existing PR review teams from https://github.com/orgs/DjangoGirls/teams (an idea brought up by https://github.com/DjangoGirls/tutorial/issues/1271#issuecomment-392542583)

This will automatically request reviews of the respective teams for pull requests touching "their" files, but won't require these reviews for merging.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,17 @@
+# default:
+*	@DjangoGirls/committers
+
+# language PR review teams for the respective translations:
+/de/	@DjangoGirls/de-pr
+/en/	@DjangoGirls/en-pr
+/es/	@DjangoGirls/es-pr
+/fr/	@DjangoGirls/fr-pr
+/it/	@DjangoGirls/it-pr
+/ja/	@DjangoGirls/ja-pr
+/ko/	@DjangoGirls/ko-pr
+/pl/	@DjangoGirls/pl-pr
+/pt/	@DjangoGirls/pt-pr
+/ru/	@DjangoGirls/ru-pr
+/tr/	@DjangoGirls/tr-pr
+/uk/	@DjangoGirls/uk-pr
+/zh/	@DjangoGirls/cn-pr


magul:
May I suggest here to use the same language code for a resource prefixes and translators group on GitHub? We right now have `/zh/` as a prefix and `cn` as group distinction.

What do You think about that @aniav, @RachellCalhoun?

ekohl:
One downside of this is that you can't merge anything unless there's an ack from the translation team. Sometimes there are trivial changes and it would be blocked. There is https://probot.github.io/apps/auto-assign/ but I have no experience with it.

das-g:
> May I suggest here to use the same language code for a resource prefixes and translators group on GitHub? We right now have `/zh/` as a prefix and `cn` as group distinction.

The group name is probably due to a confusion, as `CN` isn't an ISO 639 **language** code at all. (It [is the 3166-2 **country** code](https://en.wikipedia.org/wiki/ISO_3166-2:CN) for the People's Republic of China.)

The [ISO 639-1 (=two-letter) **language** code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes) for the group of languages known as "Chinese"  is indeed `zh`. So I guess we should indeed rename that GitHub team.

I've filed a separate issue to discuss this further:  #1517

das-g:
> One downside of this is that you can't merge anything unless there's an ack from the translation team. Sometimes there are trivial changes and it would be blocked.

If I understood the [documentation](https://help.github.com/en/articles/about-code-owners) correctly, particularly

> When someone with admin or owner permissions has enabled required reviews, they also can optionally require approval from a code owner before the author can merge a pull request in the repository.

I'd say it isn't the case that by default, merges are blocked as long as no "code owner" has approved the PR. So I think users with write permission to the target repo would still be able to merge, even without an approval.

If I understand the quoted documentation part correctly, there would be a two-step opt-in for having a blocking code owner approval requirement:
- Option "Require pull request reviews before merging" would have to be enabled (it doesn't seem to currently be enabled for DjangoGirls/tutorial)
- Option "Require review from Code Owners" would have to be enabled (only available when the first option is enabled)

(See also steps 5 and 8 of https://help.github.com/en/articles/enabling-required-reviews-for-pull-requests)

Thus, adding the CODEOWNERS file should automatically _request_ reviews by the respective code owners, but not _require_ approval by them for merging.

PR #1523, opened 2019-05-22
Add text for those who have installed Anaconda

kaizumaki:
I added the text for the Anaconda user to the "Python Installation" chapter.
I followed the comments on this PR #1519 .

Thank you.

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -4,7 +4,7 @@
 
 Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install the latest version of Python 3, so if you have any earlier version, you will need to upgrade it. If you already have version 3.4 or higher you should be fine.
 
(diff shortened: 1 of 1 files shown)


en/python_installation/instructions.md
```diff
@@ -4,6 +4,7 @@
 
 Django is written in Python. We need Python to do anything in Django. Let's start by installing it! We want you to install the latest version of Python 3, so if you have any earlier version, you will need to upgrade it. If you already have version 3.4 or higher you should be fine.
 
+If you have Anaconda installed on your computer, please install normal Python as follows.
```
das-g: Those who don't have Anaconda installed, should of course install normal Python, too. Thus I'd suggest to phrase it with "even when":

```suggestion
Please install normal Python as follows, even when you have Anaconda installed on your computer.
```

kaizumaki: @das-g Thank you for fixing. I also think that is better.

PR #1525, opened 2019-05-29
Wrong variable name displays an error

KunihikoKido:
Changes in this pull request:

- Change to Wrong variable name

`ctiy` looks similar to the actual word `city` if you do not pay enough attention. Actually my co-worker who just started to do tutorial typed  `city` and wondered why it did not show as an error.

### DIFF
--- a/ja/python_introduction/README.md
+++ b/ja/python_introduction/README.md
@@ -230,10 +230,10 @@ name イコール（=）"Ola" とタイプします。
 
 ```python
 >>> city = "Tokyo"
->>> ctiy
+>>> citty
 Traceback (most recent call last):
   File "<stdin>", line 1, in <module>
-NameError: name 'ctiy' is not defined
+NameError: name 'citty' is not defined
 ```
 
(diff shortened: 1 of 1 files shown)


magul:
Hi @KunihikoKido, first of all, thank you for that change.

Can You, please, provide some context here, as I don't understand why `citty` is better than `ctiy`. Both of them can be considered as a typo and will raise the same error.

KunihikoKido:
Hi @magul 

`ctiy` looks similar to the actual word `city` if you do not pay enough attention. Actually my co-worker who just started to do tutorial typed  `city` and wondered why it did not show as an error.haha

das-g:
In the English version of the tutorial, this code block is preceded by some prose that introduces the intent:
> But what if we used the wrong name? Can you guess what would happen? Let's try!

To me, this seems sufficient. If this isn't the case in the Japanese version, maybe the translation needs to be improved regarding these sentences? If possible, I'd like to keep variable names consistent between the translations to various languages.

Thus, I'm voting to decline this PR, i.e. close it without merging it.

KunihikoKido:
OK. I understand.

PR #1553, opened 2019-08-27
Add note about pre-opened cmd on Windows

shaib:
Added it to the Note at the bottom about handling "command not found" on Windows.

I considered adding a note in the Windows installation instructions, telling the user to close any pre-open command prompt. This is probably better operationally (prevent the error instead of handle it), but I couldn't find a satisfactory way to put it in the narrative.


shaib:
Thanks!

PR #1561, opened 2019-09-07
fix capitalization of "URL"

das-g:
Changes in this pull request:

- in prose, spell "URL" all uppercase (rather than all lowercase "url" or with only initial capital "Url")
- when "url" refers to the module name (defined by `url.py`), mark it up as code (i.e. "`url`") 

I've found these occurrences with:
```bash
git grep -P '(?<!% |\.|_|`|c|/)[Uu]rl(?!s.py|patterns|resolver)' -- en/
```


aniav:
💃

PR #1562, opened 2019-09-07
set lang code in Polish tutorial to Polish (Poland)

krzysztofzuraw:
Changes in this pull request:

- change lang code in polish settings.py


das-g:
This wasn't a bug, as `de-ch` is just an example. But for non-English versions of the tutorial, it makes sense to use their own language as the example, so let's adopt this change! :+1:

das-g:
I forgot: Can you also add that change [on Crowdin](https://crowdin.com/translate/django-girls-tutorial/306/en-pl#9506), please?

aniav:
I changed it on Crowdin just now 👍

PR #1563, opened 2019-09-07
fix: rename lottery to polish equivalent

krzysztofzuraw:
Changes in this pull request:

- rename lottery to polish equivalent


das-g:
:+1: Well spotted!

Without this change, the Polish translation is inconsistent, as it already partially uses the Polish variable name for that list.

Can you please also add that change [on Crowdin](https://crowdin.com/translate/django-girls-tutorial/318/en-pl), please?

krzysztofzuraw:
@das-g done - thank you

PR #1581, opened 2019-10-06
Fix download url of VSCode

kaizumaki:
I modified the download URL for VSCode.
I think this URL is simpler to download.

Thank you.

### DIFF
--- a/en/code_editor/instructions.md
+++ b/en/code_editor/instructions.md
@@ -6,7 +6,7 @@ Our suggestions are below, but feel free to ask your coach what their preference
 
 Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS. It includes support for debugging, embedded Git control, syntax highlighting, intelligent code completion, snippets, and code refactoring.
 
-[Download it here](https://code.visualstudio.com/download)
+[Download it here](https://code.visualstudio.com/)
 
 ## Gedit
 


en/code_editor/instructions.md
```diff
@@ -6,7 +6,7 @@ Our suggestions are below, but feel free to ask your coach what their preference
 
 Visual Studio Code is a source code editor developed by Microsoft for Windows, Linux and macOS. It includes support for debugging, embedded Git control, syntax highlighting, intelligent code completion, snippets, and code refactoring.
 
-[Download it here](https://code.visualstudio.com/download)
+[Download it here](https://code.visualstudio.com/)
```
das-g: What's wrong with https://code.visualstudio.com/download ? That seems to work fine for me. In what regard is https://code.visualstudio.com/ "easier"? It seems to offer the same choices, but more additional information on what VisualStudio Code is.

kaizumaki: Please compare with Atom's download URL.
```
Download it here https://atom.io/
```
You can find a download button according to the OS you are using.
Similarly, in the case of VSCode, https://code.visualstudio.com/  is easier for beginners to understand that the download button corresponding to the OS being used is displayed.
https://code.visualstudio.com/download wonders which download button the user should press.

das-g: Makes sense. I didn't notice that difference before, but indeed on https://code.visualstudio.com/ the download for the detected operating system is pre-selected, while on https://code.visualstudio.com/download there's just a bunch of links.

PR #1584, opened 2019-10-07
Indicate that directory venv is also there when showing where to put requirements.txt

kaizumaki:
I added a note about the directory structure in the description of `requirements.txt` .
This is because beginners are confused because the description of the `venv` directory is not written.

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -183,9 +183,10 @@ First create a `requirements.txt` file inside of the `djangogirls/` folder, usin
 
 ```
 djangogirls
+├── myvenv
+│   └── ...
 └───requirements.txt
 ```
-> **Note**: in your directory structure, you will also see your `venv` directory.
 
 In your `djangogirls/requirements.txt` file you should add the following text:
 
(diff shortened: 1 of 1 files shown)


das-g:
> +1 I think this is very valuable, a few people in my group were also confused by this. I’d suggest adding it directly in the directory structure as "myvenv", as the tutorial calls it, but the note is useful too.
> 
> ```
> djangogirls
> ├───myvenv
> └───requirements.txt
> ```

I like that approach better than the note. Maybe we can even indicate that `myvenv` is a directory with more content in it:
```
djangogirls
├── myvenv
│   └── ...
└── requirements.txt
```

en/django_installation/instructions.md
```diff
@@ -185,6 +185,7 @@ First create a `requirements.txt` file inside of the `djangogirls/` folder, usin
 djangogirls
 └───requirements.txt
 ```
+> **Note**: in your directory structure, you will also see your `venv` directory.
```
das-g: If we stick to the note approach, please start the sentence with a capital letter:

```suggestion
> **Note**: In your directory structure, you will also see your `venv` directory.
```

kaizumaki: I prefer to modify the directory tree rather than writing notes. I fixed the pull request.

PR #1585, opened 2019-10-08
how to map github repository locally

Anjalizi:
Changes in this pull request:

- make changes in README.md in en/deploy/README.md
- add a faster alternate way to add remote in console

### DIFF
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -110,6 +110,12 @@ $ git remote add origin https://github.com/<your-github-username>/my-first-blog.
 $ git push -u origin master
 ```
 
+Alternatively, you could simply copy the clone URL by clicking on the button at the end of the bar and paste it in the console.
+```
+$ git remote add origin <paste-here>
+$ git push -u origin master
+``` 
+
 When you push to GitHub, you'll be asked for your GitHub username and password (either right there in the command-line window or in a pop-up window), and after entering credentials you should see something like this:
 
 {% filename %}command-line{% endfilename %}


en/deploy/README.md
```diff
@@ -110,6 +110,12 @@ $ git remote add origin https://github.com/<your-github-username>/my-first-blog.
 $ git push -u origin master
 ```
 
+Alternatively, you could simply copy the clone URL by clicking on the button at the end of the bar and paste it in the console.
```
ekohl: These instructions are very similar to the ones we already have which may be confusing. Is the current issue that finding the URL is difficult?

Anjalizi: It isn't actually difficult but copying a URL is any day simpler than writing it on your own.

das-g: The page of a freshly created GitHub repo actually state the exact two command lines we need here:
screenshot of newly created GitHub repo

Should we simply refer to those? If so, just to those two lines
> ```bash
> git remote add origin https://github.com/das-g2/my-first-blog.git
> git push -u origin main
> ```

or to the whole set of "…or create a new repository on the command line" instructions there?
> ```bash
> echo "# my-first-blog" >> README.md
> git init
> git add README.md
> git commit -m "first commit"
> git branch -M main
> git remote add origin https://github.com/das-g2/my-first-blog.git
> git push -u origin main
> ```

We might still want to explain what each line does, which could become confusing when GitHub changes what instructions they display.

The `git branch -M main` part though would solve #1676 without depending on Git versions &ge; 2.28, which aren't pervasive, yet.

PR #1589, opened 2019-10-11
Fix typo in div class name

PierreGuilmin:
I fixed a typo in the `django_forms` page. The tutorial suggests to create a `div` called `page-header` whereas the code underneath uses the name `top-menu`. The typo has been fixed for all the languages.

Lot of thanks for your wonderful tutorial 🎉

### DIFF
--- a/cs/django_forms/README.md
+++ b/cs/django_forms/README.md
@@ -42,7 +42,7 @@ Takže ještě jednou: budeme vytvářet odkaz na stránky, URL, pohled a šablo
 
 ## Odkaz na stránku s formulářem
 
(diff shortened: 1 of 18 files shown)


das-g:
@PierreGuilmin, @magul Is it OK if I revert this change until this is cleared up?

PierreGuilmin:
Sure @das-g !

PierreGuilmin:
You're right @das-g, there is some ambiguity across the different translations. I think the best phrasing would be:
>It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link with class name `top-menu`:

PR #1591, opened 2019-10-11
rephrase sentence to fix "page header" ambiguity

das-g:
rephrase sentence to make clear which element the name `page-header` refers to. (It's the `div`, not the link.)

This should fix the ambiguity noticed in PR #1589, which has been reverted with PR #1590.

See https://github.com/DjangoGirls/tutorial/pull/1589#pullrequestreview-300581137

/cc @PierreGuilmin

### DIFF
--- a/en/django_forms/README.md
+++ b/en/django_forms/README.md
@@ -44,7 +44,7 @@ So once again we will create a link to the page, a URL, a view and a template.
 
 ## Link to a page with the form
 
-It's time to open `blog/templates/blog/base.html` in the code editor. We will add a link in `div` named `page-header`:
+It's time to open `blog/templates/blog/base.html` in the code editor. In `div` named `page-header`, we will add a link:
 
 {% filename %}blog/templates/blog/base.html{% endfilename %}
 ```html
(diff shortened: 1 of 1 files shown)


en/django_forms/README.md
```diff
@@ -44,7 +44,7 @@ So once again we will create a link to the page, a URL, a view and a template.
 
 ## Link to a page with the form
 
-It's time to open `blog/templates/blog/base.html` in the code editor. We will add a link in `div` named `page-header`:
+It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link:
```
das-g: @PierreGuilmin suggested in https://github.com/DjangoGirls/tutorial/pull/1589#issuecomment-541002831
```suggestion
It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link with class name `top-menu`:
```

While I'm not sure whether we need that level of explicitness, I'm not against it, either. What are others' opinions on this?

das-g: As there hasn't been any feedback, I'll merge this PR as-is for now. Further changes can be suggested in new PRs.

PR #1598, opened 2019-10-17
changed the version number and link for most updated version of Pytho…

lwyso:
…n for macOS

Changes in this pull request:

-
-
-

### DIFF
--- a/en/python_installation/instructions.md
+++ b/en/python_installation/instructions.md
@@ -36,10 +36,10 @@ data-collapse=true ces-->
 
 > **Note** Before you install Python on OS X, you should ensure your Mac settings allow installing packages that aren't from the App Store. Go to System Preferences (it's in the Applications folder), click "Security & Privacy," and then the "General" tab. If your "Allow apps downloaded from:" is set to "Mac App Store," change it to "Mac App Store and identified developers."
 
-You need to go to the website https://www.python.org/downloads/release/python-361/ and download the Python installer:
+You need to go to the website https://www.python.org/downloads/release/python-374 and download the Python installer:
 
-* Download the *Mac OS X 64-bit/32-bit installer* file,
-* Double click *python-3.6.1-macosx10.6.pkg* to run the installer.
+* Download the *macOS 64-bit/32-bit installer* file,
+* Double click *python-3.7.4-macosx10.6* to run the installer.
 
 <!--endsec-->
 


das-g:
As long as the code doesn't use features from the newer Python versions, it should be compatible to older ones. So using Python 3.7 or 3.8 locally on macOS while still using Python 3.6 on PythonAnywhere should work fine, as long as we only use Python 3.6-compatible source code, which we have to anyway, as long as participants on some other operating systems will use Python 3.6 locally for the tutorial.

thibaudcolas:
@das-g I think that’s true in theory but it might be confusing to participants that they are using different versions in different places, and it doesn’t seem that far-fetched to me that the tutorial might be updated to use newer Python features without realising that PythonAnywhere is still on 3.6.

It probably doesn’t matter much but it’s not like Python 3.6 is going away anytime soon so it also doesn’t hurt to stay on it until the whole tutorial can be switched over to a newer version.

thibaudcolas:
For reference here is the Python 3.5 to 3.6 PR: #1086.

das-g:
> @das-g I think it will be confusing to participants that they are using different versions in different places,

That might indeed be an issue. Note though, that the Windows instructions already tell the participants to install the latest stable version of Python, without referring to a specific version number, while the Linux instructions tell you that Python 3.6 or any version above will be fine (which should in fact be the case).

> and it doesn’t seem that far-fetched to me that the tutorial might be updated to use newer Python features without realising that PythonAnywhere is still on 3.6.

I don't think there's much danger of that happening, as those making those changes would hopefully also test them on PythonAnywhere and as those reviewing them would hopefully test them on other Python versions. What _could_ happen is that when trying some stuff for themselves during or after the tutorial, participants might start to use features that work on their local python version but not on the one on Python anywhere. But that too seems unlikely to happen to total novices who wouldn't be able to understand that situation and for more advanced learners it can be a valuable learning experience of its own.

das-g:
> It probably doesn’t matter much but it’s not like Python 3.6 is going away anytime soon so it also doesn’t hurt to stay on it until the whole tutorial can be switched over to a newer version.

Agreed.

@lwyso Are you OK if we close this pull request without merging it for now? There's probably not much value in keeping it open until PythonAnywhere supports a newer Python version.

Or do we want to make the macOS instructions similar to the Windows and Linux ones, telling that 3.6 _and above_ can be used and for installation just tell them to install the latest stable without referring to a specific version number?

fltfx:
closing due to no action

PR #1603, opened 2019-10-21
German translation improvements from Crowdin made by:

magul:
* Raphael Das Gupta (https://crowdin.com/profile/das-g)


de/django_installation/instructions.md
```diff
@@ -120,7 +120,7 @@ Starte deine virtuelle Umgebung, indem du Folgendes eingibst:
 >         Execution Policy Change
 >         The execution policy helps protect you from scripts that you do not trust. Changing the execution policy might expose you to the security risks described in the about_Execution_Policies help topic at http://go.microsoft.com/fwlink/?LinkID=135170. Do you want to change the execution policy? [Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "N"): A
 >     
-
+>
```
das-g: These should remain two separate blockquote blocks (one per note), as in the English version. (I'm unsure whether GitBooks markdown dialect supports that by just having an empty line between the blocks. GitHub's MarkDown rendering respects the empty line as a blockquote block separator. If that isn't sufficient, we should probably add an HTML comment between them (also in the English original).)

```suggestion
<!-- (This comment separates the two blockquote blocks, so MarkDown doesn't merge them into a single one.) -->
```

magul: Few issues here:
 * This paragraph was introduced originally in #1315 in a different place. It was moved here in this particular spot in #1593.
 * After #1593 these two paragraphs were merged into one blockquote in English version (see:
Screenshot from 2019-10-23 18-42-35
)
* Don't really know why French translation from Crowdin render there
```
>

>
```
and German one insists on putting there
```
>
>
>
```
That looks like 
* Unfortunately you cannot see the problem on https://tutorial.djangogirls.org because the newest build on GitBook is date September 24th and all further webhooks from GitHub end with Service Timeout error.  Maybe we should start working on moving away from GitBook Legacy (as it's legacy) ASAP?

That's not a solution for problem pointed here, but probably we should address at least first bullet point here before merging that translation?

das-g: While it won't fix the underlying problem, I think an HTML comment between the blockquote blocks should suffice to keep them interpreted as separate subsequent blocks in most MarkDown dialects / for most MarkDown parsers, most probably including those of GitBook legacy and Crowdin (whichever version). So for now, let's do that as a workaround: #1610

magul: I've just refreshed it, so, please, do another review round.

PR #1624, opened 2019-12-08
Portuguese error and note

Sphinxs:


magul:
Hi @Sphinxs , first thing - thanks for your contribution here!

We also want ask you to apply your improvements in our Crowdin project, as we tend to maintain our translations there. Portuguese translation is available  here: https://crowdin.com/project/django-girls-tutorial/pt-BR#.

Thanks in advance!

PR #1628, opened 2020-01-12
[Spanish] [CSS] Fix selector of headings when changing font family

Sumolari:
With previous snippet the result did not match the screenshot.

Also, the text before mentions modifying a CSS snippet previously implemented as:

```css
h1 a, h2 a {
     color: #C25100;
}
```


magul:
Thanks guys!

PR #1635, opened 2020-01-24
Clarify calling functions explanation

lindboe:
The original explanation says that we re-write the name of the example function at the bottom of the file "because Python reads the file from top to bottom, so we need to re-write the name of the function." I find that explanation a bit confusing myself; the primary reason that we need to write `hi()` is because we need to execute it, not because of Python's ordering properties.

I think two things were conflated here:
1) The difference between defining and running a function
2) Explaining that Python reads files from top to bottom, and needs functions to be defined before they can be called

I tried to more explicitly explain that what we're doing is calling a function we just defined, and break these concepts apart more. Very open to feedback on ways to do this better.

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -815,7 +815,7 @@ hi()
 
 Okay, our first function is ready!
 
-You may wonder why we've written the name of the function at the bottom of the file. When we write `def hi():` and the indented lines following, this is us writing instructions for what the `hi()` function should do. Python will read and remember these instructions, but won't run the function yet. To tell Python we want to run the function, we have to call the function with `hi()`. Python reads the file and executes it from top to bottom, so we have to define functions in the file before we call them.
+You may wonder why we've written the name of the function at the bottom of the file. When we write `def hi():` and the indented lines following, this is us writing instructions for what the `hi()` function should do. Python will read and remember these instructions, but won't run the function yet. To tell Python we want to run the function, we have to call the function with `hi()`. Python reads the file and executes it from top to bottom, so we have to define the function in the file before we call it.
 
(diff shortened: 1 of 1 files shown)


en/python_introduction/README.md
```diff
@@ -815,7 +815,7 @@ hi()
 
 Okay, our first function is ready!
 
-You may wonder why we've written the name of the function at the bottom of the file. This is because Python reads the file and executes it from top to bottom. So in order to use our function, we have to re-write it at the bottom.
+You may wonder why we've written the name of the function at the bottom of the file. When we write `def hi():` and the indented lines following, this is us writing instructions for what the `hi()` function should do. Python will read and remember these instructions, but won't run the function yet. To tell Python we want to run the function, we have to call the function with `hi()`. Python reads the file and executes it from top to bottom, so we have to define functions in the file before we call them.
```
das-g: For simplicity, I'd only mention this one function:

```suggestion
You may wonder why we've written the name of the function at the bottom of the file. When we write `def hi():` and the indented lines following, this is us writing instructions for what the `hi()` function should do. Python will read and remember these instructions, but won't run the function yet. To tell Python we want to run the function, we have to call the function with `hi()`. Python reads the file and executes it from top to bottom, so we have to define the function in the file before we call it.
```

Readers can hopefully themselves conclude, that this goes for each other function, too. And
> [W]e have to define functions in the file before we call them.

could be misinterpreted such as that all functions have to be defined before you can call even one of them, which isn't true. One could define one, calls it, define the next one, call that, etc.

lindboe: Great points! I like this suggestion a lot, I will commit it.

PR #1644, opened 2020-02-14
Clarify which name should be the same

shaib:


en/extend_your_application/README.md
```diff
@@ -81,7 +81,8 @@ Do you remember what the next step is? It's adding a view!
 
 ## Add a post's detail view
 
-This time our *view* is given an extra parameter, `pk`. Our *view* needs to catch it, right? So we will define our function as `def post_detail(request, pk):`. Note that we need to use exactly the same name as the one we specified in `urls` (`pk`). Omitting this variable is incorrect and will result in an error!
+This time our *view* is given an extra parameter, `pk`. Our *view* needs to catch it, right? So we will define our function as `def post_detail(request, pk):`. Note that we need the parameter to have exactly the same name as the one we specified in `urls` (`pk`). Omitting this variable is incorrect and will result in an error!
```
ekohl: How about this version?
```suggestion
This time our *view* is given an extra parameter, `pk`. Our *view* needs to catch it, right? So we will define our function as `def post_detail(request, pk):`. Note that this parameter must have the exact same name as the one we specified in `urls` (`pk`) and omitting this variable is incorrect. It will result in an error!
```

shaib: works for me just as well.

das-g: I think the previous sentence boundaries were more in line with text's content. So how about

```suggestion
This time our *view* is given an extra parameter, `pk`. Our *view* needs to catch it, right? So we will define our function as `def post_detail(request, pk):`. Note that this parameter must have the exact same name as the one we specified in `urls` (`pk`). Omitting this variable is incorrect and will result in an error!
```
or
```suggestion
This time our *view* is given an extra parameter, `pk`. Our *view* needs to catch it, right? So we will define our function as `def post_detail(request, pk):`. Note that this parameter must have the exact same name as the one we specified in `urls` (`pk`). Also note that omitting this variable is incorrect and will result in an error!
```

shaib: The exact phrasing does not matter to me. What bugged me is the vague reference to "the same name" when, in this context, it also happens that the view function's name and the URL name are the same (although that's not really necessary) so I thought we should clarify that we're talking about the parameter name.

ekohl:
Looks like Travis is still running, but other than that :+1: to merge.

PR #1645, opened 2020-02-14
Give more details about 'pk' field and underlying 'id'

shaib:


en/extend_your_application/README.md
```diff
@@ -41,7 +41,7 @@ We will start with adding a link inside `blog/templates/blog/post_list.html` fil
 
 The `post_detail` part means that Django will be expecting a URL in `blog/urls.py` with name=post_detail
 
-And how about `pk=post.pk`? `pk` is short for primary key, which is a unique name for each record in a database. Because we didn't specify a primary key in our `Post` model, Django creates one for us (by default, a number that increases by one for each record, i.e. 1, 2, 3) and adds it as a field named `pk` to each of our posts. We access the primary key by writing `post.pk`, the same way we access other fields  (`title`, `author`, etc.) in our `Post` object!
+And how about `pk=post.pk`? `pk` is short for primary key, which is a unique name for each record in a database. Every Django model has a field which serves as its primary key, and whatever other name it has, it can also be referred to as "pk". Because we didn't specify a primary key in our `Post` model, Django creates one for us (by default, a field named "id" holding a number that increases by one for each record, i.e. 1, 2, 3) and adds it as a field to each of our posts. We access the primary key by writing `post.pk`, the same way we access other fields  (`title`, `author`, etc.) in our `Post` object!
```
ekohl: I feel that the different uses of `name` (a unique name for each record vs whatever other name it has) can be confusing. One refers to the "name" for a row and the other the name of a field.

Technically the numbers are not guaranteed to be incrementing by one and there can be gaps so the `by one` can be left out.

How about this:

```suggestion
And how about `pk=post.pk`? `pk` is short for primary key, which is a unique identifier for each record in a database. Every Django model has a field which serves as its primary key, and whatever other name it has, it can also be referred to as "pk". Because we didn't specify a primary key in our `Post` model, Django creates one for us (by default, a field named "id" holding a number that increases for each record, i.e. 1, 2, 3) and adds it as a field to each of our posts. We access the primary key by writing `post.pk`, the same way we access other fields  (`title`, `author`, etc.) in our `Post` object!
```

PR #1662, opened 2020-05-23
Added description for the .gitignore file

aaryanporwal:
Description of the .gitignore file avoids common problem of not finding it with normal `ls` command.

The green underlined sentence will be added from the picture after merge.


das-g:
This change adds the sentences
> The .gitignore file is known as a 'dotfile' and you won't be able to see the file with normal `ls` command! Instead use `ls -a` to see the `.gitignore` file.

en/deploy/README.md
```diff
@@ -48,7 +48,7 @@ db.sqlite3
 
 And save it as `.gitignore` in the "djangogirls" folder.
 
-> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`.
+> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`. The .gitignore file is known as a 'dotfile' and you won't be able to see the file with normal `ls` command! Instead use `ls -a` to see the `.gitignore` file.
```
das-g, line 51: > and you won't be able to see the file with normal `ls` command

Is this true on Windows (with `cmd`), too?

das-g, line 51: > Instead use `ls -a` to see the `.gitignore` file.

That probably doesn't work on Windows' `cmd` unless the Unix tools brought along by GitBash were made available to `cmd`.

das-g, line 51: > The .gitignore file is known as a 'dotfile'

Is that categorization relevant to workshop participants? While we don't want to withhold essential information from them, we also don't want to overload them with unnecessary jargon.

Maybe a more useful statement (for most participants) is that Linux and MacOS (or for that matter, all "Unix-like systems"; but that's another piece of jargon we avoid) consider files whose name starts with a dot hidden.

das-g, line 51:
```suggestion
> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`. The `.gitignore` file is known as a 'dotfile' and you won't be able to see the file with normal `ls` command! Instead use `ls -a` to see the `.gitignore` file.
```

aaryanporwal: I'll check it out for windows

aaryanporwal: It happened with my friend on Mac, he wasn't able to see the .gitignore file and kept on creating more!

aaryanporwal: Well a line like - "If you're not able to find the .gitignore file in MacOS and Linux, it's because they are hidden files and to find them use the `ls -a' command!" will solve the common problem!

en/deploy/README.md
```diff
@@ -48,7 +48,7 @@ db.sqlite3
 
 And save it as `.gitignore` in the "djangogirls" folder.
 
-> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`.
+> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`. The `.gitignore` file is known as a 'dotfile' and you won't be able to see the file with normal `ls` command! Instead use `ls -a` to see the `.gitignore` file.
```
das-g, line 51: How about phrasing it as follows?

```suggestion
> **Note** The dot at the beginning of the file name is important!  If you're having any difficulty creating it (Macs don't like you to create files that begin with a dot via the Finder, for example), then use the "Save As" feature in your editor; it's bulletproof. And be sure not to add `.txt`, `.py`, or any other extension to the file name -- it will only be recognized by Git if the name is just `.gitignore`.
Linux and MacOS treat files with a name that starts with `.` (such as `.gitignore`) as hidden
and the normal `ls` command won't show these files.
Instead use `ls -a` to see the `.gitignore` file.
```
That way, it
- avoids the term "dotfile" (which I don't feel contributes much here)
- explains the general pattern (filename begins with `.` &rarr; file is hidden on Linux and MacOS)
- makes explicit that this is specific to Linux and MacOS

aaryanporwal: Yeah it looks good!

Issue #1652, opened 2020-04-08
GitBook Legacy will be shutdown


magul:
![Screenshot from 2020-04-08 21-06-04](https://user-images.githubusercontent.com/1165138/78823658-57715580-79dd-11ea-8283-aeda5fb70dc7.png)

das-g:
Thanks for letting us know. Can you copy in the text as text instead of screenshot, please?

magul:
Sure, I thought about just after posting image.

> Hi Tomasz Magulski,
> We are reaching out to you because after the 30th of April, one year after being deprecated, our legacy version of GitBook, legacy.gitbook.com, will be shutdown.
The brand new app.gitbook.com, now live for more than two years, offers a much better experience for both your readers and your team.
You can learn more about the complete list of differences in our official documentation.
> Since you first signed up to GitBook on May 7th 2017, you have created or contributed to content on legacy.gitbook.com.
Until the 30th of April, you can easily migrate your content in one click to the new version using the banner on your profile at https://legacy.gitbook.com/@magul.
> After the 30th of April, your public content will be archieved and remain accessible using its public URL, but will no longer be accessible on legacy.gitbook.com.
We invite you to read our blog article to learn more about what's new and what remains the same on the new version of GitBook.
> If you are having any trouble migrating your content, we encourage you to reach out to our support team by responding to this email directly, attaching your legacy username magul.
Hoping to see you soon on app.gitbook.com, we wish you all the best.
> Best regards,
> The GitBook team

RachellCalhoun:
Here is some more info about migration: 
https://docs.gitbook.com/resources/gitbook-legacy

magul:
I haven't yet found any information about GitBook legacy to be removed altogether but we need to be aware of #1664 here.

Maybe we should for a time being move hosting from GitBook to GitHub Pages (where static assets will be build on Travis CI). What do you think about that?

@aniav @das-g @RachellCalhoun ?

ekohl:
> Maybe we should for a time being move hosting from GitBook to GitHub Pages (where static assets will be built on Travis CI). What do you think about that?

I don't think we should be using Travis CI. Github Actions are more reliable and faster in my experience.

However, it's probably the quickest way to get back online now. Migrating to modern GitBook might be harder.

aniav:
[...]

PR #1666, opened 2020-06-27
Add a Github action to deploy

ekohl:
This replaces Travis as the integration tool with Github Actions. The benefit is that it can also deploy to the gh-pages branch. This can then be deployed automatically after merge.

It also switches from Gitbook legacy to Honkit. Since Github legacy is being phased out, this is needed.

This will address most of #1652. After it's merged, the admins still need to do set up of the Github Pages. Most importantly a DNS change.

I only did a quick scan of the result (see https://ekohl.github.io/tutorial/) and it looks good to me. It also has some nice additional features (dark mode, quick language switcher). A more thorough read through might be in order though.


.bookignore
```diff
@@ -0,0 +1,7 @@
+.bookignore
+.github
+.gitignore
+.travis.yml
```
magul, line 4: no reason for that

.gitignore
```diff
@@ -2,5 +2,6 @@ MANIFEST
 .DS_Store
 _book
 node_modules
+package-lock.json
```
magul, line 5: that should be tracked in the repository to ensure we have deterministic build

ekohl: I considered that, but I generally hate having such huge files in a repo. You just end up with a lot more maintenance to keep it up to date. It also greatly pollutes your git log.

I'd like to get some more opinions from people around this. Normally I stay as far away from NPM and the Javascript ecosystem as I can but here I recognize we don't want to rewrite the entire tutorial.

.bookignore
ekohl: Perhaps I should have separated that, but having a single way to build, test and deploy was a good idea IMHO.

magul: My argument here is that there's no need to ignore that particular path, as it's not use used and you have removed that file.

.gitignore
magul: Without that file, we're exposing ourselves into breaking changes when anything upgrades (which actually will increase the maintenance rate, at least forced maintenance).

To keep it up to date we should probably introduce something like @dependabot (which will also check on every pull request that upgraded dependencies will pass our "tests").

Leaving it without `package-lock.json` we're asking for failing `master` builds for no reason.

.bookignore
ekohl: Oh good point - this was a leftover from an earlier iteration. Will update.

.gitignore
ekohl: Right now we just install the latest gitbook (legacy). The plugins are also not all pinned. Not sure how much difference it would be since we've hardly seen build failures (AFAIK).

das-g: > I'd like to get some more opinions from people around this.

TL;DR: Do track `package-lock.json` in Git.

Here's my arguments for it in detail (mostly the same as what @magul already mentioned):

> I considered that, but I generally hate having such huge files in a repo. [...] It also greatly pollutes your git log.

Pinning versions with a lock file might not be pretty, but it's a necessity.

> You just end up with a lot more maintenance to keep it up to date.

We're using the software in question in our build process. If it isn't always the latest and greatest, no sweat. This won't usually cause any security issues for us. If it causes security issues for GitHub actions, GitHub will probably tell us.

On the other hand, we can probably blindly bump the versions to the latest whenever we feel like it and just see whether the result still works. The upside of having the lock file checked in, is, that whenever it turns out to not work as intended, we have known-working sets of versions that we can roll back to. Without that, we'd be in a situation of knowing it once worked but probably without any idea how to get it back to that state until upstream has fixed all issues **and** until we've eliminated all incompatibilities of our content with the new versions.

However rare that might be, having a lock file around for these situations will be well worth its downsides.

> Right now we just install the latest gitbook (legacy). The plugins are also not all pinned.

That's fine if the now-latest versions work for us. (And it seems they do.) Whenever other latest versions are around that might not work, we'll want to be able to go back to these now-latest (then not-anymore-latest) known working versions.

Lock files aren't usually used to carefully curate what versions are to be used. (That's what version ranges and other constraints in `package.lock` or equivalent files are for, if one chooses to do that.) They're there to be able to roll back to (or simply stay on) known-working version combinations, whatever they might be, whenever the latest versions aren't working.

aniav:
Just wanted to let you know I've seen this and I'll try to find time to get through the conversations and code this week and make relevant changes for DNS when we merge 🙌

aniav:
Ah, it's actually not that easy with DNS, as my account on Django Girls 1pass got suspended. Will update the DNS when I get access to it 🙌

aniav:
@das-g ah, sorry, @magul just made me aware of what happened, that your requested changes didn't actually get addressed. So to add to the voices here already: yes, we should lock the packages versions and keep them in the repository. For now, it's not a huge issue that I merged the PR without that change, I will make sure to apply this, so I'm not reverting. Just wanted to say I didn't ignore anyone intentionally 🙌

ekohl:
I'll try to get back to it, but wouldn't mind if someone else beats me to it.

It at least looks like it correctly deploys to `gh-pages`. Current https://djangogirls.github.io/tutorial redirects to the canonical domain so you can't really preview it.

@aniav you can also disable Travis in case you didn't already.

aniav:
Travis disabled. Domain enabled 👌

ekohl:
Looks like it was migrated correctly and is now served via Github Pages.

aniav:
🎉 🎉 🎉

PR #1667, opened 2020-07-12
Add example of method result comparison

henziger:
As the following paragraph states "As you can see, not only can Python compare numbers, but it can also compare method results." I think it makes sense to include a comparison of method results in the example itself. Since len() has been used earlier in the chapter its behavior should be familiar to the reader already.

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -460,6 +460,8 @@ True
 True
 >>> 5 != 2
 True
+>>> len([1, 2, 3]) > len([4, 5])
+True
 ```
 
 We gave Python some numbers to compare. As you can see, not only can Python compare numbers, but it can also compare method results. Nice, huh?


das-g:
Good catch. :+1:

:thinking: Though I'm not sure whether your suggested addition (alone) is the right remedy for this inconsistency.

The phrasing
> As you can see, Python can compare not only numbers, but it can also compare method results.

was introduced all the way back with 6a4a6945de0f66785f882e4a50ff1589f43445b8. Even back then, [that section](https://github.com/DjangoGirls/tutorial/blob/6a4a6945de0f66785f882e4a50ff1589f43445b8/try_python/README.md#compare-things) didn't contain any function invocations, let alone method invocations: https://github.com/DjangoGirls/tutorial/blob/6a4a6945de0f66785f882e4a50ff1589f43445b8/try_python/README.md#L195-L208

(Note that `len(...)` is a _function_ invocation, not a _method_ invocation. It'll call the `__len__()` method of the object passed as the argument, but that's an advanced perspective that we don't need to impose on beginners just yet. Other than in Java, not all functions in Python are methods.)

@olasitarska Might it be that by "method result" you just meant the value of a non-literal expression, as was already present with `2 * 2`?

In any case, showing that function results can be compared is valuable, but I think we should change the prose, too.

PR #1678, opened 2020-11-11
Migration to Django 3.1 version of Tutorial

nikhiljohn10:
### Migration updates

- [x] Updated Django version in book.json to 3.1.3
- [x] Updated Python version to 3.9.0
- [x] Updated Path library syntax for Django settings
- [x] Updated year in dates (Thought this would give users a feel of the updated tutorial)
- [x] Updated naming convention of apps in INSTALLED_APPS
- [x] Updated git usage for Github
- [x] Updated Python version in pythonanywhere to 3.8
- [x] Updated Bootstrap version to 4.5.3 without Theme
- [x] Updated final html and css scripts
- [x] Updated .gitignore file
- [x] Replaced glyphicon with Bootstrap Icons as SVG tags
- [x] Added HTML doctype as per HTML5 to Templates
- [x] Partially updated screenshot images. (Need to updated further)
- [x] Fixed bash prompt in `Deploy!`
- [x] Migrated Django related URLs from 2.2 to 3.1 version


das-g:
Thank you for taking this up! :+1:

das-g (changes requested):
Besides changes that are on-topic, this pull request makes quite some changes that seem completely unrelated to the Django version used, such as:

- changing the initialized repo's default Git branch to `main` to match GitHub's new default
- taking advantage of revised behavior of `git add .` in newer Git versions (other than in earlier versions, it now also "adds" removals)
- adding an HTML DOCTYPE
- changing the indentation of the HTML in our templates
- changing a CSS color (make to make it consistent across the tutorial?)
- changing the capitalization of the label "published: " on the blog

Many of these changes might be worth making. But unless they fall into one of the following categories, **please file them as separate pull requests directly towards `master`**, so that the scope of this pull request can remain limited to the switch to Python 3.1 and to an amount of changes that is more easily reviewable, and so that changes unrelated to the Django version change can be reviewed and accepted independent of it. Changes in this pull request here and in future pull requests towards branch `django-3.1` should be any, that:

- change the Django version to 3.1
- are required for or due to the change of the Django version (e.g. different Python versions)
- make use of new features that become available due to the above types of changes (e.g. `pathlib.Path`-based paths in settings instead of `os.path.join`) or due to previous Django updates
- are a result of the above changes (e.g. changed terminal outputs)

For any changes that fall into these categories, but where it's not obvious that they do, please leave an explanation for why they're on-topic in the pull request description or in comments in the pull request.

das-g:
>> NOTE: This update is only for the English language.

:+1:  That's good, as we'll handle updating the translations through Crowdin.

nikhiljohn10:
@das-g I will create seperate pull request for the topics metioned and revert those commit from this PR.

PR #1680, opened 2020-11-12
Git Updated

nikhiljohn10:
### Git updates

- [x] Updated branch from master to main
- [x] Updated git add command 
- [x] Updated .gitignore

This is a followup from #1678

### DIFF
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -80,7 +80,7 @@ And finally we save our changes. Go to your console and run these commands:
 
 {% filename %}command-line{% endfilename %}
 ```
-$ git add --all .
+$ git add .
 $ git commit -m "My Django Girls app, first commit"
  [...]
  13 files changed, 200 insertions(+)
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -24,7 +24,7 @@ Git tracks changes to a particular set of files in what's called a code reposito
 
 {% filename %}command-line{% endfilename %}
 ```
-$ git init
+$ git init -b main
 Initialized empty Git repository in ~/djangogirls/.git/
 $ git config --global user.name "Your Name"
 $ git config --global user.email you@example.com
@@ -60,7 +60,7 @@ It's a good idea to use a `git status` command before `git add` or whenever you 
 {% filename %}command-line{% endfilename %}
 ```
 $ git status
-On branch master
+On branch main
 
 No commits yet
 
@@ -111,7 +111,7 @@ Type the following into your console (replace `<your-github-username>` with the 
 {% filename %}command-line{% endfilename %}
 ```
 $ git remote add origin https://github.com/<your-github-username>/my-first-blog.git
-$ git push -u origin master
+$ git push origin main
 ```
 
 When you push to GitHub, you'll be asked for your GitHub username and password (either right there in the command-line window or in a pop-up window), and after entering credentials you should see something like this:
@@ -122,8 +122,8 @@ Counting objects: 6, done.
 Writing objects: 100% (6/6), 200 bytes | 0 bytes/s, done.
 Total 3 (delta 0), reused 0 (delta 0)
 To https://github.com/ola/my-first-blog.git
- * [new branch]      master -> master
-Branch master set up to track remote branch master from origin.
+ * [new branch]      main -> main
+Branch main set up to track remote branch main from origin.
 ```
 
 <!--TODO: maybe do ssh keys installs in install party, and point ppl who dont have it to an extension -->
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -135,7 +135,7 @@ Type the following into your console (replace `<your-github-username>` with the 
 {% filename %}command-line{% endfilename %}
 ```
 $ git remote add origin https://github.com/<your-github-username>/my-first-blog.git
-$ git push origin main
+$ git push -u origin main
 ```
 
 When you push to GitHub, you'll be asked for your GitHub username and password (either right there in the command-line window or in a pop-up window), and after entering credentials you should see something like this:
--- a/en/deploy/README.md
+++ b/en/deploy/README.md
@@ -36,14 +36,38 @@ Git will track changes to all the files and folders in this directory, but there
 
 {% filename %}.gitignore{% endfilename %}
 ```
+# Python
 *.pyc
 *~
-/.vscode
 __pycache__
-myvenv
+
+# Env
+.env
+myvenv/
+venv/
+
+# Database
 db.sqlite3
-/static
+
+# Static folder
+static/
+
+# macOS
+._*
 .DS_Store
+.fseventsd
+.Spotlight-V100
+
+# Windows
+Thumbs.db*
+ehthumbs*.db
+[Dd]esktop.ini
+$RECYCLE.BIN/
+
+# Visual Studio
+.vscode/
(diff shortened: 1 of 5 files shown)


en/deploy/README.md
```diff
@@ -111,7 +111,7 @@ Type the following into your console (replace `<your-github-username>` with the
 {% filename %}command-line{% endfilename %}
 ```
 $ git remote add origin https://github.com/<your-github-username>/my-first-blog.git
-$ git push -u origin master
+$ git push origin main
```
ekohl, line 138: Why did you drop `-u`? The tracking is quite useful.

nikhiljohn10: Fixed it

en/html/README.md
```diff
@@ -167,10 +167,10 @@ Make sure you're in the `djangogirls` directory and let's tell `git` to include
 
 {% filename %}command-line{% endfilename %}
 ```
-$ git add --all .
+$ git add .
 ```
 
-> __Note__ `--all` means that `git` will also recognize if you've deleted files (by default, it only recognizes new/modified files). Also remember (from chapter 3) that `.` means the current directory.
+> __Note__ By default, `git add` will recognize if you've deleted files (by default, it only recognizes new/modified files). Also remember (from chapter 3) that `.` means the current directory.
```
ekohl, line 173: Do you recall which version of git changed this?

nikhiljohn10: https://github.com/git/git/blob/e31aba42fb12bdeb0f850829e008e1e3f43af500/Documentation/RelNotes/2.0.0.txt#L106

nikhiljohn10: Check line 106

ekohl: I'm getting old. Somehow I thought git 2.0 was much more recent. Given it was released in 2014 I think we can safely assume everyone has it by now.

ekohl (commented):
I don't see any issues with this (though I still don't see why we need to rename from master). Leaving it open for other reviewers to take a look.

nikhiljohn10:
> I don't see any issues with this (though I still don't see why we need to rename from master). Leaving it open for other reviewers to take a look.

Every commits I did in this PR as well as my other PR is from the current feedbacks i found from Djangogirls gitter chat and my personal experience of using Djangogirls tutorial in the beginning of my Django learning.

What I learnt is that it creates more confusion for users to see one thing in their pc and different in tutorial. Since tutorial depends 100% on Github.com for deployment, I highly recommend updating `master` to `main` inside the tutorial.

I have read the discussion on changing to `main` branch on development side of tutorial and I agree that it is not needed while contributing to Tutorial. But I feel otherwise about it, inside tutorial.

en/html/README.md
das-g, line 173: This is contradictory. Should probably be
```suggestion
> __Note__ By default, `git add` will recognize if you've deleted files. Also remember (from chapter 3) that `.` means the current directory.
```

das-g, line 173: Should be mentioned that `git add` only behaves this way when given a path argument? Or could the note be dropped completely?

das-g (changes requested):

en/html/README.md
das-g: (Resolved conversation, as it seems unlikely that Git versions from before 2014 are in use by programming newcomers.)

nikhiljohn10:
@ekohl  The option -b which we used with `git init` is described below from https://git-scm.com/docs/git-init:

> -b <branch-name>
> --initial-branch=<branch-name>
>     Use the specified name for the initial branch in the newly created repository. If not specified, fall back to the default name: master.

This means that the version of git does not affect the update I proposed.

en/html/README.md
nikhiljohn10: Your suggestion is accurate.

nikhiljohn10: @das-g   I just went to https://tutorial.djangogirls.org/en/html/ and read it and it is not needed actually if `--all` option is not mentioned.

So I think it would not make any difference if it is there or not

das-g:
> @ekohl The option -b which we used with `git init` is described below from https://git-scm.com/docs/git-init:
> 
> > -b 
> > --initial-branch=
> > Use the specified name for the initial branch in the newly created repository. If not specified, fall back to the default name: master.
> 
> This means that the version of the git of the user does not affect the update I proposed.

It seems like `-b` / `--initial-branch` was added in Git [2.28.0](https://git-scm.com/docs/git-init/2.28.0) released 2020-07-27. The [documentation for Git 2.27.0](https://git-scm.com/docs/git-init/2.27.0) doesn't mention it, yet.

en/deploy/README.md
```diff
@@ -24,7 +24,7 @@ Git tracks changes to a particular set of files in what's called a code reposito
 
 {% filename %}command-line{% endfilename %}
 ```
-$ git init
+$ git init -b main
```
das-g, line 27: The `-b` option was introduced to `git init` with Git 2.28 in 2020-07-27. I guess we can't yet assume that everyone has that or a newer version of Git.

nikhiljohn10: @das-g  I agree.

Let's say I am a new user and I am using Ubuntu focal (20.04LTS). So I have git version 2.25.1. Now I executed following commands as in tutorial:

```
$ git init
$ git config --global user.name "Your Name"
$ git config --global user.email you@example.com
$ git add .
$ git commit -m "My Django Girls app, first commit"
$ git remote add origin https://github.com/<your-github-username>/my-first-blog.git
$ git push -u origin master
```

So now I assume there is a new branch called `master` in Github along with `main` branch.

I think the problem comes in Pythonanywhere.com:

(PythonAnywhere command-line)
```
$ pip3.6 install --user pythonanywhere
$ pa_autoconfigure_django.py --python=3.6 https://github.com/<your-github-username>/my-first-blog.git
```

Now the pythonanywhere script pulls the main branch from Github because this is the default branch set in Github. So, the command will propably result in "Could not find your settings.py".

**NOTE:** I have not tested this theory. But what if this happens? Correct me if I went wrong somewhere.

das-g: I've tried that with a new GitHub account: https://github.com/das-g2/my-first-blog

That series of actions seems to set `master` as the new repo's default branch. (I didn't change that setting manually in the repo settings page.) So at least for now, that doesn't seem to be a problem.

nikhiljohn10: I think this might be the temporary fallback solution. This is provided to all user when a new repo is made on GIthub now.

```
echo "# my-first-blog" >> README.md
git init
git add README.md
git commit -m "first commit"
git branch -M main
git remote add origin git@github.com:<username>/my-first-blog.git
git push -u origin main
```

@das-g try check this

nikhiljohn10: > I've tried that with a new GitHub account: https://github.com/das-g2/my-first-blog
> 
> That series of actions seems to set `master` as the new repo's default branch. (I didn't change that setting manually in the repo settings page.) So at least for now, that doesn't seem to be a problem.

I just checked too. So nothing to worry. @das-g Should I close this pull request?

das-g: I'm not sure. The more comprehensive `.gitignore` and dropping `--all` for `git add .` might still be worthwhile changes already now.

nikhiljohn10: @das-g check PR #1682

PR #1681, opened 2020-11-12
Migration to html5 and bootstrap4

nikhiljohn10:
### HTML and Bootstrap updates

- [x] Updated bootstrap link URLs
- [x] Updated DOCTYPE as per HTM5 standards
- [x] Updated CSS for Bootstrap4
- [x] Added container to page-header
- [ ] Fixed text case inside HTML
- [x] Updated example year to 2020 in the tutorial
- [x] Added actions division for alignment
- [x] Converted Glyphicon to Bootstrap4 Icons

HTML, Bootstrap and Icons have to be updated together. Otherwise, either one will make the page look broken.

This is a followup from #1678


en/django_forms/README.md
```diff
@@ -44,32 +44,38 @@ So once again we will create a link to the page, a URL, a view and a template.
...
 ## Link to a page with the form
 
-It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link:
+It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `container` inside `page-header` `div` tag, we will add a link:
 
 {% filename %}blog/templates/blog/base.html{% endfilename %}
 ```html
-<a href="{% url 'post_new' %}" class="top-menu"><span class="glyphicon glyphicon-plus"></span></a>
+<a href="{% url 'post_new' %}" class="top-menu">
+    <svg width="1em" height="1em" viewBox="0 0 16 16" class="bi bi-file-earmark-plus" fill="currentColor" xmlns="http://www.w3.org/2000/svg"><path d="M4 0h5.5v1H4a1 1 0 0 0-1 1v12a1 1 0 0 0 1 1h8a1 1 0 0 0 1-1V4.5h1V14a2 2 0 0 1-2 2H4a2 2 0 0 1-2-2V2a2 2 0 0 1 2-2z"/><path d="M9.5 3V0L14 4.5h-3A1.5 1.5 0 0 1 9.5 3z"/><path fill-rule="evenodd" d="M8 6.5a.5.5 0 0 1 .5.5v1.5H10a.5.5 0 0 1 0 1H8.5V11a.5.5 0 0 1-1 0V9.5H6a.5.5 0 0 1 0-1h1.5V7a.5.5 0 0 1 .5-.5z"/></svg>
```
ekohl, line 52: From a developer perspective I understand this, but if I was a student this would confuse me completely. I think this deserves a bit more explanation.

nikhiljohn10: I have thought about that. As it conserned to bootstrap4, they dont provide a icon package as css. This is where complexity begun. Next best option is font awesome. It is very similar to what it already have(glyphicon). But they have change the way of providing CND link publicly. I can still get the link from other CDN provides.

Then I thought of pushing the default icons of Bootstrap as SVG initially and change it to font awesome if required with an aditional commit to this PR.

If we are doing that, we need to fix up the best possible version of Font Awesome to be use.

One more thing is about using `integrity` and `crossorigin` attributes in `<link>` tags. @ekohl As you said, this decreases the readablility of code and increase complexity for users.

If we were to keep using the Bootstrap4 icons instead of Font awesome, I really feel like saving these svg files as assets and using it from `{% static 'img/icon-new-post.svg' %}` with `` tag or `{% include './icons/new-psot.svg' %}`. The descriptive information regarding the same have to be added as well.

nikhiljohn10: @ekohl Please suggest how to proceed?

ekohl: > One more thing is about using `integrity` and `crossorigin` attributes in `<link>` tags. @ekohl As you said, this decreases the readablility of code and increase complexity for users.

I don't mind these too much. On https://getbootstrap.com/ they use the same thing so it'd be the same thing if they were to follow that.

> If we were to keep using the Bootstrap4 icons instead of Font awesome, I really feel like saving these svg files as assets and using it from `{% static 'img/icon-new-post.svg' %}` with `` tag or `{% include './icons/new-psot.svg' %}`. The descriptive information regarding the same have to be added as well.

This sounds like a nice solution.

nikhiljohn10:
@ekohl I have tested all commits above by doing the updated tutorial from the stagging site https://django.nikz.in/en

image
---
image
---
image
---

There is some additional code I have added to the Django site shown above. The base.html file has a vacant space on the right side when using `col-md-8` for layout. I have filled up this vacant space using a bootstrap card.

```
# base.html
...
        <div class="content container">
            <div class="row">

                <div class="col-md-8">
                  {% block content %}
                  {% endblock %}
                </div>

                <div class="col-md-4">
                    <div class="card">
                      <div class="card-body">
                        <h5 class="card-title">Ads</h5>
                        <p class="card-text">Advertisements goes here</p>
                        <a href="#" class="btn btn-default">Learn more</a>
                      </div>
                    </div>
                </div>

            </div>
        </div>
...
```

Since this code is not inside the above commit, I would like your suggestion on it. Should I add it and push it into this PR?

Following is the current structure of the blog app:

image

en/django_forms/README.md
```diff
@@ -44,32 +44,58 @@ So once again we will create a link to the page, a URL, a view and a template.
...
+</a>
 ```
 
-Note that we want to call our new view `post_new`. The class `"glyphicon glyphicon-plus"` is provided by the bootstrap theme we are using, and will display a plus sign for us.
+Note that we want to call our new view `post_new`. The [svg icon](https://icons.getbootstrap.com/icons/file-earmark-plus/) is provided by the [Bootstrap Icons](https://icons.getbootstrap.com/) and it will display a page icon with plus sign.
 
-After adding the line, your HTML file should now look like this:
+The addition of SVG code for the icon made it look very complicated. So we can try a different approach to achieve the same result.
+
+Let's save the svg code in a separate file. Save the following code in the file `blog/templates/blog/icons/add-post.svg`
```
das-g, line 60: I'd rather not make the tutorial longer than it already is. I'd also rather not introduce yet another language (We already have at least 4: command line (Bash or CMD), Python, HTML and CSS.) and thus would prefer to treat SVG just as an image format, without showing the SVG file's source code.

So can we let participants
- use the SVG as file without first having it inline in the template's HTML?
- download the whole SVG file (e.g. from https://raw.githubusercontent.com/twbs/icons/main/icons/file-earmark-plus.svg) and save it at the right location, rather than copying or typing its textual content?

For interested participants, we can link to https://en.wikipedia.org/wiki/Scalable_Vector_Graphics or some other information about the format. That'd be less overwhelming than the wall of ~~text~~ code we'd place before them with the changes currently suggested here.

en/css/README.md
```diff
@@ -130,7 +129,7 @@ Your file should now look like this:
 
         {% for post in posts %}
             <div>
-                <p>published: {{ post.published_date }}</p>
+                <p>Published: {{ post.published_date }}</p>
```
das-g, line 132: Is there a reason to change the capitalization of "published:"? (I'm not saying it shouldn't be changed, I'm simply wondering why it shall be changed.)

en/django_forms/README.md
nikhiljohn10: Will do that

en/css/README.md
nikhiljohn10: No reason. It looked cleaner. I'll remove all those changes of "published:".

ekohl:
I don't think we should encourage ads. It's fine to keep it small and empty if it means we don't overwhelm students. Perhaps you can add a side bar in the extension tutorial.

nikhiljohn10:
> I don't think we should encourage ads. It's fine to keep it small and empty if it means we don't overwhelm students. Perhaps you can add a side bar in the extension tutorial.

@ekohl How about this much code to just fill the blank space?
```
# base.html
...
        <div class="content container">
            <div class="row">

                <div class="col-md-8">
                  {% block content %}
                  {% endblock %}
                </div>

                <div class="col-md-4">
                    <div class="card">
                      <div class="card-body">
                        <h5 class="card-title">Side Bar</h5>
                        <p class="card-text">Add sidebar content here</p>
                      </div>
                    </div>
                </div>

            </div>
        </div>
...
```

das-g:
> > I don't think we should encourage ads. It's fine to keep it small and empty if it means we don't overwhelm students. Perhaps you can add a side bar in the extension tutorial.
> 
> @ekohl How about this much code to just fill the blank space?

I like @ekohl's idea of making that a [tutorial extension](https://tutorial-extensions.djangogirls.org/).

The more often we let participants cut and paste and the longer the cut-and-pasted code blocks are, the less likely it is that participants will look at these code carefully, try to understand them and maybe ask their coaches about what those code blocks' purpose is and how the code works. **The goal of the Django Girls tutorial is to make programming and software development approachable and to make it seem less like obscure magic.** So it shouldn't be like a lengthy and opaque incantation, that has been passed down to you by higher powers and that you have to exactly reproduce to mystically attain its intended effect. Instead, it should provide participants with a glimpse of how software development works (including aspects other than programming, like revision control and deployment) and provide them with a chance of understanding what they're told to do, so that they can become creative themselves.

Therefore, the Django and web development part of the Django Girls tutorial proper should stick to only making a ["minimal viable product"](https://en.wikipedia.org/wiki/Minimum_viable_product) for the blog software, and to getting that deployed. (Actually, what the tutorial has the participants currently build, is not even "minimal viable" but just "minimal" in even aspects that are more central than a polished layout with perfectly used screen estate, such as security. E.g., for the sake of brevity, the site's secret key is simply left card-coded and pushed into a public GitHub repo.)

nikhiljohn10:
@das-g That is a valid point. I'll follow that concept with the content. 

@ekohl In that light, I modified "col-md-8" to just "col". It will expand the content to 100% width inside container. Otherwise we can make it "col-md-12". 

Either way, that vacant space has to go. It troubled my focus when I started django girls tutorial at my learning stage.

```
# base.html
...
        <div class="content container">
            <div class="row">

                <div class="col">
                  {% block content %}
                  {% endblock %}
                </div>
...
```

en/css/README.md
nikhiljohn10: Reverted changes in https://github.com/DjangoGirls/tutorial/pull/1681/commits/6f899d338caa430613984553ae065cfc69465496

en/django_forms/README.md
nikhiljohn10: Updated https://github.com/DjangoGirls/tutorial/pull/1681/commits/56d846c0746237e1cf3f2758d5cf3d5d67cf4bce

nikhiljohn10: Please check the updated changes in the following sections:

https://django.nikz.in/en/django_forms/#link-to-a-page-with-the-form

https://django.nikz.in/en/django_forms/#edit-form

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
 
 ## Link to a page with the form
 
-It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link:
+Before we add the link, we need some icons to use as buttons for the link. For this tutorial, download [file-earmark-plus.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/file-earmark-plus.svg) and save it in the folder `blog/templates/blog/icons/`
+
+> Note: To download the svg image, click on the link and save the file using the keyboard shortcut `Ctrl + S`(Windows/Linux) or `Cmd + S`(MacOS).
```
das-g, line 49: Another (maybe better?) option: Open the context menu on the link (usually by right-click on it) and choose "Save as ..."

das-g, line 49: Please write the abbreviation with capital letters:
> SVG

where it occurs in prose. (Don't change it in file names, as they are case sensitive on Linux.)

das-g (changes requested):
Thank you! Much less overwhelming now, I think.

IMO only minor issues left:

en/django_forms/README.md
nikhiljohn10: @das-g I was thinking same thing. But then I remember users use different browsers and they have different content menu. For example, firefox show "Save link as". So if we use context menu, do we have to cover all browsers's context menu option to save?

en/django_forms/images/file-earmark-plus.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="1em" height="1em" viewBox="0 0 16 16" class="bi bi-file-earmark-plus" fill="currentColor" xmlns="http://www.w3.org/2000/svg">
```
ekohl, line 1: I don't think this file is used anywhere, right?

nikhiljohn10: You are correct. It was initially added to provide download link and found a bug. So changed link to an absolute url from GitHub. I'll remove it.

nikhiljohn10: Resolved in https://github.com/DjangoGirls/tutorial/pull/1681/commits/a8b94e64223709fdddf5c36bf9510722cf764155

en/django_forms/README.md
nikhiljohn10: Resolved in https://github.com/DjangoGirls/tutorial/pull/1681/commits/4d6ffbc4920caa307ba7a8e13619ce8b8237f4fc

nikhiljohn10: @das-g I tested the save from link option on the latest stable distributions of Firefox, Chrome, Edge, Safari, and IE 11. Found "Save link as" for all the browsers except Internet Explorer 11. Internet Explorer 11 has "Save target as" instead.

So should I change the note as follows?

> Note: To download the SVG image,  open the context menu on the link (usually by right-clicking on it) and choose "Save link as" option. Now you can navigate to the desired directory in the window on your screen and save the file.

das-g: Yes, that seems like a more natural way to save the file.

Funny that IE's menu entry label is actually more accurate, as it _is_ the link _target_ that's being saved (the file the link points at), and indeed not the link itself. :upside_down_face: (Off course, our description should use the more common "Save link as")

nikhiljohn10: Resolved in https://github.com/DjangoGirls/tutorial/pull/1681/commits/27483eb7bf7703eb6c02e51aaaf9702ceae631b8

nikhiljohn10:
@ekohl @das-g All the review changes have been updated.

en/css/README.md
```diff
@@ -273,11 +299,11 @@ in the `blog/templates/blog/post_list.html` with this:
 ```html
 <div class="content container">
     <div class="row">
-        <div class="col-md-8">
+        <div class="col">
             {% for post in posts %}
                 <div class="post">
                     <div class="date">
-                        <p>published: {{ post.published_date }}</p>
```
ekohl, line 280: What's the reason to remove `published` here?

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
...
 {% filename %}blog/templates/blog/base.html{% endfilename %}
 ```html
-<a href="{% url 'post_new' %}" class="top-menu"><span class="glyphicon glyphicon-plus"></span></a>
+<a href="{% url 'post_new' %}" class="top-menu">
+    {% include './icons/file-earmark-plus.svg' %}
+</a>
 ```
 
-Note that we want to call our new view `post_new`. The class `"glyphicon glyphicon-plus"` is provided by the bootstrap theme we are using, and will display a plus sign for us.
+Note that we want to call our new view `post_new`. The [SVG icon](https://icons.getbootstrap.com/icons/file-earmark-plus/) is provided by the [Bootstrap Icons](https://icons.getbootstrap.com/) and it will display a page icon with plus sign. We use a django template directive called `include`. This will inject the file's content in to django template since web browser know how to handle this type of content without any further processing.
```
ekohl, line 60:
```suggestion
Note that we want to call our new view `post_new`. The [SVG icon](https://icons.getbootstrap.com/icons/file-earmark-plus/) is provided by the [Bootstrap Icons](https://icons.getbootstrap.com/) and it will display a page icon with plus sign. We use a Django template directive called `include`. This will inject the file's content into the Django template. The web browser knows how to handle this type of content without any further processing.
```

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
...
+
+> You can download the whole bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files in to a new folder inside `blog/templates/blog/` called `icons`. So that you can access a icon, e.g. pencil-fill.svg using the file path `blog/templates/blog/icons/pencil-fill.svg`
```
ekohl, line 62:
```suggestion
> You can download the all Bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files into a new folder inside `blog/templates/blog/` called `icons`. So that you can access a icon, e.g. `pencil-fill.svg` using the file path `blog/templates/blog/icons/pencil-fill.svg`
```

en/django_forms/README.md
```diff
@@ -271,11 +283,17 @@ Django is taking care to validate that all the fields in our form are correct. I
 
 Now we know how to add a new post. But what if we want to edit an existing one? This is very similar to what we just did. Let's create some important things quickly. (If you don't understand something, you should ask your coach or look at the previous chapters, since we covered all these steps already.)
 
+First, let's save the icon which represent edit button. Download [pencil-fill.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/pencil-fill.svg) and save it in location `blog/templates/blog/icons/`
```
ekohl, line 286:
```suggestion
First, let's save the icon which represent edit button. Download [pencil-fill.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/pencil-fill.svg) and save it to the location `blog/templates/blog/icons/`.
```

en/css/README.md
nikhiljohn10: Uniformity. If you do the tutorial, that feel like a anomaly. Please see for that same line in rest of the tutorial.

en/django_forms/README.md
nikhiljohn10: @ekohl 

"all the bootstrap.." is better than "the all bootstrap.." right??

ekohl: Yes, that's what I meant.

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
...
+<a href="{% url 'post_new' %}" class="top-menu">
+    {% include './icons/file-earmark-plus.svg' %}
+</a>
 ```
 
-Note that we want to call our new view `post_new`. The class `"glyphicon glyphicon-plus"` is provided by the bootstrap theme we are using, and will display a plus sign for us.
+Note that we want to call our new view `post_new`. The [SVG icon](https://icons.getbootstrap.com/icons/file-earmark-plus/) is provided by the [Bootstrap Icons](https://icons.getbootstrap.com/) and it will display a page icon with plus sign. We use a Django template directive called `include`. This will inject the file's content into the Django template. The web browser knows how to handle this type of content without any further processing.
+
+> You can download the whole bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files in to a new folder inside `blog/templates/blog/` called `icons`. So that you can access a icon, e.g. pencil-fill.svg using the file path `blog/templates/blog/icons/pencil-fill.svg`
```
nikhiljohn10, line 62:
```suggestion
> You can download all the Bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files into a new folder inside `blog/templates/blog/` called `icons`. So that you can access a icon, e.g. `pencil-fill.svg` using the file path `blog/templates/blog/icons/pencil-fill.svg`
```

nikhiljohn10: @ekohl Check this.

das-g: Is the "So that you can access a icon, e.g. `pencil-fill.svg` using the file path `blog/templates/blog/icons/pencil-fill.svg`" part necessary? Is so, it should either be part of the previous sentence ("so that" introduces a subclause)
```suggestion
> You can download all the Bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files into a new folder inside `blog/templates/blog/` called `icons`, so that you can access an icon like `pencil-fill.svg` using the file path `blog/templates/blog/icons/pencil-fill.svg`
```
or be rephrased to actually be a sentence of its own
```suggestion
> You can download all the Bootstrap icons [here](https://github.com/twbs/icons/releases/download/v1.1.0/bootstrap-icons-1.1.0.zip). Unzip the file and copy all the SVG image files into a new folder inside `blog/templates/blog/` called `icons`. That way you can access an icon like `pencil-fill.svg` using the file path `blog/templates/blog/icons/pencil-fill.svg`
```

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
 
 ## Link to a page with the form
 
-It's time to open `blog/templates/blog/base.html` in the code editor. In the `div` named `page-header`, we will add a link:
+Before we add the link, we need some icons to use as buttons for the link. For this tutorial, download [file-earmark-plus.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/file-earmark-plus.svg) and save it in the folder `blog/templates/blog/icons/`
+
+> Note: To download the SVG image, open the context menu on the link (usually by right-clicking on it) and choose "Save link as" option. Now you can navigate to the desired directory in the window on your screen and save the file.
```
das-g, line 49:
```suggestion
> Note: To download the SVG image, open the context menu on the link (usually by right-clicking on it) and select "Save link as". In the dialog asking you where to save the file, navigate to the `djangogirls` directory of your Django project, and within that to subdirectory `blog/templates/blog/icons/`, and save the file there.
```

en/django_forms/README.md
```diff
@@ -44,36 +44,48 @@ So once again we will create a link to the page, a URL, a view and a template.
...
+
+It's time to open `blog/templates/blog/base.html` in the code editor. Now we can use this icon file inside the base template as follow. In the `div` named `container` inside `page-header` `div` tag, we will add a link:
```
das-g, line 51:
```suggestion
It's time to open `blog/templates/blog/base.html` in the code editor. Now we can use this icon file inside the base template as follows. In the `div` named `container` inside `page-header` `div` tag, we will add a link:
```

Is "In the `div` named `container` inside `page-header` `div` tag" comprehensible to someone who's just begun learning HTML?

en/django_forms/README.md
```diff
@@ -271,11 +283,17 @@ Django is taking care to validate that all the fields in our form are correct. I
 
 Now we know how to add a new post. But what if we want to edit an existing one? This is very similar to what we just did. Let's create some important things quickly. (If you don't understand something, you should ask your coach or look at the previous chapters, since we covered all these steps already.)
 
+First, let's save the icon which represent edit button. Download [pencil-fill.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/pencil-fill.svg) and save it to the location `blog/templates/blog/icons/`.
```
das-g, line 286:
```suggestion
First, let's save the icon which represents the edit button. Download [pencil-fill.svg](https://raw.githubusercontent.com/twbs/icons/main/icons/pencil-fill.svg) and save it to the location `blog/templates/blog/icons/`.
```

das-g (changes requested):

en/django_forms/README.md
nikhiljohn10: @das-g True. I only followed the pattern. Similar pattern of referencing html tags and class through out the tutorial in different parts.

So how can we rewrite the sentence in a simpler terms?

en/css/README.md
```diff
@@ -182,12 +181,14 @@ Great!
...
 
 {% filename %}blog/templates/blog/post_list.html{% endfilename %}
 ```html
 <div class="page-header">
-    <h1><a href="/">Django Girls Blog</a></h1>
+    <div class="container">
+        <h1><a href="/">Django Girls Blog</a></h1>
+    </div>
 </div>
 ```
```
das-g, line 193: Do we need the nested `div` element, or would one div with two classes work, too?
```html
<div class="page-header container">
    <h1><a href="/">Django Girls Blog</a></h1>
</div>
```

That'd make it easier to talk about in the Django Forms chapter.

en/django_forms/README.md
das-g: > So how can we rewrite the sentence in a simpler terms?

We'd probably have to split it up into several sentences, maybe showing also the code before. E.g.

> It's time to open `blog/templates/blog/base.html` in the code editor. Now we can use this icon file inside the base template as follows. At the beginning of the `body` element (so, right after the tag `<body>`) you should have code like this:
> ```html
> <div class="page-header">
>     <div class="container">
>         <h1><a href="/">Django Girls Blog</a></h1>
>     </div>
> </div>
> ```
> Inside the inner `div` element (so, right after `<div class="container">`), we will add a link, just before the `<h1>...` part:

However, the explanation could remain vastly simpler, if there weren't nested `div` elements here. Are they really needed for this migration or would a single `div` element with two classes work, too? (See other comment in this review.)

en/css/README.md
nikhiljohn10: In the css definitions, `page-header` class gives the site wide orange background color for header. `container` class gives the max width property. When we put both together, it looks as follows.

image

nikhiljohn10: @das-g @ekohl I just had an alternative thought. Lets use semantic tag in templates.

https://developer.mozilla.org/en-US/docs/Glossary/Semantics

This is the recommend way of HTML5 after all. This will make our job of referring tags easier.

#### `base.html`
```jinja
{% load static %}
<!DOCTYPE html>
<html>
    <head>
        <title>Django Girls blog</title>
        <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css" integrity="sha384-TX8t27EcRE3e/ihU7zmQxVncDAy5uIKz4rEkgIXeMed4M0jlfIDPvg6uqKI2xXr2" crossorigin="anonymous">
        <link href="//fonts.googleapis.com/css?family=Lobster&subset=latin,latin-ext" rel="stylesheet" type="text/css">
        <link rel="stylesheet" href="{% static 'css/blog.css' %}">
    </head>
    <body>
        <header class="page-header">
            <div class="container">
            {% if user.is_authenticated %}
                <a href="{% url 'post_new' %}" class="top-menu">
                    {% include './icons/file-earmark-plus.svg' %}
                </a>
            {% endif %}
                <h1><a href="/">Django Girls Blog</a></h1>
            </div>
        </header>
        <main class="container">
            <div class="row">
                <div class="col">
                {% block content %}
                {% endblock %}
                </div>
            </div>
        </main>
    </body>
</html>
```

#### `post_detail.html`
```jinja
{% extends 'blog/base.html' %}

{% block content %}
    <article class="post">
    	<aside class="actions">
        {% if post.published_date %}
            <time class="date">
                {{ post.published_date }}
            </time>
        {% endif %}
        {% if user.is_authenticated %}
             <a class="btn btn-default btn-sm" href="{% url 'post_edit' pk=post.pk %}">
                {% include './icons/pencil-fill.svg' %}
             </a>
        {% endif %}
		</aside>
        <h2>{{ post.title }}</h2>
        <p>{{ post.text|linebreaksbr }}</p>
    </article>
{% endblock %}
```

#### `post_list.html`
```jinja
{% extends 'blog/base.html' %}

{% block content %}
    {% for post in posts %}
        <article class="post">
            <time class="date">
                {{ post.published_date }}
            </time>
            <h2><a href="{% url 'post_detail' pk=post.pk %}">{{ post.title }}</a></h2>
            <p>{{ post.text|linebreaksbr|truncatechars:77 }}</p>
        </article>
    {% endfor %}
{% endblock %}
 ```

#### `blog.css`
```css
.page-header {
    background-color: #C25100;
    margin-top: 0;
    margin-bottom: 40px;
    padding: 20px 20px 20px 40px;
}

.page-header h1,
.page-header h1 a,
.page-header h1 a:visited,
.page-header h1 a:active {
    color: #ffffff;
    font-size: 36pt;
    text-decoration: none;
}

h1,
h2,
h3,
h4 {
    font-family: 'Lobster', cursive;
}

.date {
    color: #828282;
}

.save {
    float: right;
}

.post-form textarea,
.post-form input {
    width: 100%;
}

.top-menu,
.top-menu:hover,
.top-menu:visited {
    color: #ffffff;
    float: right;
    font-size: 26pt;
    margin-right: 20px;
}

.post {
    margin-bottom: 70px;
}

.post h2 a,
.post h2 a:visited {
    color: #000000;
}

.post > .date,
.post > .actions {
    float: right;
}


.btn-default,
.btn-default:visited {
    color: #C25100;
    background: none;
    border-color: #C25100;
}

.btn-default:hover {
    color: #FFFFFF;
    background-color: #C25100;
}
```

das-g: Semantic elements indeed sound like a good idea. If I understand correctly, the simplification would be due to not fewer elements in total, but fewer of the same type at the same place, therefore making it easier to convey what element or tag we're referring to.

nikhiljohn10: @das-g Exactly. Same code with unqiuely identifiable tags. 

Btw, I am not sure if `<aside>` tag is used in correct context there. Never used that tag in my life. xD Maybe we should simple use `div` tag instead of `aside` tag. Same goes to `time` tag. Never used in real senario before. But I think it probably helps search engine to show new contents from the site on searchs based on these tags.

I have some research on new SEO techniques and found that following semantics tags and HTML5 structure is huge jump up the search ranking.

das-g: Looking at [the HTML5 spec for `aside`](https://www.w3.org/TR/html52/sections.html#the-aside-element) (especially [example 15](example-650ac100)) and [MDN's documention for `time`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time), I think these are semantically correct uses of these two elements. :+1:

ekohl: I like that solution.

nikhiljohn10: @ekohl @das-g Please review https://github.com/DjangoGirls/tutorial/pull/1681/commits/a9dbbf582e5e863bbb4ef50a16201a2b758a81f9

nikhiljohn10:
@das-g Is there anything more needed in this PR to merge?

das-g (approved):
> Is there anything more needed in this PR to merge?

I must confess I lost track, but I think it's fine now. What do you say @ekohl?

nikhiljohn10:
> > Is there anything more needed in this PR to merge?
> 
> I must confess I lost track, but I think it's fine now. What do you say @ekohl?

Thanks @das-g. Hi @ekohl, does this PR need any more changes?

nikhiljohn10:
I guess this PR might be falling in to lost and found.

das-g:
Eh, I'll merge this as-is.

PR #1682, opened 2020-11-12
Updating git inside tutorial

nikhiljohn10:
### Git updates

- [x] Removed --all option from git add 
- [x] Updated .gitignore

This is a followup from #1680


das-g (approved):
As indicated in https://github.com/DjangoGirls/tutorial/pull/1680#discussion_r522108270, the new behavior of `git add $SOME_PATH` is there since Git 2.0.0 in 2014, so we can assume that it is by now available to all programming newcomers.

ekohl (approved):
Thanks!

PR #1683, opened 2020-11-13
Python 3 Upgrade

nikhiljohn10:
### Draft for Python 3 upgrade

- [x] Added python versions to book.json
- [x] Converted python versions to book variable

If this draft is acceptable, i'll create a PR using these commits.

@das-g @ekohl Please review. Thanks


nikhiljohn10:
I got some questions: 

### 1.
- [ ] Do I have to create screenshots for the new upgrade within this PR regarding Python? 
-or-
- [ ] Should I create a new PR will new screenshots of every screenshots found inside tutorial?

### 2.

- [ ] Manually hard code version of python in tutorial?
-or-
- [ ] Automate Python versions using book.json and variable replacement?

### 3.

- [ ] Does python intro need any changes according to Python 3.8 version?
-or-
- [ ] Should I not touch the python Intro tutorial at all?

das-g:
> ### 1.
> - [ ] Do I have to create screenshots for the new upgrade within this PR regarding Python? 
> -or-
> - [ ] Should I create a new PR will new screenshots of every screenshots found inside tutorial?

Whichever you prefer. Third option:
- File an issue about updating the screenshots, so someone else can do it.

That's the idea behind preparing the updates that depend on each other on a separate branch: We don't have to cover everything in a single PR, and more than one person can work on different aspects.

das-g:
> ### 2.
> 
> - [ ] Manually hard code version of python in tutorial?
> -or-
> - [ ] Automate Python versions using book.json and variable replacement?

Automating it with the variable might be a good idea, especially to keep the lesser-active translation more in sync. Though, it can result in these translations getting internally inconsistent, if stuff that depends on the version but hasn't been automated doesn't get re-translated.

Would 4863b610f306bb91b2495a6fe2f4872ebaf06ad3 already cover all occurrences of the version number?

das-g:
> ### 3.
> 
> - [ ] Does python intro need any changes according to Python 3.8 version?
> -or-
> - [ ] Should I not touch the python Intro tutorial at all?

As far as I know, Python 3.8 and 3.9 are fully backwards-compatible to the Python version(s) we currently use, i.e., everything that will work in Python 3.6 should work in 3.8 and 3.9, too. So there is no need to update Python intro. (We should probably test it with the new Python version(s), but I don't expect to find any differences in behavior except _maybe_ for slightly different output formats here and there.)

While I very much like the newly introduced features of the newer Python versions, I don't think any of them are relevant enough for newcomers that they'd have to be covered in the Intro. If we ever come to a different conclusion, we can still make that change then, in a new PR.

nikhiljohn10:
> Would [4863b61](https://github.com/DjangoGirls/tutorial/commit/4863b610f306bb91b2495a6fe2f4872ebaf06ad3) already cover all occurrences of the Python version number(s)?

@das-g  Yes, it covers 100% of version texts inside the tutorial and converts to the GitBook variable.

nikhiljohn10:
On file `django_installation/instructions.md`, Line 176:
```
(myvenv) ~$ python -m pip install --upgrade pip
```

Can simply use `pip install -U pip` instead? I have tested and both commands result in upgrading `pip` inside `myvenv` directory.

das-g:
> On file `django_installation/instructions.md`, Line 176:
> 
> ```
> (myvenv) ~$ python -m pip install --upgrade pip
> ```
> 
> Can simply use `pip install -U pip` instead? I have tested and both commands result in upgrading `pip` inside `myvenv` directory.

I'm not sure. The `python3 -m` part was added in #1274 (and changed to the `python -m` variant without version number in #1450). @anapaulagomes, @ekohl can you remember what the problem was with plain `pip install --upgrade pip` instead of `python -m pip install --upgrade pip`?

IIRC the `python -m pip install --upgrade pip` variant proved to work on more different system (various Linux distributions, probably) than plain `pip install --upgrade pip`. Why that might have been, I'm not quite sure, as within a Python 3 virtualenv, the `pip` command should always be available, whether pip installed system-wide or not. (Some Linux distributions package pip separately from Python, thus it's possible that system-wide, Python 3 is installed, but the pip for it isn't. But in a virtualenv, that shouldn't be an issue.)

das-g:
> On file `django_installation/instructions.md`, Line 176:
> 
> ```
> (myvenv) ~$ python -m pip install --upgrade pip
> ```
> 
> Can simply use `pip install -U pip` instead? I have tested and both commands result in upgrading `pip` inside `myvenv` directory.

As for `-U` vs. `--upgrade`: Although it's more to type, I'd stick with the long option, as it's more obvious what it might do than with the short one, for which a new user would probably have to consult the documentation.

nikhiljohn10:
> As for `-U` vs. `--upgrade`: Although it's more to type, I'd stick with the long option, as it's more obvious what it might do than with the short one, for which a new user would probably have to consult the documentation.

I don't find that as a valid statement since there are other short forms of options used inside tutorial. For eg, -r, --requirement in pip install, -u, --set-upstream in git push etc. When I started learning python, i have also google all these to understand what they really do.

We can add a 'NOTE' under pip install -U p about its -U option is --upgrade shortform.

How about that?

nikhiljohn10:
> > On file `django_installation/instructions.md`, Line 176:
> 
> > 
> 
> > ```
> 
> > (myvenv) ~$ python -m pip install --upgrade pip
> 
> > ```
> 
> > 
> 
> > Can simply use `pip install -U pip` instead? I have tested and both commands result in upgrading `pip` inside `myvenv` directory.
> 
> 
> 
> I'm not sure. The `python3 -m` part was added in #1274 (and changed to the `python -m` variant without version number in #1450). @anapaulagomes, @ekohl can you remember what the problem was with plain `pip install --upgrade pip` instead of `python -m pip install --upgrade pip`?
> 
> 
> 
> IIRC the `python -m pip install --upgrade pip` variant proved to work on more different system (various Linux distributions, probably) than plain `pip install --upgrade pip`. Why that might have been, I'm not quite sure, as within a Python 3 virtualenv, the `pip` command should always be available, whether pip installed system-wide or not. (Some Linux distributions package pip separately from Python, thus it's possible that system-wide, Python 3 is installed, but the pip for it isn't. But in a virtualenv, that shouldn't be an issue.)

I am sure this is only for Windows OS. So lets try the approach we used in this link. https://tutorial.djangogirls.org/en/django_models/#creating-an-application

Here both options are given in separate command line sections.

For eg:

(Linux and MacOS Users)
```
(myvenv) ~ $ pip install -U pip
```

(Windows Users)
```
(myvenv) C:\> python -m pip install -U pip
```

@das-g let me know your thoughts on above method.

anapaulagomes:
Sorry, I don't remember. 😐

das-g:
> > Would [4863b61](https://github.com/DjangoGirls/tutorial/commit/4863b610f306bb91b2495a6fe2f4872ebaf06ad3) already cover all occurrences of the Python version number(s)?
> 
> @das-g Yes, it covers 100% of version texts inside the tutorial and converts to the GitBook variable.

:+1:


> Does that mean to commit in this draft PR is ok approval?

I'm not sure I understand your question. Can you rephrase it? (If English isn't your native language, maybe write it in your native language.)


> If yes, I don't need a new PR. I can change the draft status of this PR for review and merge. There is nothing else to update regarding python versions.

Yes, converting a draft PR to a normal, merge-able PR is completely OK. You don't have to file a new PR for the same thing.

das-g:
> We can add a 'NOTE' under pip install -U p about its -U option is --upgrade shortform.

Yeah, with such an explanation (be it in a note admonition or just in the prose) it should be fine. Feel free to open a separate PR for that. (Separate, as it's independent from the python version change.)

das-g:
About `python -m pip ...` vs. just `pip ...`:

I don't think it was a Windows specific issue. On Windows, too, a virtualenv created by `python -m venv` should include `pip`. And when the virtualenv is activated, the `pip` from the virutalenv should be available as simply `pip`.

After some research, I guess the real reason is, that Python 3's `venv` module didn't always install `pip`:
- [in Python 3.4](https://docs.python.org/3.4/library/venv.html) and greater it does (unless used with the `--without-pip` option)
- [in Python 3.3](https://docs.python.org/3.3/library/venv.html) and lower it didn't

(Though I'm not quite sure how the `pip` module (which is used by `-m pip`) could have been available then, as it isn't a [standard library](https://docs.python.org/3/library) module, and wasn't one [in Python 3.3](https://docs.python.org/3.3/library), either.)

Since Django 3.1 requires Python &ge; 3.6 we can probably drop the workaround for all operating systems. (Though, we should test on all of them, first.)

book.json
```diff
@@ -1,6 +1,9 @@
 {
     "gitbook": ">=3.2.0",
     "variables": {
+      "py_version": "3.9",
+      "py_release": "3.9.0",
+      "pyaw_version": "3.8",
```
das-g, line 6: As this isn't the version of PythonAnywhere, but of Python **on** PythonAnywhere, maybe abbreviate it as `pa_py_version` or `py_version_pa`. Or spell "PythonAnywhere" out completely: `py_version_on_pythonanywhere`

nikhiljohn10:
> About `python -m pip ...` vs. just `pip ...`:
> 
> 
> 
> I don't think it was a Windows specific issue. On Windows, too, a virtualenv created by `python -m venv` should include `pip`. And when the virtualenv is activated, the `pip` from the virutalenv should be available as simply `pip`.
> 
> 
> 
> After some research, I guess the real reason is, that Python 3's `venv` module didn't always install `pip`:
> 
> - [in Python 3.4](https://docs.python.org/3.4/library/venv.html) and greater it does (unless used with the `--without-pip` option)
> 
> - [in Python 3.3](https://docs.python.org/3.3/library/venv.html) and lower it didn't
> 
> 
> 
> (Though I'm not quite sure how the `pip` module (which is used by `-m pip`) could have been available then, as it isn't a [standard library](https://docs.python.org/3/library) module, and wasn't one [in Python 3.3](https://docs.python.org/3.3/library), either.)
> 
> 
> 
> Since Django 3.1 requires Python &ge; 3.6 we can probably drop the workaround for all operating systems. (Though, we should test on all of them, first.)

Hope this helps

[Packaging status](https://repology.org/project/python/versions)

en/python_installation/instructions.md
```diff
@@ -16,13 +16,13 @@ First check whether your computer is running a 32-bit version or a 64-bit versio
...
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
 When the installation completes, you may see a dialog box with a link you can follow to learn more about Python or about the version you installed. Close or cancel that dialog -- you'll be learning more in this tutorial!
 
-Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python 3.6.x installer fails with an error, you can try either:
+Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python {{ book.py_version }}.x installer fails with an error, you can try either:
 1. install all Windows Updates and try to install Python again; or
 2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [3.4.6](https://www.python.org/downloads/release/python-346/).
```
das-g, line 27: [Django 3.1 requires Python 3.6 or greater](https://docs.djangoproject.com/en/3.1/faq/install/#faq-python-version-support), so we shouldn't recommend Python 3.4.6 anymore, nor use it as an example.

Does the issue of installers for newer Python versions failing for older Windows versions still occur at all? If so, do we still have to assume some participants are using these Windows versions? (Do they still get security updates?) And does trying with an older version of Python have any chance of success, if that version would have to be 3.6 or greater, and the previous version of this instruction assumed the issue would occur with the Python 3.6 installer?

en/python_installation/instructions.md
```diff
@@ -52,10 +52,10 @@ It is very likely that you already have Python installed out of the box. To chec
 {% filename %}command-line{% endfilename %}
 ```
 $ python3 --version
-Python 3.6.1
+Python {{ book.py_release }}
 ```
 
-If you have a different version of Python installed, at least 3.4.0 (e.g. 3.6.0), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:
+If you have a different version of Python installed, at least 3.4.0 (e.g. {{ book.py_release }}), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:
```
das-g, line 58: [Django 3.1 requires Python 3.6 or greater](https://docs.djangoproject.com/en/3.1/faq/install/#faq-python-version-support).
```suggestion
If you have a different version of Python installed, at least 3.6.0 (e.g. {{ book.py_release }}), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:
```

(Maybe we should introduce a variable for the minimum Python version, too?)

en/python_installation/instructions.md
```diff
@@ -108,11 +108,11 @@ Verify the installation was successful by opening a command prompt and running t
...
 ```
 $ python3 --version
-Python 3.6.1
+Python {{ book.py_release }}
 ```
-The version shown may be different from 3.6.1 -- it should match the version you installed.
+The version shown may be different from {{ book.py_release }} -- it should match the version you installed.
 
-**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) and check if it still might be a version of Python that is 3.4.0 or higher. If that doesn't work either, you may open a new command prompt and try again; this happens if you use a command prompt left open from before the Python installation. 
+**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) and check if it still might be a version of Python that is 3.4.0 or higher. If that doesn't work either, you may open a new command prompt and try again; this happens if you use a command prompt left open from before the Python installation.
```
das-g, line 115: [Django 3.1 requires Python 3.6 or greater](https://docs.djangoproject.com/en/3.1/faq/install/#faq-python-version-support).
```suggestion
**NOTE:** If you're on Windows and you get an error message that `python3` wasn't found, try using `python` (without the `3`) and check if it still might be a version of Python that is 3.6.0 or higher. If that doesn't work either, you may open a new command prompt and try again; this happens if you use a command prompt left open from before the Python installation.
```

(Maybe we should introduce a variable for the minimum Python version, too?)

das-g (changes requested):
:+1: Looks good, except for some details:

book.json
nikhiljohn10: @das-g pushed commit: https://github.com/DjangoGirls/tutorial/pull/1683/commits/bebcedb20dad37022c6ce0adcf41394df71035c9

en/python_installation/instructions.md
nikhiljohn10: @das-g Found current compatibility versions for Windows

[Python 3.4.10](https://www.python.org/downloads/release/python-3410/) is the last python version that can be used in Windows XP or earlier as per Python.org. It is also reached its end of development. 

[Python 3.8.6](https://www.python.org/downloads/release/python-386/) is the last python version that can be used in Windows 7 or earlier as per Python.org.

If we can assume that no one is using Windows XP or earlier to learn Django using Djangogirls, then we can recommend, Python 3.8.6 as the latest version (Python 3.9.0 which came afterwards have dropped support for Window 7 or earlier versions)

So using version 3.8.6 is a better choice for Django girls in general. But Pythonanywhere is using Python 3.8.0. I don't think there will be much issues here with different patch versions. Version 3.8 in general will be safe choice I think.

In that case, we don't have to worry about backward compatibility of Python with Windows 7 or Vista.

en/python_installation/instructions.md
nikhiljohn10: That's a great idea. How does the `{{ book.py_min_version }}` sounds?

das-g:
Sounds good to me.

On November 14, 2020 3:57:23 PM GMT+01:00, Nikhil John <notifications@github.com> wrote:
>@nikhiljohn10 commented on this pull request.
>
>
>
>>  ```
> 
>-If you have a different version of Python installed, at least 3.4.0
>(e.g. 3.6.0), then you don't have to upgrade. If you don't have Python
>installed, or if you want a different version, first check what Linux
>distribution you are using with the following command:
>+If you have a different version of Python installed, at least 3.4.0
>(e.g. {{ book.py_release }}), then you don't have to upgrade. If you
>don't have Python installed, or if you want a different version, first
>check what Linux distribution you are using with the following command:
>
>That's a great idea. How does the `{{ book.py_min_version }}` sounds?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/DjangoGirls/tutorial/pull/1683#discussion_r523427963

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

en/python_installation/instructions.md
nikhiljohn10: @das-g Check commit https://github.com/DjangoGirls/tutorial/pull/1683/commits/d90b4a38a4531b28701c866cff611f1cc7c8161e

en/python_installation/instructions.md
nikhiljohn10: @das-g Check commit https://github.com/DjangoGirls/tutorial/commit/d90b4a38a4531b28701c866cff611f1cc7c8161e

en/python_installation/instructions.md
nikhiljohn10: @das-g Check commit https://github.com/DjangoGirls/tutorial/commit/d90b4a38a4531b28701c866cff611f1cc7c8161e

Can I have confirmation about the following change on python version?

```
{
    "gitbook": ">=3.2.0",
    "variables": {
      "py_version": "3.8", # Previously 3.9
      "py_release": "3.8.6", # Previously 3.9.0
      "py_min_version": "3.6.0",
      "pa_py_version": "3.8",
      "django_version": "2.2.4"
    },

...
```

das-g: > Can I have confirmation about the following change on python version?
> 
> ```
>       ...
>       "py_version": "3.8", # Previously 3.9
>       "py_release": "3.8.6", # Previously 3.9.0
>       ...
> ```

:+1: fine with me

en/python_installation/instructions.md
```diff
@@ -40,7 +40,7 @@ data-collapse=true ces-->
 You need to go to the website https://www.python.org/downloads/release/python-361/ and download the Python installer:
```
das-g, line 40: This URL is for Python 3.6.1 and should probably be changed to conform with `book.py_version`.

en/python_installation/instructions.md
```diff
@@ -16,15 +16,15 @@ First check whether your computer is running a 32-bit version or a 64-bit versio
...
 
 ![Don't forget to add Python to the Path](../python_installation/images/python-installation-options.png)
 
 When the installation completes, you may see a dialog box with a link you can follow to learn more about Python or about the version you installed. Close or cancel that dialog -- you'll be learning more in this tutorial!
 
-Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python 3.6.x installer fails with an error, you can try either:
+Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python {{ book.py_version }}.x installer fails with an error, you can try either:
 1. install all Windows Updates and try to install Python again; or
-2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [3.4.6](https://www.python.org/downloads/release/python-346/).
+2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [{{ book.py_min_version }}](https://www.python.org/downloads/release/python-346/).
```
das-g, line 27: The second URL is for Python 3.4.6 and should probably be changed to conform with `book.py_min_version`.

das-g (changes requested):
Some URLs point to specific Python versions other than they should refer now.

Should we have variables for them, too, so that they can be changed at the same place as the versions themselves?

en/python_installation/instructions.md
nikhiljohn10: Solution 1: Remove URL and keep only version text

Solution 2: Give static URL updated with each minimum version changes and set it to 3.6.0 for now and set URL

Solution 3. Set all minimum version as static text (3.6.0) as well as URL

@das-g  What do you think we can go forward with?

nikhiljohn10: @das-g Found a mistake. Django 3 only supports 3.6. So we can not provide backward compatibility with Python 3.4. So why do we need this section if we are using the 3.8.6 version which is providing everything it can as the primary version we are using?

```
Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python {{ book.py_version }} installer fails with an error, you can try either:
1. install all Windows Updates and try to install Python again; or
2. install an [older version of Python](https://www.python.org/downloads/windows/), e.g., [3.4.10](https://www.python.org/downloads/release/python-3410/).

If you install an older version of Python, the installation screen may look a bit different than shown above. Make sure you scroll down to see "Add python.exe to Path", then click the button on the left and pick "Will be installed on local hard drive":

![Add Python to the Path, older versions](../python_installation/images/add_python_to_windows_path.png)
```

So the above part is no more needed at all. Maybe we can try the following:

```
Note: if you are using an older version of Windows (7, Vista, or any older version) and the Python {{ book.py_version }} installer fails with an error, then install all Windows Updates and try to install Python again
```

en/python_installation/instructions.md
nikhiljohn10: This one can be changed as follows:

```

You need to go to the website https://www.python.org/downloads/mac-osx/ and download the latest Python installer:

* Download the *Mac OS X 64-bit/32-bit installer* file,
* Double click *python-{{ book.py_release }}-macosx10.9.pkg* to run the installer.

```

So it becomes independent of the version. Users can download whichever version is latest and it won't affect our code.

en/python_installation/instructions.md
das-g: Yeah, let's modify it that way, i.e., remove option 2. See also https://github.com/DjangoGirls/tutorial/pull/1683#discussion_r523408737

en/python_installation/instructions.md
das-g: :+1: Good idea.

en/python_installation/instructions.md
nikhiljohn10: > Does the issue of installers for newer Python versions failing for older Windows versions still occur at all? 

With version 3.8.6, it still supports  Windows 7 and Vista as per Python.org

> If so, do we still have to assume some participants are using these Windows versions? (Do they still get security updates?)

We have to assume that our participants are not using the Windows version released before 7 or Vista versions. Version 3.8 still has its security updates comming until October of 2024 as per [Python.org](https://www.python.org/dev/peps/pep-0569/#id4) and [DevGuide](https://devguide.python.org/#status-of-python-branches)

>And does trying with an older version of Python have any chance of success, if that version would have to be 3.6 or greater, and the previous version of this instruction assumed the issue would occur with the Python 3.6 installer?

From the information I gathered, I came to know that only version 3.4 have further backward compatibilty. Versions 3.6 and 3.8 should have same result as per Python.org. Since we can't support 3.4 with Django 3.1 in our hands, there is a last resort kind of solution. 

Create a copy of the current version of Django (2.2.4) and Publish a new subdomain like "windows.djangogirls.org". So that this url will act as an archive and help users who have older versions of Windows.

@das-g Check the above suggestion.

en/chromebook_setup/instructions.md
```diff
@@ -74,7 +74,7 @@ Run:
 ```
 mkdir djangogirls
 cd djangogirls
-python3.6 -mvenv myvenv
+python{{ book.py_version }} -mvenv myvenv
```
ekohl, line 77: Is 3.9 really available on Cloud9? https://docs.aws.amazon.com/cloud9/latest/user-guide/sample-python.html suggests it's only 3.6.

das-g: > Is 3.9 really available on Cloud9?

Good point! Seems like it isn't. Thanks for catching that.

> https://docs.aws.amazon.com/cloud9/latest/user-guide/sample-python.html suggests it's only 3.6.

I've just tried, choosing their "Amazon Linux 2 (recommended)" platform. It comes with Python 3.**7**.9, and it seems like it offers no greater Python versions (at least not through `yum`):
```
$ yum list python3
Loaded plugins: extras_suggestions, langpacks, priorities, update-motd
219 packages excluded due to repository priority protections
Installed Packages
python3.x86_64              3.7.9-1.amzn2.0.1              @amzn2-core
Available Packages
python3.i686                3.7.9-1.amzn2.0.1              amzn2-core 
```
(If that's the right way to ask `yum`, that is. I don't really know that package manager.)

> in various places we just use python3 without the minor version. Perhaps that's also better to use here?

Probably, as it doesn't seem like there are typically several Python 3 versions in the same AWS environment. So that'd be
```suggestion
python3 -mvenv myvenv
```

das-g: (I've now also tried with the (apparently not anymore recommended?) plattform "Amazon Linux" (without "2"). It comes indeed with Python 3.**6**.12, with no obvious possibility to update to a greater version.)

en/python_installation/instructions.md
das-g: > We have to assume that our participants are not using the Windows version released before 7 or Vista versions.

Yeah. My question is: Is that a suitable assumption?

> Version 3.8 still has its security updates comming until October of 2024 as per Python.org and DevGuide

Sorry, what I meant was: Do _these Windows versions_ still receive Windows security updates from Microsoft?

Any Windows version that doesn't receive security updates anymore can be considered unsupported by Microsoft, so we probably don't have to support it either. (As those still using it should update to a newer version or change to some other operating system anyway, lest their system is left in undue risk.)

> Since we can't support 3.4 with Django 3.1 in our hands, there is a last resort kind of solution.
> 
> Create a copy of the current version of Django (2.2.4) and Publish a new subdomain like "windows.djangogirls.org". So that this url will act as an archive and help users who have older versions of Windows.

I'd rather not go that way, as we struggle enough to keep one version of the tutorial maintained.

nikhiljohn10: > Yeah. My question is: Is that a suitable assumption?

If you are a person who is working on Windows XP, 100% of the things can go wrong badly while installing. Every in windows 7, new softwares are very hard to install without Windows updateds.

> Sorry, what I meant was: Do _these Windows versions_ still receive Windows security updates from Microsoft?

https://en.wikipedia.org/wiki/List_of_Microsoft_Windows_versions

As per the above link, Microsoft ended support for all versions before Windows 8.1 on the consumer side. Which means Windows 8, 7, Vista, XP, etc. So currently Microsoft is only providing updates for Windows 8.1 and Windows 10.

> Any Windows version that doesn't receive security updates anymore can be considered unsupported by Microsoft, so we probably don't have to support it either. (As those still using it should update to a newer version or change to some other operating system anyway, lest their system is left at undue risk.)

As per Python.org, 3.6.0 does not support any windows version of Windows XP or earlier. Django 3+ need Python 3.6.0. So we can reach a logical conclusion that users need to follow minimal requirement for using this tutorial as Windows 7 or later since we can support it even without Microsoft support. But If you want only support Windows versions with MS updates, then we have to stick to Windows 8.1 and Windows 10 as mentioned above.

Fun fact: https://stackoverflow.com/questions/53821523/how-to-deploy-a-django-app-locally-on-a-windows-xp-machine

en/chromebook_setup/instructions.md
nikhiljohn10: You can find all the current python versions in this comment: https://github.com/DjangoGirls/tutorial/pull/1683#issuecomment-727190424

I assume this is resolved now.

en/python_installation/instructions.md
nikhiljohn10: Btw, **Python 3.6.12** is the lastest under Python 3.6. Our current version set to 3.6.0. Should we change it to 3.6.12 which has all the new patchs?

en/python_installation/instructions.md
nikhiljohn10: @das-g Here is an updated version of same.

```
Note: If you are using an older version of Windows (7, Vista, or any older version) and the Python {{ book.py_version }} installer fails with an error, then install all Windows Updates and try to install Python again. If you still have the error, try installing Python version {{ book.py_min_version }} from [Python.org](https://www.python.org/downloads/windows/).

> Django {{ book.django_version }} need {{ book.py_min_version }} or greater which does not support Windows XP or earlier versions. If you are using Windows XP or earlier, please upgrade to your Windows OS.
```

en/python_installation/instructions.md
nikhiljohn10: I just noticed that Python version 3.6.12 does not have any downloadable files for windows. Only version [3.6.8](https://www.python.org/downloads/release/python-368/) has all the downloadable files for Windows as well as macOS.

nikhiljohn10:
I have generated a stagging site under my personal domain. This contains all the above commits.

#### https://django.nikz.in/en/

en/python_installation/instructions.md
das-g: I can see

>     > Django {{ book.django_version }} need {{ book.py_min_version }} or greater which does not support Windows XP or earlier versions. If you are using Windows XP or earlier, please upgrade to your Windows OS.

on https://django.nikz.in/en/python_installation/#python-installation, but not in the diff of the PR here. Did you maybe make that addition locally, without pushing it to the branch on your GitHub repo?

Do we need that note at all? If so, I think it should be

>     > Django {{ book.django_version }} needs Python {{ book.py_min_version }} or greater, which does not support Windows XP or earlier versions.

Changes:
- "need**s**" instead of "need"
- "**Python** {{ book.py_min_version }}" instead of just "{{ book.py_min_version }}"
- leave out the second sentence. How to resolve the issue is IMO up to the participant or whoever maintains the participant's computer. Some might want to buy a newer Windows, others might decide to switch to Linux.

nikhiljohn10: Yes i didn't pushed it to PR since I am not sure. I changed it in local master only for deploy. I'll push it with your changes mentioned to this PR. Looks good.

nikhiljohn10: Resolved with commits https://github.com/DjangoGirls/tutorial/pull/1683/commits/0aab1eb0df8e2b31486d4dd2f21adab47be5d3d7 and https://github.com/DjangoGirls/tutorial/pull/1683/commits/f99a47c1f1960df4cf6361627ac9a12f3bee7171

en/python_installation/instructions.md
nikhiljohn10: Resolved with commit https://github.com/DjangoGirls/tutorial/pull/1683/commits/e448e25ca043033b7ee91b14482b1aab2b245879

en/python_installation/instructions.md
nikhiljohn10: Resolved with commit https://github.com/DjangoGirls/tutorial/pull/1683/commits/3eacfdef9efbb4d18b8db3c2762a7c9cf18e0088

das-g: https://www.python.org/downloads/release/python-3612/ mentions
>  Python 3.6.8 was the last bugfix release for 3.6. **Python 3.6 is now in the security fix phase** of its life cycle. Only security-related issues are accepted and addressed during this phase. We plan to provide security fixes for Python 3.6 as needed through 2021, five years following its initial release. **Security fix releases** are produced periodically as needed and **are source-only releases; binary installers are not provided.**

(emphasis mine)

I guess security fix releases are meant mostly for old but still supported releases of Linux distributions that package an old minor version of Python, and therefore don't require upstream binaries. (As Linux distributions tend to compile their own binaries from the upstream source code.) The idea is probably that Windows users can simply get a later minor version instead, which seems true for all still-maintained versions of Windows.

en/python_installation/instructions.md
```diff
@@ -52,10 +48,10 @@ It is very likely that you already have Python installed out of the box. To chec
 {% filename %}command-line{% endfilename %}
 ```
 $ python3 --version
-Python 3.6.1
+Python {{ book.py_release }}
 ```
 
-If you have a different version of Python installed, at least 3.4.0 (e.g. 3.6.0), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:
+If you have a different version of Python installed, at least {{ book.py_min_version }} (e.g. {{ book.py_release }}), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:
```
das-g, line 54: `{{ book.py_release }}` can't be an example of "**different** version of Python [than `{{ book.py_release }}`]".

I think it should be
>  If you have a different version of Python installed, at least 3.6.0 (e.g. 3.6.8), then you don't have to upgrade. If you don't have Python installed, or if you want a different version, first check what Linux distribution you are using with the following command:

Maybe we should set `book.py_min_version` back to `3.6.0` and introduce another variable for the fully patched release of the same minor version. That'd also make the other statements where `{{ book.py_release }}` is now used more true, as the requirement really _is_ 3.6._0_ or higher, and not the latest security release.

das-g (changes requested):
I noticed another thing:

en/python_installation/instructions.md
nikhiljohn10: How about this:

```
"py_min_version": 3.6,
"py_min_release": 3.6.8,
```

We can use it as follow:

... at least {{ book.py_min_version }} (e.g. {{ book.py_min_release }})

das-g: Yeah, that sounds good.

nikhiljohn10: Added https://github.com/DjangoGirls/tutorial/pull/1683/commits/1958088cceaa83483f4eae96a8a04ff60e519b79

das-g (approved):
Looks good to me now.

ekohl (approved):
Looks good and nice that it'll be easier in the future.

ekohl:
@das-g shall we merge it or is there anything else to do?

das-g:
> @das-g shall we merge it or is there anything else to do?

I'm not aware of anything more that would have to be done for changing the Python version(s).

According to its [release notes](https://docs.djangoproject.com/en/3.1/releases/2.2/#python-compatibility), the currently used Django version (2.2) _does_ support both Python versions the tutorial will mention with this PR: Python 3.6 & Python 3.8. Thus it should be fine to merge this right into `master` instead of (or maybe better, additionally to) merging it into `django-3.1`.

I didn't yet have time to test the complete tutorial with any of the new Python versions, and probably won't have time for that in the near future. So if anyone did test it or will test it, please comment here. Nonetheless, I'm confident that this shouldn't introduce any major problems.

Shall I go ahead an merge it, @ekohl ? (Or, if you agree, feel free to merge it yourself.)

ekohl:
> I didn't yet have time to test the complete tutorial with any of the new Python versions, and probably won't have time for that in the near future. So if anyone did test it or will test it, please comment here. Nonetheless, I'm confident that this shouldn't introduce any major problems.

I'm in a similar position so I was also a bit hesitant to merge. However, like you I'm also confident it won't introduce major problems. Let's merge it.

Thanks @nikhiljohn10!

PR #1686, opened 2020-11-14
write "installation steps" in lowercase

das-g:
Changes in this pull request:

- write "installation steps" in lowercase (But keep "Installation chapter" with an upper case "I", as it refers to a
title.)


ekohl (approved):
Does it make sense to convert `installation steps` to a link to make it clear what we're talking about?

das-g:
> Does it make sense to convert `installation steps` to a link to make it clear what we're talking about?

Yeah, probably. I'll do that in a separate PR.

PR #1688, opened 2020-11-15
link to installation chapter

das-g:
Link to installation chapter where we refer to it

(Follow-up to https://github.com/DjangoGirls/tutorial/pull/1686#pullrequestreview-530807696)

### DIFF
--- a/en/code_editor/README.md
+++ b/en/code_editor/README.md
@@ -6,6 +6,6 @@ You're about to write your first line of code, so it's time to download a code e
 
 > **Note** If you're using a Chromebook, skip this chapter and make sure you follow the [Chromebook Setup](../chromebook_setup/README.md) instructions. The cloud IDE you chose (PaizaCloud Cloud IDE or AWS Cloud9) includes a code editor, and when you open a file in your IDE from the File menu, you will automatically be using the editor.
 
-> **Note** You might have done this earlier in the Installation chapter – if so, you can skip right ahead to the next chapter!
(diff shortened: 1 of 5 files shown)


magul (approved):

en/code_editor/README.md
```diff
@@ -6,6 +6,6 @@ You're about to write your first line of code, so it's time to download a code e
 
 > **Note** If you're using a Chromebook, skip this chapter and make sure you follow the [Chromebook Setup](../chromebook_setup/README.md) instructions. The cloud IDE you chose (PaizaCloud Cloud IDE or AWS Cloud9) includes a code editor, and when you open a file in your IDE from the File menu, you will automatically be using the editor.
 
-> **Note** You might have done this earlier in the Installation chapter – if so, you can skip right ahead to the next chapter!
+> **Note** You might have done this earlier in the [Installation chapter](../installation/README.md) – if so, you can skip right ahead to the next chapter!
```
ekohl, line 9: Would it be consistent to also write this as lower case? Also interesting that this one mentions chapter where the others use steps.

das-g: In #1686 I left this occurrence alone, as I assumed that it refers to the chapter title, which is in title case.

Shall we change it to the same wording ("installation steps") as in the other notes?

ekohl: That would be more consistent so I think I'd prefer that.

PR #1690, opened 2020-11-17
Pip upgrade command update for django installation tutorial

nikhiljohn10:
### pip command update

- [x] Updated pip upgrade command
- [x] Added note on pip upgrade option

Ref: https://github.com/DjangoGirls/tutorial/pull/1683#issuecomment-727185235 by @das-g

### DIFF
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -181,6 +181,8 @@ Before we do that, we should make sure we have the latest version of `pip`, the 
 (myvenv) C:\Users\Name\djangogirls> python -m pip install -U pip
 ```
 
+> Note: Option -U or --upgrade in the above command allows pip to install a newer version of the installed package if available
+
 ### Installing packages with requirements
 
 A requirements file keeps a list of dependencies to be installed using
--- a/en/django_installation/instructions.md
+++ b/en/django_installation/instructions.md
@@ -171,9 +171,14 @@ Now that you have your `virtualenv` started, you can install Django.
 
 Before we do that, we should make sure we have the latest version of `pip`, the software that we use to install Django:
 
-{% filename %}command-line{% endfilename %}
+{% filename %}Mac OS X and Linux:{% endfilename %}
 ```
-(myvenv) ~$ python -m pip install --upgrade pip
+(myvenv) ~/djangogirls$ pip install -U pip
+```
+
+{% filename %}Windows:{% endfilename %}
+```
+(myvenv) C:\Users\Name\djangogirls> python -m pip install -U pip
 ```
 
 ### Installing packages with requirements


das-g:
As I've written in https://github.com/DjangoGirls/tutorial/pull/1683#issuecomment-727189372, I don't think `python -m pip ...` vs. just `pip ...` was a Windows specific issue. If possible, I'd prefer to keep the instructions for different operating systems the same. So I think we should:

- test whether the command works fine without `python -m` on Windows, MacOS and Linux
- depending on the result, either keep `python -m` in the common instruction for all three, or remove it from that common instruction

nikhiljohn10:
As far as I know, I use Linux at home and Mac at the office and both work well with `pip install -U pip`. I really do not like programming on a windows pc unless it's just HTML and CSS. So I have to test it. I'll let you know the result.

nikhiljohn10:
```
(myvenv) C:\Users\me\Desktop\mysite>pip -V
pip 20.2.3 from c:\users\me\desktop\mysite\myvenv\lib\site-packages\pip (python 3.9)

(myvenv) C:\Users\me\Desktop\mysite>python -m pip -V
pip 20.2.3 from C:\Users\me\Desktop\mysite\myvenv\lib\site-packages\pip (python 3.9)

(myvenv) C:\Users\me\Desktop\mysite>pip install -U pip
Collecting pip
  Using cached pip-20.2.4-py2.py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.2.3
    Uninstalling pip-20.2.3:
      Successfully uninstalled pip-20.2.3
ERROR: Could not install packages due to an EnvironmentError: [WinError 5] Access is denied: 'C:\\Users\\me\\AppData\\Local\\Temp\\pip-uninstall-o1mvxqmn\\pip.exe'
Consider using the `--user` option or check the permissions.

(myvenv) C:\Users\me\Desktop\mysite>pip install --user -U pip
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.

(myvenv) C:\Users\me\Desktop\mysite>python -m pip install -U pip
Collecting pip
  Using cached pip-20.2.4-py2.py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.2.3
    Uninstalling pip-20.2.3:
      Successfully uninstalled pip-20.2.3
Successfully installed pip-20.2.4
```

@das-g  The above result I got from testing Python and Pip concludes that the problem still exists in Windows OS. Thus the proposed changes in this PR are necessary for Windows users.

das-g:
Researching why that might be, I've now read [Why you should use `python -m pip`](https://snarky.ca/why-you-should-use-python-m-pip/).

It explains why `pip install -U pip` doesn't work on Windows (not even with the virtual environment activated, even though the `pip` from the virtual environment is being used) and, as the title says, why one should use `python -m pip` instead of just `pip` even when not on Windows and even when working with virtual environments. TL;DR:
- finer control over Python version used to call `pip` (and by extension, for which Python `pip` will install the packages)
- consistency between different operating systems
- consistency between with vs. without virtual environments
- show and teach best practices

Therefore: Sorry for the back and forth, but I now think we should stick to the `python -m pip` way to call `pip` for all operating systems.

PR #1704, opened 2021-02-19
fix grammar: "as follow" -> "as follows", add articles, minor rephrasing

das-g:
Changes in this pull request:

- replace "as follow" by "as follows"
- add indefinite articles ("a") at some places
- specify that the stuff mentioned at https://github.com/DjangoGirls/tutorial/blob/11e71b5b351292f54648b615712ebea3ebc77b9a/en/html/README.md#L155 indeed are (HTML) _"elements"_
- minor rephrasing (11e71b5) for more natural language flow

(I noticed this when trying to translate the changes from #1681 to German.)


ekohl (approved):
Thanks!

PR #1705, opened 2021-02-19
fix logic: These can't be "another two" article elements

das-g:
These can't be "another two" article elements if no other article elements have been mentioned, yet. (Previously they were `div` elements and the now-`header` element preceding them was then also a `div`. But since #1681 not anymore.)

Changes in this pull request:

- replace "**Another** two `article` elements" with "**The** two `article` elements"


ekohl:
Could you resolve the conflicts?

das-g:
> Could you resolve the conflicts?

@ekohl Done.

ekohl:
Thanks!

Issue #1728, opened 2021-06-24
[Improvement] Development Script


nikhiljohn10:
#### Description

This script verifies all the dependencies and installs all of them except the node environment. It also creates language isolation for fast build and server reload. It is compatible with all languages found inside the LANGS.md file. It can also reset the LANGS.md file after the development process to deploy or commit.

#### Structure

**File Name:** dev.sh (or dev)
**Permission:** `-rwxr--r--` (744)
**Location:** Project root directory
**Command:** ./dev.sh [options] [language]
**Options:**
| Option  | Parameter | Description |
| ------------- | ------------- | ------------- |
| -h  |  | Display usage and options. |
| -l  | Language short form from LANGS.md (en, es, fr, de, pl etc) | Set the temporary language for development. Default is 'en'. |
| -r  |  | Reset the Language file to its original state. |

#### Script
```bash
#!/usr/bin/env bash

HONKIT_LANG=en

help() {
  cat <<EOF
Usage:  ./dev.sh [options]
Options:
           -h       : Display usage and options.
           -l <es>  : Set the temporary language for development. Default is 'en'.
           -r       : Reset the Language file to its original state.
EOF
}

reset_dev() {
  unset HONKIT_LANG
  unset -f help __reset_language __set_language serve __first_run main
}

__reset_language() {
  if [ -f "LANGS.md.bak" ]; then
    mv -f LANGS.md.bak LANGS.md && echo "Language file is reset"
  else
    echo "Language file backup is not found."
  fi
}

__set_language() {
  cp LANGS.md{,.bak} && echo "Language file backup is created."
  HONKIT_LANG_NAME=$(grep "$HONKIT_LANG/" LANGS.md.bak | sed 's/.*\[\(.*\)\].*/\1/')
  grep "$HONKIT_LANG/" LANGS.md.bak 1> LANGS.md && echo "Language list is reduced to $HONKIT_LANG_NAME for development"
  echo "NOTE: You shold not commit 'LANGS.md.bak' in git. You can use './dev.sh -r' to reset the LANGS.md file."
}

__first_run() {
  # Using 'LANGS.md.bak' as a marker to identify first time execution/reset condition
  if [ ! -f "LANGS.md.bak" ]; then
    if which node > /dev/null && which npx > /dev/null; then
      __set_language
    else
      echo "Node.js/npx is not found. Try installing/reinstall Node.js(LTS) using nvm. (https://github.com/nvm-sh/nvm)"
      exit 1
[...]

PR #1729, opened 2021-06-24
[Improvement] Makefile for Development

nikhiljohn10:
### Features

- [x] Verifies all the dependencies and installs all of them except the node environment.
- [x] Provide language isolation for fast build and server reload. **The build time is decreased by 90%**. 
- [x] Compatible with all languages found inside the `LANGS.md` file. 
- [x] The `exit` command is available for restoring the LANGS.md file after the development for deployment or commit. 

### Make
```
Usage: make command [LANG=<language_short_code>]

    LANG: Language shortcodes are found in LANGS.md. Default is 'en'(English).
           ( Refer LANGS.md for shortcodes of Languages available. )

Commands:
  help      - Display make command list.
  dev       - Setup the project and start the development server with debugging enabled.
  check     - Check for the root directory for various dependencies.
  setup     - Setup the temporary language and install node dependencies for the development.
  build     - Build the honkit project.
  build-dev - Build the honkit project with debug log.
  serve     - Start honkit server locally for development.
  pdf       - Generate the PDF version of DjangoGirls tutorial.
  epub      - Generate the EPUB version of DjangoGirls tutorial.
  mobi      - Generate the MOBI version of DjangoGirls tutorial.
  mode      - Shows the development mode status.
  exit      - Exit development mode.

Example:

$ make dev LANG=es

The above command will start the development server using the language Español.
'LANG' argument is only required the first time until the exit command is executed.
```

#1728 Completed


Makefile
```diff
@@ -0,0 +1,40 @@
...
+	fi
+
+setup: reset
+ifneq ($(LANG_FILE), "LANGS.md.bak")
+	@cp LANGS.md LANGS.md.bak && echo "Language file backup is created."
+	@echo "$(LANG_DATA)" > LANGS.md && echo "Language list is reduced to $(LANG_NAME) for development"
+endif
+
+serve:
+	@npx honkit serve || (npm install && npx honkit serve)
```
ekohl, line 36: Shouldn't you create some rule like:

```make
node_modules:
 	@npm install

serve: node_modules check
 	@npx honkit serve
```

Makefile
```diff
@@ -0,0 +1,40 @@
...
+	@echo "Default Language is 'en'(English)."
+	@echo "Commands:"
+	@echo "  init	- Initialise and Start the server."
+	@echo "  serve	- Start honkit server locally for development."
+	@echo "  setup	- Setup the temporary language for the development."
+	@echo "  reset	- Reset the temporarily set language."
+	@echo "  check	- Check for root directory for various dependencies."
+	@echo "  help	- Display command list."
+
+check: book.json LANGS.md package.json $(shell which node) $(shell which npx)
```
ekohl, line 20: Why is the `node` dependency here?

ekohl (commented):
I do wonder why you bother with a `LANGS.md.bak` file. The original is in git so you can revert it.

Another thing to consider is changing the path to LANGS.md. https://honkit.netlify.app/config.html says there's `structure.languages`. However, other than changing `book.json` I'm not sure you can easily pass an override to `honkit serve`.

Makefile
nikhiljohn10:
```
check: package.json book.json LANGS.md
ifneq ($(shell which node && echo $?), "0")
	@echo "Node.js not found. Please install/reinstall Node.js. NVM is recommended for installation(https://github.com/nvm-sh/nvm)."
endif
```
How about this one? (Yes, checking npx was unnecessary I think) It helps to avoid errors in further steps and give out a recommendation. NVM install node without the need for sudo access for running commands.

Makefile
nikhiljohn10: Lets take the scenario where npm modules are already installed. npm will have to run again before start serving when `make serve` is ran. But it is possible since it live reload server. Personally, I dont prefer this method incase I have network problem and my local server will not running. This is waste of time. But if your method is more acceptable generally, i can change that. Let me know your thoughts on my point of view.

nikhiljohn10:
@ekohl Regarding `LANGS.md.bak`, it is used as a marker for detecting development mode. If this file exist, this means few things -> language is isolated, `make check` is success and server is ready to start.

The whole idea behind this script when I begun making it was based on one thought "Why am I waiting for all the languages to build each time I run a build or serve command, while I only dealing with english?"

Using this methods, the build time is decreased by 90%. Without using the script, it take 90s-100s. WIth this script, it takes only 7s-8s. Which means that I got around 12 times faster in my development process (ideal case). This is my philosphy. 

I have checked the `structure.languages` variable. It would be much more complicated than this one. Making error inside the file is so much higher. I have been testing this script in WSL, MacOS and Ubuntu. Everything works fine now. I have added few more commands on the way. I'll post it as command before pushing it to the PR.

nikhiljohn10:
Considering all previous converstion, I have made some tweaking is the whole structure and many targets. Not fully tested.

### New Commands Added

- dev - Initialise and start the development server with debugging enabled. ( Previously `init` )
- build-dev - Build the honkit project with debug log.
- pdf - Generate PDF version of DjangoGirls tutorial.
- epub - Generate EPUB version of DjangoGirls tutorial.
- mobi - Generate MOBI version of DjangoGirls tutorial.
- mode - Shows the development mode status.
- exit - Reset the project and exit development mode. ( Previously `reset` )

nikhiljohn10:
@ekohl I have squashed the commits and pushed it. Let me know you thoughts.

Makefile
```diff
@@ -0,0 +1,81 @@
+LANG      := en
+LANG_FILE := $(shell test -f LANGS.md.bak && echo LANGS.md.bak || echo LANGS.md)
+LANG_DATA := $(shell grep "$(LANG)/" $(LANG_FILE))
+LANG_NAME := $(shell echo "$(LANG_DATA)" | sed 's/.*\[\(.*\)\].*/\1/')
+
+define ebook_support
+	@if which ebook-convert 1> /dev/null; then\
+		npx honkit $(1) ./ ./djangogirls.$(1);\
+	else\
+		echo "Error: ebook-convert is not found. Follow the guide at https://honkit.netlify.app/ebook";\
```
ekohl, line 10: Can this somehow trigger a non-zero exit code so `make` will fail?

Makefile
```diff
@@ -0,0 +1,81 @@
...
+	@echo "  serve     - Start honkit server locally for development."
+	@echo "  pdf       - Generate PDF version of DjangoGirls tutorial."
+	@echo "  epub      - Generate EPUB version of DjangoGirls tutorial."
+	@echo "  mobi      - Generate MOBI version of DjangoGirls tutorial."
+	@echo "  mode      - Shows the development mode status."
+	@echo "  exit      - Reset the project and exit development mode."
+
+check: package.json book.json LANGS.md
+	@if ! which node 1> /dev/null; then\
+		echo "Node.js not found. Please install/reinstall Node.js. NVM is recommended for installation(https://github.com/nvm-sh/nvm).";\
```
ekohl, line 37: Can this somehow trigger a non-zero exit code so `make` will fail?

ekohl (commented):
> Lets take the scenario where npm modules are already installed. npm will have to run again before start serving when make serve is ran. But it is possible since it live reload server. Personally, I dont prefer this method incase I have network problem and my local server will not running. This is waste of time. But if your method is more acceptable generally, i can change that. Let me know your thoughts on my point of view.

I generally agree with that, so another option can be to make the `node_modules` target fail with a message "you need to run `npx install`.  Also note that I think should be a regular target, not a `PHONY`. That way, if the directory exists, make will be happy and not run anything.

nikhiljohn10:
> I generally agree with that, so another option can be to make the `node_modules` target fail with a message "you need to run `npx install`.  Also note that I think should be a regular target, not a `PHONY`. That way, if the directory exists, make will be happy and not run anything.

I'll try that out and get back to you. Thanks for the reply.

Makefile
nikhiljohn10: we can use `$(error ERROR_MESSAGE)` instead of `echo`. That exit with non-zero.

das-g: > we can use `$(error ERROR_MESSAGE)` instead of `echo`. That exit with non-zero.

Cool, I didn't know `make` [provided such functions](https://www.gnu.org/software/make/manual/html_node/Make-Control-Functions.html). Let's use that!

(According to [this StackOverflow answer](https://stackoverflow.com/a/29424978/674064) they are specific to _GNU make_, but I guess that's sufficiently acceptable for us. This Makefile is unlikely to be used with anything else.)

nikhiljohn10: I thought `$(error)` would work. But it did not work as required. It ignores any bash command nearby. So I used the bash command `false` to get the same result. Check [c2c3522](https://github.com/DjangoGirls/tutorial/pull/1729/commits/c2c352223c77ee6f1c4717dc59e33c3d459340f2).

LANGS.md
```diff
@@ -1,20 +1 @@
...
-* [🇮🇹 Italiano (beta)](it/)
-* [🇯🇵 日本語](ja/)
-* [🇰🇵/🇰🇷 한국어 (beta)](ko/)
-* [🇵🇱 Polski](pl/)
-* [🇧🇷 Português-brasileiro](pt/)
-* [🇷🇺 Русский (beta)](ru/)
-* [🇸🇰 Slovenčina (beta)](sk/)
-* [🇹🇷 Türkçe (beta)](tr/)
-* [🇺🇦 Українська](uk/)
-* [🇨🇳 简体中文](zh/)
```
nikhiljohn10, line 20: Found another bug. I accidentally forgot to use the exit command before committing. So it pushed it like this. So I guess I need to block committing while inside dev mode.

nikhiljohn10: Solution is `.git/hooks/pre-commit` file

```bash
#!/bin/sh

if test -f "./.langs"; then
  echo "Error: You can't commit without exiting development mode.\n\n\
Try the following command to exit development mode:\n\n\
$ make exit\n" 1>&2;
  exit 1
fi
```

I'll try to push the changes with this commit. But I am not sure if works as it is inside the `.git` folder.

nikhiljohn10:
> > Lets take the scenario where npm modules are already installed. npm will have to run again before start serving when make serve is ran. But it is possible since it live reload server. Personally, I dont prefer this method incase I have network problem and my local server will not running. This is waste of time. But if your method is more acceptable generally, i can change that. Let me know your thoughts on my point of view.
> 
> I generally agree with that, so another option can be to make the `node_modules` target fail with a message "you need to run `npx install`. Also note that I think should be a regular target, not a `PHONY`. That way, if the directory exists, make will be happy and not run anything.

@ekohl It is fixed in [c01634f](https://github.com/DjangoGirls/tutorial/pull/1729/commits/c01634fd197422794f55dedbb76bd655e7ed9832)

LANGS.md
nikhiljohn10: @das-g Check you test it and let me know if the pre-commit block is working.

README.md
```diff
@@ -32,6 +32,38 @@ First fork the [DjangoGirls/tutorial](https://github.com/DjangoGirls/tutorial) r
...
+  exit      - Exit development mode.
+
+Example:
+
+$ make dev LANG=es
+
+The above command will start the development server using the language Español.
+'LANG' argument is only required the first time until the exit command is executed.
+```
+
```
nikhiljohn10, line 66: Is this required to be mentioned in README?

das-g: *Some* mention in the README is probably good to have. Though if we mention the whole usage instructions, we'll also have to keep them in sync with future evolution of the Makefile.

Maybe it would suffice to just mention that usage instructions are available though `make help`?

nikhiljohn10: Sounds good. Will make changes and push the final commit.

LANGS.md
das-g: AFAIK, you can't include anything in `.git` or its subfolders in a Git repository. And the repository shouldn't be able to force hooks onto its users, as that would lead to arbitrary code execution in situations where the user wouldn't expect it (when just using Git).

The usual solution is to provide project-specific hooks in a separate directory, e.g. `.githooks` at the repo root and
- either instruct users to copy or symlink them into `.git/hooks/`
- or have the build system "install" the hooks (also by copying or symlinking)

As the build system's UX would depend on your hook, I'd go for the second option. See answers to question [Putting Git hooks into a repository](https://stackoverflow.com/q/3462955/674064) on Stack Overflow. [One answer](https://stackoverflow.com/a/3464399/674064) also gives suggestions on how to allow project-provided and user-provided hooks for the same trigger to co-exist. [Another answer](https://stackoverflow.com/a/54281447/674064) mentions an alternative to copying or symlinking the hooks: A setting available since Git 2.9.0 for configuring where the hooks directory shall be.

nikhiljohn10: That is a very good idea. Thanks

nikhiljohn10: Resolved with [b91263f](https://github.com/DjangoGirls/tutorial/pull/1729/commits/b91263fa0980391d6ddf15653f9927dd12c8d0e9)

README.md
nikhiljohn10: Resolved with [e8742d9](https://github.com/DjangoGirls/tutorial/pull/1729/commits/e8742d979c92497d929d011822b8925d12f85dd2)

nikhiljohn10:
@das-g Anymore changes required?

das-g:
> @das-g Anymore changes required?

Sorry, I don't have time for a review right now (maybe mid-next-week). @ekohl can you take another look?

ekohl:
Sorry, also swamped with work.

das-g:
As I doubt that this can break anything already existing, I'll merge now. If needed, we can improve upon that later on.

PR #1730, opened 2021-06-28
[BUGFIX] Anchor tags in Installation page fixed

nikhiljohn10:
In the [Installation page](https://tutorial.djangogirls.org/en/installation/), the `#github` and `#pythonanywhere` anchor tags where having unexpected behaviour. So I have added a suffix `-account` to fix its behaviour. Not sure what is causing this behaviour. In the local testing, it is working fine.


das-g:
Huh. Any idea what causes this and why this change fixes it?

das-g (approved):
LGTM

nikhiljohn10:
> Huh. Any idea what causes this and why this change fixes it?

@das-g Honestly, I did not went to the core of the issue. But my speculation is that there might be multiple id tags and some conflicts occurred between when tag to move to. My logic only thought of making that id unique. That is why i added a suffix to create uniqueness. That way, there will only this tag to move to.

PR #1747, opened 2021-11-15
Fix `dict` `len` confusion

### LINKED ISSUE #1745: Dictionary tutorial can be a bit misleading
### Issue description

This came up in a workshop today.

In the dictionary tutorial, a dict is created with three keys:

```
>>> participant = {'name': 'Ola', 'country': 'Poland', 'favorite_numbers': [7, 42, 92]}
```

Later, there is a section showing how to add a key:

```
>>> participant['favorite_language'] = 'Python'
```

And then showing how you can get the length:

```
>>> len(participant)
3
```

The first command is preceded by "try writing the following command". The second doesn't have an explicit instruction "you can". And the third starts with "go ahead and type this command". So the result of `len(participant)` is correct if the user _didn't_ type the second example (it should be 4). This caused a bit of confusion. I think this could be reworded a bit so that the second part is an instruction and then the return value of `len(participant)` can be 4.

### Language

English but I didn't check others.

das-g:
Changes in this pull request:

- check `len(…)` of `dict` before (optionally) adding additional items, so that result is predictable

fixes #1745

### DIFF
--- a/en/python_introduction/README.md
+++ b/en/python_introduction/README.md
@@ -391,12 +391,6 @@ When should you use a dictionary or a list? Well, that's a good point to ponder.
 - Do you just need an ordered sequence of items? Go for a list.
 - Do you need to associate values with keys, so you can look them up efficiently (by key) later on? Use a dictionary.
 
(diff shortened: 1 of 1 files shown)


ekohl (approved):
Thanks!

PR #1750, opened 2021-12-26
Fix: Command-line Windows tooltip incorrectly built

adamantike:
Because of missing newlines in some languages, the `Opening: Windows` tooltip is incorrectly created as an empty section (included in previous `h1` tag).

This change follows the same format the EN language has, where the `include` is done in a separate line, and each of the included files has a newline at the beginning, to avoid this issue in the future.

Before:
imagen

After:
imagen


das-g:
I'm not quite sure how we can avoid re-introducing this problem when updating the translations from Crowdin. While Crowdin is aware of Markdown syntax, it doesn't really know about GitBook-/HonKit-specific preprocessing so can mess up the syntax of the respective pragmas when re-combining the translated text-snippets.

Maybe we should begin a new paragraph before each expandable box (i.e., have blank line (= two consecutive linebreaks)) already in the English original?

das-g (approved):

PR #1751, opened 2021-12-26
Fix: Anchor tags in Installation page (Spanish)

adamantike:
Fix anchor tags in Installation page, which are the only ones referenced by `SUMMARY.md`, thus they need to match for sidebar navigation to work correctly.


das-g:
We maintain translations [on Crowdin](https://crowdin.com/project/django-girls-tutorial). Can you also make these changes there?

(I'm probably going to merge this PR as a quick-fix, but if it isn't changed in the translation on Crowdin, the problem would be re-introduced when we next update the translation from there.)

adamantike:
I don't think this one depends on Crowdin, though? I don't see anchor tags in there (for example, https://crowdin.com/translate/django-girls-tutorial/290/en-es and https://crowdin.com/translate/django-girls-tutorial/314/en-es

das-g:
> I don't think this one depends on Crowdin, though? I don't see anchor tags in there (for example, https://crowdin.com/translate/django-girls-tutorial/290/en-es and https://crowdin.com/translate/django-girls-tutorial/314/en-es)

In WYSIWYG view on Crowdin, scroll down to "Hidden texts" and expand "Link addresses" by clicking on it. There the target URLs of the links are, including the anchors if the link shall lead to a specific subsection of a chapter.

adamantike:
Thank you, that's really useful! So, is https://support.crowdin.com/cli-tool/ a good starting point for me to validate how different Crowdin is compared to this repo's `master` branch?

das-g:
> So, is https://support.crowdin.com/cli-tool/ a good starting point for me to validate how different Crowdin is compared to this repo's `master` branch?

I don't really know. @magul do you?

PR #1755, opened 2021-12-27
remove mention of PRs for translations from README

das-g:
Let all changes to translatios be made through Crowdin, as otherwise we won't know where the translation is more up-to-date.

related to #1271


fltfx (approved):
Agree with changes

Issue #1694, opened 2020-11-23
variable pa_py_version not substituted at two places


nikhiljohn10:
### Issue description

https://tutorial.djangogirls.org/en/deploy/#configuring-our-site-on-pythonanywhere

In this link pythonanywhere commands are broken. Honkit did not replaces the `pa_py_version` variables.

---

![image](https://user-images.githubusercontent.com/5616249/99990246-9a9bff80-2dd9-11eb-84bb-e3c4a1e751a4.png)
---

Suspected cause: Build time error. But passed through tests.

Ref: Here are 2 piece of code which is deployed.

https://github.com/DjangoGirls/tutorial/blob/221695ee4003b6b90ce6a2c99798e5cdc6eee576/en/deploy/index.html#L796

https://github.com/DjangoGirls/tutorial/blob/221695ee4003b6b90ce6a2c99798e5cdc6eee576/en/deploy/index.html#L800

@das-g @ekohl The pushed code look fine. I am not able to figure out why honkit did not render that variable. Any clue? As for now, can anyone push a hardcoded string "3.8" inside deploy section to replace those variables until we figure this out?

das-g:
Fix proposed in #1695. Feel free to leave a review.

nikhiljohn10:
@ekohl The problem is solved for now with the hotfix. But the variable `pa_py_version` still exists in `book.json`. We should either remove the unnecessary variable or fix the substitution problem at its root. I think you should have kept this issue open until the actual problem is resolved.

das-g:
> @ekohl The problem is solved for now with the hotfix. But the variable `pa_py_version` still exists in `book.json`. We should either remove the unnecessary variable or fix the substitution problem at its root. I think you should have kept this issue open until the actual problem is resolved.

Ah, sorry. The issue being closed way done automatically upon merge based on my commit message.

I agree that either the variable should be removed or the root cause fixed so that the variable can be used. Reopening thus.

nikhiljohn10:
@das-g I have found the root cause after a deep dive in to honkit. Inside `_book`, each page start with `index.html` which is converted from markdown file. This file is compiled with all the included files. In the case of included files, any gitbook variables inside the code area is replaced with the varialbe value defiend inside `book.json`. But in the case of `README.md`(a.k.a. root for index.html), the variables inside code area is not replace and encapsulated inside a pre-code html tags.

[...]

PR #1757, opened 2021-12-27
Upgrade tutorial to Django 3.2

### LINKED ISSUE #1677: Update Django Girls tutorial to Django 3.1
Currently Django released 3.1.3 version. Due to this new people who follow it get in to lots of trouble. I found this from answering then in gitter channel of Djangogirls.

Is there any plans of publishing updated version of tutorial?

At least updating to version 3.0 can resolve most issues.

adamantike:
This diff includes the needed changes for the tutorial to be available for Django 3.2 in the English version.

Most of these changes were already available in the `django-3.1` branch, and the #1678 pull request. Big props to @nikhiljohn10 for the huge initial effort!

Closes #1677


en/deploy/README.md
```diff
@@ -245,7 +245,7 @@ You can also go to the "Files" page and navigate around using PythonAnywhere's b
 Your site should now be live on the public Internet!  Click through to the PythonAnywhere "Web" page to get a link to it. You can share this with anyone you want. :)
 
 
-> **Note** This is a beginners' tutorial, and in deploying this site we've taken a few shortcuts which aren't ideal from a security point of view.  If and when you decide to build on this project, or start a new project, you should review the [Django deployment checklist](https://docs.djangoproject.com/en/2.2/howto/deployment/checklist/) for some tips on securing your site.
+> **Note** This is a beginners' tutorial, and in deploying this site we've taken a few shortcuts which aren't ideal from a security point of view.  If and when you decide to build on this project, or start a new project, you should review the [Django deployment checklist](https://docs.djangoproject.com/en/3.2/howto/deployment/checklist/) for some tips on securing your site.
```
ekohl, line 248: Should we introduce a variable like `django_minor`, `django_doc_version` or something like that?

en/django_urls/README.md
```diff
@@ -93,10 +93,23 @@ The last part, `name='post_list'`, is the name of the URL that will be used to i
 
 If you try to visit http://127.0.0.1:8000/ now, then you'll find some sort of 'web page not available' message. This is because the server (remember typing `runserver`?) is no longer running. Take a look at your server console window to find out why.
 
-![Error](images/error1.png)
+{% filename %}{{ warning_icon }} command-line{% endfilename %}
+
+```shell
```
ekohl, line 98: I'm not sure this really is shell.

en/django_urls/README.md
```diff
@@ -93,10 +93,23 @@ The last part, `name='post_list'`, is the name of the URL that will be used to i
...
+    return _bootstrap._gcd_import(name[level:], package, level)
+  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
+  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
+  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
+  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
+  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
+  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
+  File "/Users/dana/Codes/djangogirls/blog/urls.py", line 5, in <module>
```
ekohl, line 106: Typically the name `ola` is used
```suggestion
  File "/Users/ola/Codes/djangogirls/blog/urls.py", line 5, in <module>
```

en/extend_your_application/README.md
```diff
@@ -75,7 +75,20 @@ That means if you enter `http://127.0.0.1:8000/post/5/` into your browser, Djang
...
+
+```shell
+    return _bootstrap._gcd_import(name[level:], package, level)
+  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
+  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
+  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
+  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
+  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
+  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
+  File "/Users/dana/Codes/djangogirls/blog/urls.py", line 6, in <module>
```
ekohl, line 88:
```suggestion
  File "/Users/ola/Codes/djangogirls/blog/urls.py", line 6, in <module>
```

en/deploy/README.md
adamantike: I think we could, though is there any gotchas we should consider for this templating system where the variable could not be replaced? (e.g. related to #1694?)

Maybe this is something we can try to tackle in a separate PR. Django docs are also available in other languages, so playing around to replace the `/en/` in the docs URL could be useful too.

en/django_urls/README.md
adamantike: It's actually stdout, as it's generated by the `runserver` command, but I don't think we will get any syntax highlighting, no matter the language we specify here. Are we OK with it just being ```?

en/django_urls/README.md
adamantike: What about the rest of the path? Do you have any convention? Preferences about it prioritizing Windows/OSX/Linux?

en/django_urls/README.md
ekohl: Yes, I think that's fine.

en/django_urls/README.md
ekohl: If you look at https://tutorial.djangogirls.org/en/django_start_project/ it's in `~/djangogirls`. I'm not sure about which OS to prioritize.

adamantike: Noted, let's go with `/Users/ola/djangogirls/...` for now!

en/deploy/README.md
ekohl: Sounds ok to me.

ekohl (commented):
I didn't follow the whole tutorial, but this looks ok to me.

das-g (changes requested):
There's another Django 2.2 docs URL [in `en/dynamic_data_in_templates/README.md`](https://github.com/DjangoGirls/tutorial/blob/1a47238ae20fb7ceab6109dbf372257f8bf35a6b/en/dynamic_data_in_templates/README.md?plain=1#L78).

adamantike:
Good catch! Fixed the pending link, and double-checked again. Feel free to review one more time.

das-g:
Has anyone gone through and tested the whole tutorial with these changes? (I didn't have time to do so, yet, and might not get to it anytime soon.)

If not, shall we merge this anyway already, and just wait whether readers report any discrepancies? I'd much prefer if this small but major change was properly tested before releasing, but I also don't want to drag it out too long.

ekohl:
I'm in the same boat @das-g so I'm also leaning to merging.

ekohl (approved):

adamantike:
To try to automate these kinds of checks in the short term, considering there's no much time available, would a simple Dockerfile that runs these commands and apply incremental `diff` files to the Django project be useful?

At least, that will allow building/running the container with a specific Python version, check the stdout to compare with the provided ones in the tutorial, and review that the Django server correctly runs at the end of the tutorial. It would involve duplicating most of the code and the steps provided in the tutorial, though.

das-g:
> To try to automate these kinds of checks in the short term, considering there's no much time available, would a simple Dockerfile that runs these commands and apply incremental `diff` files to the Django project be useful?
> 
> At least, that will allow building/running the container with a specific Python version, check the stdout to compare with the provided ones in the tutorial, and review that the Django server correctly runs at the end of the tutorial. It would involve duplicating most of the code and the steps provided in the tutorial, though.

Setting that up would probably require about as much time as going through the tutorial manually. So it'd only pay off, it can be re-used for checking further changes. That would mean that the duplicated code would have to be kept around and to be kept in sync with the code in the tutorial, which already sounds like quite an anti-pattern.

If we wanted something like that, there should be a single source of truth from which both the rendered tutorial as well as the code to be tested can be derived completely automatically. That'd probably require using something like [Python doctest](https://docs.python.org/3/library/doctest.html) or similar.

While that be pretty cool, it's quite different from our current setup, and I don't know whether the hassle to migrate to a different workflow (that we'd have to define / figure out first) would be worth the hassle.

ekohl:
Given there's been no response in 2 weeks, I'm leaning to merging. Right now @das-g has requested changes, which I think are addressed but that's why I won't push the big green button now.

das-g (approved):

das-g:
Sorry I was (and will be in the near future) busy with other stuff.

:shipit: Let's merge and see if anyone complains about something being broken. :see_no_evil:

adamantike:
Thank you for coming back to it!

If any issue is reported regarding the upgrade, I'm available to work on it.

PR #1758, opened 2021-12-27
Fix: Refer to article tags instead of div

adamantike:
In commit https://github.com/DjangoGirls/tutorial/commit/a9dbbf582e5e863bbb4ef50a16201a2b758a81f9, the `<div>` tags for the blog posts where changed to `<article>`, but this reference wasn't updated.

### DIFF
--- a/en/django_templates/README.md
+++ b/en/django_templates/README.md
@@ -19,7 +19,7 @@ To print a variable in Django templates, we use double curly brackets with the v
 {{ posts }}
 ```
 
-Try this in your `blog/templates/blog/post_list.html` template. Open it up in the code editor, and replace everything from the second `<div>` to the third `</div>` with `{{ posts }}`. Save the file, and refresh the page to see the results:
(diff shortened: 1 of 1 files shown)


das-g:
Good catch!

en/django_templates/README.md
```diff
@@ -19,7 +19,7 @@ To print a variable in Django templates, we use double curly brackets with the v
 {{ posts }}
 ```
 
-Try this in your `blog/templates/blog/post_list.html` template. Open it up in the code editor, and replace everything from the second `<div>` to the third `</div>` with `{{ posts }}`. Save the file, and refresh the page to see the results:
+Try this in your `blog/templates/blog/post_list.html` template. Open it up in the code editor, and replace the existing `<article>` blocks with `{{ posts }}`. Save the file, and refresh the page to see the results:
```
das-g, line 22: Let's use the proper terminology:
```suggestion
Try this in your `blog/templates/blog/post_list.html` template. Open it up in the code editor, and replace the existing `<article>` elements with `{{ posts }}`. Save the file, and refresh the page to see the results:
```

("blocks" would be doubly confusing, as Django's templates have a concept named like that.)

das-g (changes requested):

das-g (approved):

PR #1769, opened 2022-09-08
Add a job to automatically update translations

ekohl:
This job runs weekly (or explicitly manual). It lists the languages in a way tht GitHub Actions can reuse it in a matrix. Then for every language it updates the translations and creates a PR in a branch for every language.

### DIFF
--- a/.github/list_languages
+++ b/.github/list_languages
@@ -1,0 +1,20 @@
+#!/usr/bin/env python3
+
+import json
+import os
+from pathlib import Path
+
+
+def main():
+    directory = Path(__file__).absolute().parent.parent
+    languages = sorted(x.name for x in directory.iterdir()
+                       if x.is_dir() and len(x.name) == 2 and x.name != 'en')
+
+    if 'GITHUB_ACTION' in os.environ:
+        print(f'::set-output name=languages::{json.dumps(languages)}')
+    for language in languages:
+        print(language)
+
+
+if __name__ == '__main__':
+    main()
--- a/.github/workflows/translations.yml
+++ b/.github/workflows/translations.yml
@@ -1,0 +1,52 @@
+name: Update translations
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '4 4 * * 0'
+
+env:
+  GIT_AUTHOR_NAME: Django Girls Automation
+  GIT_AUTHOR_EMAIL: automation@theforeman.org
+
+jobs:
+  list_languages:
+    name: List languages
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'djangogirls'
+    outputs:
+      languages: ${{ steps.set_list.outputs.languages }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set the language list
+        id: set_list
+        run: ./.github/list_languages
+
+  update_language:
+    name: 'Update ${{ matrix.language }} translations from Crowdin'
+    needs: list_languages
+    if: ${{ needs.list_languages.outputs.languages != '[]' }}
+    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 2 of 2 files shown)


ekohl:
> This job runs weekly (or explicitly manual).

I changed my mind. Let's start with explicitly manual first and then move to cron.

.github/workflows/translations.yml
```diff
@@ -0,0 +1,50 @@
...
+          rsync -av master/${{ matrix.language }}*/* ${{ matrix.language }}/
+          rm -rf ${{ matrix.language }}.zip master
+      - name: Open a PR
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: "Update ${{ matrix.language }} translations from Crowdin"
+          branch: "update_translations/${{ matrix.language }}"
+          title: "Update ${{ matrix.language }} translations from Crowdin"
+          body: ''
+          delete-branch: true
```
ekohl, line 50: Note that this removes any commits in `update_translations` and rebases it back to master. So this isn't good if you want to work on the branch.

.github/list_languages
```diff
@@ -0,0 +1,20 @@
...
+
+import json
+import os
+from pathlib import Path
+
+
+def main():
+    directory = Path(__file__).absolute().parent.parent
+    languages = sorted(x.name for x in directory.iterdir()
+                       if x.is_dir() and len(x.name) == 2 and x.name != 'en')
```
ekohl, line 11: I don't know if there are languages which we have but no longer translate.

das-g:
For automatically scheduled PR, I think we'd want to only consider languages for which the translation of Crowdin is complete (=has 100% "translated" coverage). For languages _not_ marked "(beta)" in our [language overview](https://tutorial.djangogirls.org/), we might even want 100% proofread on Crowdin.

ekohl:
@das-g I'm leaning to maintaining a `langs.json` file. We can also write a simple converted to generate `LANGS.md` based on it. That avoids parsing and allows the additional metadata of the exact code. Thoughts?

das-g:
> Similar, `pt` is `pt-PT`.

Are you sure?
https://github.com/DjangoGirls/tutorial/blob/50d6da3352f0ee116aa329b6b39f7d2f79cf317c/LANGS.md?plain=1#L15
seems to indicate _our_ `pt` corresponds to [`pt-BR`](https://crowdin.com/project/django-girls-tutorial/pt-BR). IIRC, the efforts to [translate the tutorial to `pt-PT`, too](https://crowdin.com/project/django-girls-tutorial/pt-PT), are more recent and the result hasn't been released, yet.

> I'm leaning to maintaining a `langs.json` file. We can also write a simple converted to generate `LANGS.md` based on it. That avoids parsing and allows the additional metadata of the exact code. Thoughts?

As soon as we have two variants of the same language, we should probably switch to the language variant abbreviations (that's what Crowdin seems to use) instead of the language abbreviations for our directory names, anyway. But yes, it makes sense not to have to parse `(beta)` out of a markdown file, so having that information in JSON and generating the Markdown file might be a good idea.

ekohl:
I found that https://crowdin.com/backend/download/project/django-girls-tutorial/pt.zip is a 404. See https://github.com/ekohl/tutorial/runs/8251377015?check_suite_focus=true (ignore `en`, I skipped that explicitly).

matthieugouel:
Hey, I recently came across Crowdin and added some translations. 

I see from Crowdin contribution page that people frequently make some translations here and there. 
I'm trying to understand how the work on Crowdin is merged on the repository. Do you do manual PRs from time to time? I quickly looked at the recent commits and didn't seem to see that so I was wondering if the Crowdin project was still used (before I spend some time on it).

fltfx:
@ekohl Can this PR be closed?

PR #1770, opened 2022-09-08
Change os.path.join to Pathlib /

dannluciano:


thibaudcolas:
👋 this has already been merged in #1786 and #1787.

Issue #1771, opened 2022-09-11
Missing slovak translation in Python introduction


carnoguzu:
### Issue description

Missing translation to Slovak in https://tutorial.djangogirls.org/sk/python_introduction/: 
volume to hlasitost
```
hlasitost = 57
if hlasitost < 20:
    print("Je to pomerne tiche.")
elif 20 <= volume < 40:
    print("Je to fajn ako hudba na pozadi")
elif 40 <= volume < 60:
    print("Super, pocujem vsetky detaily")
elif 60 <= volume < 80:
    print("Fajn na party")
elif 80 <= volume < 100:
    print("Trochu hlasne!")
else:
    print("Bolia ma usi! :(")
```
### Language

Slovak

PR #1772, opened 2022-09-11
Fix: Forgotten variable translation

HonzaKral:
Variable `hlasitost` was translated but a few occurrences of the original `volume` were left behind.

Fixes https://github.com/DjangoGirls/tutorial/issues/1771


das-g:
Thank you for this fix, @HonzaKral. :heart:

I'm merging this as a quick-fix. @HonzaKral or @DjangoGirls/sk-pr, can you modify the translation on Crowdin accordingly, so that we won't have a regression when we update if from there again?

jenxness:
Thank you for the fix @HonzaKral, and @das-g for the quick merge. I'll update the Crowdin version if Honza hasn't done so already.

PR #1778, opened 2022-11-25
Remove Atom from code editor suggestions

ekohl:
Atom will be [sunset](https://github.blog/2022-06-08-sunsetting-atom/) and unmaintained after December 15th 2022.


das-g (approved):
:cry:

Issue #1676, opened 2020-11-04
Switch from master to main in tutorial


jwhitlock:
[As of October 1, 2020](https://github.com/github/renaming#new-repositories-use-main-as-default-branch-name), GitHub defaults to ``main`` as the the default branch for new projects. This means a few screenshots and instructions may be out of date.

Even though both ``main`` and ``master`` are English words, this change is being applied for all users, so it should affect all languages and operating systems.

[In July 2020](https://github.blog/2020-07-27-highlights-from-git-2-28/), git 2.2.8 introduced a new config item ``init.defaultBranch`` that could be used to sync the user's new repository to the GitHub default. [Starting our Git repository](https://tutorial.djangogirls.org/en/deploy/#starting-our-git-repository) could be updated to set this, and re-arranged to set all config before init:

```
$ git config --global --add init.defaultBranch main
$ git config --global user.name "Your Name"
$ git config --global user.email you@example.com
$ git init
Initialized empty Git repository in ~/djangogirls/.git/
```

There are additional changes that would be needed. ``git status`` would now say you are on branch ``main`` instead of ``master``, the push command would be ``git push -u origin main``, etc. There would also be new screenshots, and there may be changes to the PythonAnywhere deploy.

This repository could also make the change from ``master`` to ``main``, but I think that could be tracked as a separate issue.

ekohl:
Not everything ships git 2.2.8 yet so you need to account for that. Just because Github changed its default, doesn't mean you need to follow it. I think that in terms of support you don't really need to do anything right now for the best result during workshops. AFAIK the current instructions still work.

nikhiljohn10:
Here is the current list of platforms and the git version they use or available for download or installation:


| Platform  	| Platform version / URL                              	| Git version 	|
|-----------	|-----------------------------------------------------	|-------------	|
| source    	| https://github.com/git/git                          	| 2.29.2      	|
| download  	| https://git-scm.com/downloads                       	| 2.29.2      	|
| windwos   	|                                                     	| 2.29.2      	|
| brew      	|                                                     	| 2.29.2      	|
| macport   	|                                                     	| 2.29.2      	|
| mac-dmg   	| https://sourceforge.net/projects/git-osx-installer/ 	| 2.27.0      	|
| debian    	| jessie 8                                            	| 2.1.4       	|
|           	| stretch 9                                           	| 2.11.0      	|
|           	| buster 10                                           	| 2.20.1      	|
[...]

PR #1780, opened 2022-12-01
Ignore the default branch name problem

### LINKED ISSUE #1779: Git branch rename is allowed until a commit exists
### Issue description

On https://tutorial.djangogirls.org/en/deploy/ it tells the reader to rename the branch to main if it's master (for git < 2.28). However, you can only rename it after something has been committed.

### Language

It was in English, but probably applies to all.

### Operating system

Mac OS X.

ekohl:
With older versions of git the command to rename the branch fails if there is no branch yet. We're also making the tutorial harder than it needs to be.

This ignores it and simply uses the default by pushing HEAD instead of main. GitHub is then smart enough to use that as default. Future clones will also simply copy the default branch from GitHub.

Fixes #1779

### DIFF
--- a/en/deploy/install_git.md
+++ b/en/deploy/install_git.md
@@ -7,8 +7,6 @@ data-collapse=true ces-->
 
 You can download Git from [git-scm.com](https://git-scm.com/). You can hit "next" on all steps except for two: in the step where it asks to choose your editor, you should pick Nano, and in the step entitled "Adjusting your PATH environment", choose "Use Git and optional Unix tools from the Windows Command Prompt" (the bottom option). Other than that, the defaults are fine. Checkout Windows-style, commit Unix-style line endings is good.
 
-During installation, if you are presented with the option of "Adjusting the name of the initial branch in new repositories", please choose to "Override the default" and use "main". This will align your installation of Git with the broad direction of the global developer community, and the "main" branch will be used through the remainder of this tutorial. Please see https://sfconservancy.org/news/2020/jun/23/gitbranchname/ and https://github.com/github/renaming for further discussion of this subject. 
-
 Do not forget to restart the command prompt or PowerShell after the installation finished successfully.
 <!--endsec-->
 
@@ -16,8 +14,6 @@ Do not forget to restart the command prompt or PowerShell after the installation
 data-collapse=true ces-->
 
 Download Git from [git-scm.com](https://git-scm.com/) and follow the instructions.
-
-During installation, if you are presented with the option of "Adjusting the name of the initial branch in new repositories", please choose to "Override the default" and use "main". This will align your installation of Git with the broad direction of the global developer community, and the "main" branch will be used through the remainder of this tutorial. Please see https://sfconservancy.org/news/2020/jun/23/gitbranchname/ and https://github.com/github/renaming for further discussion of this subject. 
 
 > **Note** If you are running OS X 10.6, 10.7, or 10.8, you will need to install the version of git from here: [Git installer for OS X Snow Leopard](https://sourceforge.net/projects/git-osx-installer/files/git-2.3.5-intel-universal-snow-leopard.dmg/download)
 
@@ -31,15 +27,6 @@ data-collapse=true ces-->
 $ sudo apt install git
 ```
 
-### Adjusting your default branch name
-
-This will align your installation of Git with the broad direction of the global developer community, and the "main" branch will be used through the remainder of this tutorial. Please see https://sfconservancy.org/news/2020/jun/23/gitbranchname/ and https://github.com/github/renaming for further discussion of this subject. 
-
-{% filename %}command-line{% endfilename %}
-```
-$ git config --global --add init.defaultBranch main
-```
-
 <!--endsec-->
 
 <!--sec data-title="Installing Git: Fedora" data-id="git_install_fedora"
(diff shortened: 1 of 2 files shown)


nikhiljohn10 (commented):
LGTM

en/deploy/install_git.md
```diff
@@ -7,8 +7,6 @@ data-collapse=true ces-->
 
 You can download Git from [git-scm.com](https://git-scm.com/). You can hit "next" on all steps except for two: in the step where it asks to choose your editor, you should pick Nano, and in the step entitled "Adjusting your PATH environment", choose "Use Git and optional Unix tools from the Windows Command Prompt" (the bottom option). Other than that, the defaults are fine. Checkout Windows-style, commit Unix-style line endings is good.
 
-During installation, if you are presented with the option of "Adjusting the name of the initial branch in new repositories", please choose to "Override the default" and use "main". This will align your installation of Git with the broad direction of the global developer community, and the "main" branch will be used through the remainder of this tutorial. Please see https://sfconservancy.org/news/2020/jun/23/gitbranchname/ and https://github.com/github/renaming for further discussion of this subject.
```
ekohl, line 10: I have not checked about this question. Should we ignore it and assume users will keep the default value?

das-g:
As we give example output of `git status` and of `git push …`, shouldn't we at least mention that it might read `master` instead of `main` for some?

elena (changes requested):
Discussing and DjangoGirlsAdelaide/PyConAU23 this is a far gnarlier problem than it might seem. The `main` branch conversation is very important to the community. Just deleting it as a concept doesn't feel right.. 

The real problem is that older versions of git are still initiating a repo on the branch master and git should be updated. Removing the documentation below just doesn't necessarily feel like the right approach, perhaps we could make this hiccup less confusing though -- let's keep having the conversation. 

[edit]: just some more background: 
python-core discussion: https://discuss.python.org/t/communitys-take-on-changing-master-branch-to-main/4462/11
from that thread: https://www.cnet.com/tech/computing/microsofts-github-is-removing-coding-terms-like-master-and-slave/

ekohl:
I respectfully disagree that it's a problem. The tutorial is already hard enough and adding more concepts just confuses people. They just have a single branch and that's good enough. They have so many new things coming at them at once that I'm sure nobody will even notice. And as time goes on, it'll simply disappear when more people have a newer git version.

But I'll also note that I'm of the opinion that it'll all virtue signalling and doesn't actually make any difference at all. It still annoys me on an almost daily basis that you now have to remember what the default branch is because only part of the world has switched over. Let's not punish the participants in the same way and make their lives a little bit easier.

amakarudze (commented):
Seeing that only the Django Girls website now uses `main` as default branch while all other repos still use `master`, I am inclined to agree with all those who feel it is unnecessary to burden attendees with this. We can't really make life difficult for them when we have not completely made that change ourselves. They will still encounter `master` branch when they contribute to our repos.

Since we are already working on simplifying the tutorial, I agree with them just working with the default branch and then let them get involved in this discussion as they progress with their coding journey.

With that in mind, I would agree with adding the statement that their default branch maybe `master` or `main` in the Git section and we can merge this in.

nikhiljohn10:
The movement to change the default branch name was driven by [this](https://sfconservancy.org/news/2020/jun/23/gitbranchname/) as far as I know. Github, Gitlab and others are pushing the changes as they all felt the need for change. It is inevitable, the default branch name is going to be `main` in almost all maintained git software in the near future.

Also, I don't think that the name change is going to confuse them if we mention that "Currently, default branch name is `main` and previously it was `master`. You could encounter it in some repositories."

With that said, I hope we can merge this PR until/unless it affects any other technical aspects of DjangoGirls repos.

ekohl:
> If this specific approach is controversial and needs further consideration, perhaps in the meantime we could at least change the tutorial enough so people no longer face the error? There are two options that are really simple:

I don't think it's controversial: it works and makes lives easier.

You may see some cases where the exact example output may differ a little bit, but I doubt students will notice this.

Right now I'm providing a solution to a real problem. I don't see anyone else submitting any alternative. I'm getting really frustrated that there is no movement on this issue.

thibaudcolas:
@ekohl reading through #1676, #1742, #1779, and now this pull request – it seems reasonable to me to assume others feel differently about this than you do?

With the caveat that I don’t personally help maintain the tutorial – I’d suspect the main reason there’s no movement is that the solution you’re providing essentially undoes a lot of pre-existing work by other people. You’re also suggesting bigger changes (removing all notions of branch renaming for all tutorial users) than are strictly needed to address the problem at hand (make the branch renaming work for users below git 2.28).

I’ve proposed alternatives that I think are suitable right in the comment you’re responding to. I’ve made a PR for the one that has my preference at #1803 if it helps making it clearer it’s a suitable alternative.

ekohl:
> With the caveat that I don’t personally help maintain the tutorial – I’d suspect the main reason there’s no movement is that the solution you’re providing essentially undoes a lot of pre-existing work by other people. You’re also suggesting bigger changes (removing all notions of branch renaming for all tutorial users) than are strictly needed to address the problem at hand (make the branch renaming work for users below git 2.28).

Yes, and I see this as a feature. I've been a mentor at least 8 times and I always notice that most students are overloaded with new information. This is such an irrelevant point for them that it just distracts them from the real work. They can only learn so much in a single day, so let's make sure they learn Python, Django and HTML & CSS. At the events I attended I also heard regular suggestions to drop git, pushing to github & deploying altogether because it's too much and barely teaches them anything. I'm not that extreme, but I do think we should shield them from the default branch problem.

> I’ve proposed alternatives that I think are suitable right in the comment you’re responding to. I’ve made a PR for the one that has my preference at #1803 if it helps making it clearer it’s a suitable alternative.

While I disagree with it, I do appreciate you being constructive in presenting an alternative.

amakarudze (approved):
Merging this in to end this discussion now. We seem to be going back in circles and ending at the same solution or no solution at all.

PR #1781, opened 2023-01-03
Add RunCode Cloud instructions

amakarudze:
This PR introduces cloud development instructions using [RunCode.io](https://runcode.io) so attendees don't have to go through the long installation process, but instead can code in the cloud.

Changes in this pull request:

- Add `Installation (RunCode setup)` link to SUMMARY.md.
- Add `cloud_development_setup` folder with `README.md` and `instructions.md` for instructions on setting up RunCode Cloud environment.
- Add RunCode section to `intro_to_command_line/open_instructions.md`.
- Add RunCode section to `installation/README.md`.
- Add RunCode Django project setup instructions to `django_start_project/README.md`
- Add Runcode instructions on opening Python prompt to `python_introduction/prompt.md`


das-g:
I'll take a closer look at this PR later this week.

Meanwhile, I'd like to suggest to use ["semantic linefeeds"](https://rhodesmill.org/brandon/2012/one-sentence-per-line/) a.k.a. ["semantic line breaks"](https://sembr.org/) for newly added paragraphs. Though most of the Django Girls tutorial isn't written that way, placing source line breaks at semantically meaningful spots it the text (and especially between sentences) facilitates both, commenting on individual statements or thoughts in the text as part of the review process, as well as future editing and diff-ing.

en/SUMMARY.md
```diff
@@ -11,6 +11,7 @@
   * [GitHub](installation/README.md#github-account)
   * [PythonAnywhere](installation/README.md#pythonanywhere-account)
 * [Installation (chromebook)](chromebook_setup/README.md)
+* [Installation (RunCode Cloud)](cloud_development_setup/README.md)
```
GirlGeekUpNorth, line 14: I think in terms of organisation we should maybe have one page for Installation - use this to explain that there are two ways to do this (explain the fiferences and pros/cons of each) then from there have the two installation guides as sub-pages so people pick one or the other, before miving to the next step. It will hopefully stop people getting confused and trying to do both installation setup guides.

en/cloud_development_setup/README.md
```diff
@@ -0,0 +1,5 @@
+# RunCode Cloud Environment setup
+
+> **Note** If you already worked through the [installation steps](../installation/README.md), no need to do this again – you can skip straight ahead to [Introduction to Python](../python_introduction/README.md).
```
GirlGeekUpNorth, line 3: I would amend to "If you have already worked through the installation steps, you do not need to also complete the RunCode Cloud Environment setup. Please skip straight to Introduction to Python"

en/cloud_development_setup/instructions.md
```diff
@@ -0,0 +1,24 @@
+You can [skip right over this section](http://tutorial.djangogirls.org/en/installation/#install-python)
```
GirlGeekUpNorth, line 1: This as a starting sentence doesn't make sense, I would not begin a page with "You can skip right over this section" as then there's not point it being there?

GirlGeekUpNorth: Ah! I've realised what it should be due to the second sentence - amend to "If you have already worked through the installation steps, you do not need to also complete the RunCode Cloud Environment setup. Please skip straight to Introduction to Python"

en/cloud_development_setup/instructions.md
```diff
@@ -0,0 +1,24 @@
+You can [skip right over this section](http://tutorial.djangogirls.org/en/installation/#install-python) 
+if you're not using RunCode Cloud Development environment. If you are, your installation experience will be a little different.
```
GirlGeekUpNorth, line 2: Move "If you are" to a new line

GirlGeekUpNorth: And add in - "If you are installing the RunCode Cloud Development environment, your installation expereince will be a littel different. You do not need to insall anything locally, you just need to create three (free) online accounts.

en/cloud_development_setup/instructions.md
```diff
@@ -0,0 +1,24 @@
...
+You can ignore the rest of the installation instructions as you do not need to install anything locally, 
+you just need to create three online accounts. 
+
+## About RunCode
+RunCode is a cloud development environment which people can use without the need to install Python, Django and Visual 
+Studio Code editor locally on their machines. This cloud environment enables anyone to work from any device that has
+an internet connection, including cellphones, tablets, laptop or desktop.
+
+This removes the need for you to install packages on a borrowed laptop or the need for you to own a laptop to attend the workshop or follow this tutorial from home. This will also save you time required for setting up your development
+environment and you can always do the installation later own as a follow-up after finishing the tutorial.
```
GirlGeekUpNorth, line 12: Full stop after environment. New line - You can always give the installation steps (inc link) a go later as follow-up learning.

en/cloud_development_setup/instructions.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Create a RunCode account
+Go to [RunCode.io](https://runcode.io/) and sign up for a new, free user account. You need to have a 
+[Google.com](https://www.google.com/intl/en-GB/gmail/about/) account or [GitHub.com](https://github.com/) 
+which you can sign up with.
+
+## Create a GitHub account
```
GirlGeekUpNorth, line 19: Because you need a GitHub account (or it is useful to have one) to create the runCode account, I would put this first

en/django_installation/instructions.md
```diff
@@ -104,6 +104,26 @@ $ python3 -m venv myvenv
 
 <!--endsec-->
 
+<!--sec data-title="Virtual Environment: RunCode" data-id="virtualenv-runcode" -->
```
GirlGeekUpNorth, line 107: If this is added into an existing resource to show an alternative methods when using RunCode, you will also need to add further up that there are two ways to do this. The first, if you have followed the installation steps (followed by existing instructions), then another sub heading for if you have installed RunCode followed by these new instructions.

en/django_start_project/README.md
```diff
@@ -252,6 +252,15 @@ or on Glitch:
 https://name-of-your-glitch-project.glitch.me
 ```
 
+If you are using RunCode cloud platform, the URL will look something like this:
```
GirlGeekUpNorth, line 255: Same again, if we ever include a line such as "If you're using the RunCode environment" in the existing instructions, we need to preceed the previous instructions with "If you have followed the installation guide" to distinguish upfront that these instructions are not applicable.

en/intro_to_command_line/open_instructions.md
```diff
@@ -26,3 +26,12 @@ Go to Applications → Utilities → Terminal.
 It's probably under Applications → Accessories → Terminal, or Applications → System → Terminal, but that may depend on your system. If it's not there, you can try to Google it. :)
 
 <!--endsec-->
+
+<!--sec data-title="Opening: RunCode" data-id="runcode_prompt" data-collapse=true ces-->
```
GirlGeekUpNorth, line 30: As previous comments - sub heading also needed for existing instructions above

en/python_installation/README.md
```diff
@@ -12,5 +12,8 @@ Python originated in the late 1980s and its main goal is to be readable by human
 
 > **Note** If you already worked through the [installation steps](../installation/README.md), there's no need to do this again – you can skip straight ahead to the next chapter!
 
+> **Note** If you're using RunCode cloud environment, you do not need to install Python as it is pre-installed on their platform
```
GirlGeekUpNorth, line 15: Have we not already explained this and 'should' have stopped anyone suing RunCode form even reading this page? I want to avoid constantly saying "Don't do this if you did this" and instead, direct people to make a choicce and then only see instructions relevent to them from this point.

en/python_introduction/prompt.md
```diff
@@ -15,3 +15,12 @@ Python {{ book.py_release }} (...)
...
 ```
+
+<!--sec data-title="Opening: RunCode Cloud Environment" data-id="runcode_prompt" data-collapse=true ces-->
+
+To open the Python terminal on RunCode, go to Workspaces → New Workspace → Jupyter Lab. This will open a Python prompt 
+which is depicted by `>>>`. If you already opened an Ubuntu terminal open, in the [Intro to Command Line](../intro_to_command_line/README.md) chapter, you can just type `python3` in the prompt as shown below to open a Python prompt.
+
+`ubuntu@runcode:~$ python3`
+
+<!--endsec-->
```
GirlGeekUpNorth, line 26: My final comment would just be that if we find there are lots of places where we need to keep editing sections to say, "if you followed the installation guide do this" then "if you're using RunCode do this", then it might be worth looking at creating two seperate versions of the tutorial for each method. However, if this is minimal, you can leave as it is, as long as we're clear when we reach instructions that these will be diffent depending on setup upfront.

GirlGeekUpNorth (changes requested):
Changes/suggestions added via individual comments

en/python_introduction/prompt.md
das-g: Where different instructions are alternatives to each other, we could probably make that much clearer by having them in a tab-headed section rather than in an expandable section. We've discussed this possibility some time in the past (I'll have to look up the issues or PRs about that), but I think that was back when we used GitBook for the tutorial, so we'd have to see whether according features or plugins are available for Honkit, which is now used to render the tutorial.

das-g: > I'll have to look up the issues or PRs about that

- #188
- https://github.com/DjangoGirls/tutorial/issues/862#issuecomment-247970100
- #1173

en/django_start_project/README.md
amakarudze: I have just moved the RunCode instructions to one place so this is no longer necessary. I now just point the readers to the RunCode section.

> Same again, if we ever include a line such as "If you're using the RunCode environment" in the existing instructions, we need to preceed the previous instructions with "If you have followed the installation guide" to distinguish upfront that these instructions are not applicable.

I have just moved the RunCode instructions to one place so this is no longer necessary. I now just point the readers to the RunCode section.

en/intro_to_command_line/open_instructions.md
amakarudze: > As previous comments - sub heading also needed for existing instructions above

I have added instructions in the Installation Readme to give instructions that one should follow instructions for their platform and skip all others to avoid the repetition of adding that everywhere so this should now work as is with the new changes.

en/python_introduction/prompt.md
amakarudze: I have moved all RunCode setup instructions to one file to make a separate installation guide as is done for the Chromebook and the only place where the RunCode instructions had to be left was for opening the command line. I felt it just has to be there as that section is also part of the `Introduction to Command Line` which all users need to go through. I avoided making those comments in many sections so I just added notes on sections that RunCode users need to skip so which should make the instructions clearer.

This seemed easier than reading about tab-headed sections and changing the whole tutorial until we decided to change to that.

en/installation/README.md
```diff
@@ -14,38 +14,68 @@ If you are attending one of the [Django Girls events](https://djangogirls.org/ev
...
 # Installation
-In this tutorial you will be building a blog. In order to do that, as you go through the tutorial you'll be instructed on how to install various software on your computer and set up some online accounts as they are needed. This page gathers all of the installation and sign-up instructions in one place (which is useful for some workshop formats).
+In this tutorial you will be building a blog. In order to do that, as you go through the tutorial you'll be instructed on how to
+install various software on your computer and set up some online accounts as needed (if you are using local desktop environment) or instructed on how to create online accounts only (if you are using cloud development). This page gathers all of the installation and sign-up instructions in one place (which is useful for some workshop formats).
 
-<!--sec data-title="Chromebook setup (if you're using one)"
-data-id="chromebook_setup" data-collapse=true ces-->
-{% include "/chromebook_setup/instructions.md" %}
-<!--endsec-->
+## Local Installation
```
GirlGeekUpNorth, line 20: Move this underneath Cloud Development. This way, RunCode is first (because it's the easiest) then local, which is split into the operating system instructions

PR #1786, opened 2023-03-10
Update STATIC_ROOT to the latest Django style

ryu22e:
Changes in this pull request:

- Changed to use `pathlib.Path` instead of `os.path`


das-g (approved):
Looks good to me.

Thank you!

PR #1788, opened 2023-03-10
Fix broken links in Japanese text

ryu22e:
Changes in this pull request:

- Fix broken links in Japanese text

before
before

after
after


das-g (approved):
Good catch!

We should probably change this in the English original, too (even if it's not strictly needed there), so that other translations with different character sets won't also run into this problem.

Issue #1785, opened 2023-02-19
Ebook version


ololoe:
I cannot find anywhere ebook version of this tutorial, but I have old version on my drive. Is it gone?

das-g:
I'm not sure. I guess we might have lost that when switching from GitBook to Honkit.

Is there any interest in using e-book versions of the tutorial?

@ekohl Would it be feasible to generate e-book files again?

ololoe:
If you talk about me - I'm interested. If possible, please make readable CSS names, I had to edit CSS in order to read ebook on e-paper device, non black text is displayed as grey there, and edit text size too.

Can you tell me how to generate ebook from markdown?

das-g:
> Can you tell me how to generate ebook from markdown?

From general Markdown in common Markdown dialects, [Pandoc](https://pandoc.org/) would be an option. But for the GitBook-specific Markdown this tutorial is written in, [HonKit's ebook output formats](https://honkit.netlify.app/ebook.html) are probably our best bet.

ekohl:
https://honkit.netlify.app/ebook.html suggests we can easily do that.

By now it's also possible to generate the website in GitHub Actions without storing it in the `gh-pages` branch, so that's something I'd do before storing large files. The repository is already rather large because of it now.

I may be able to take a look tomorrow.

ololoe:
> > Can you tell me how to generate ebook from markdown?
> 
> From general Markdown in common Markdown dialects, [Pandoc](https://pandoc.org/) would be an option. But for the GitBook-specific Markdown this tutorial is written in, [HonKit's ebook output formats](https://honkit.netlify.app/ebook.html) are probably our best bet.

Can it generate index too?
If you've done it, can you share how you did it? I want to generate ebook from some other tutorials.

ekohl:
> I may be able to take a look tomorrow.

It wasn't tomorrow, but https://github.com/DjangoGirls/tutorial/pull/1791 should at least convert it to using GitHub Actions.

After that's merged, solving means changing the build steps in the action to also generate the eBooks. That looks rather trivial, but it doesn't mention anything about linking to it. That's something we may need to look into.

ololoe:
Any progress?

lapwat:
Here you go:
[Introduction_·_HonKit.epub.zip](https://github.com/DjangoGirls/tutorial/files/11311013/Introduction_._HonKit.epub.zip)

Built with [papeer](https://github.com/lapwat/papeer)

[...]

PR #1791, opened 2023-04-06
Switch to actions/deploy-pages

ekohl:
This uses the currently beta feature of using GitHub Actions to deploy GitHub Pages, rather than using a gh-pages branch. This keeps the repository size smaller.

While this feature is beta, it is used [by default](https://github.blog/2022-08-10-github-pages-now-uses-actions-by-default/).

This does require a change on the repo itself and since I don't have the permissions, someone else will need to do this. https://github.blog/changelog/2022-07-27-github-pages-custom-github-actions-workflows-beta/ shows how to.

This will make it easier to publish large artifacts, such as ebooks (see https://github.com/DjangoGirls/tutorial/issues/1785).

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,27 +1,47 @@
+---
 name: Build and Deploy
+
 on:
   push:
     branches:
       - master
 
+permissions:
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: "pages"
+  cancel-in-progress: true
+
 jobs:
-  build-and-deploy:
+  build:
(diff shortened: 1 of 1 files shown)


das-g:
I've now changed the "Source" setting at https://github.com/DjangoGirls/tutorial/settings/pages to "GitHub Actions". Are any other changes to settings needed?

ekohl:
@das-g no, that should be sufficient. Want to try and merge it to see how it performs?

ekohl:
I think that looks good. Shall we delete the `gh-pages` branch now?

PR #1795, opened 2023-04-27
build and deploy ePub e-book files

### LINKED ISSUE #1785: Ebook version

das-g:
~See a run of these actions [here](https://github.com/das-g/django-girls-tutorial/actions/runs/4824138230).~

~Results available from `https://das-g.github.io/django-girls-tutorial/book_<lang>.epub`, e.g. for English https://das-g.github.io/django-girls-tutorial/book_en.epub~

~Before merging this, we'll probably want to have more meaningful file names, e.g. `django-girls-tutorial_en.epub`.~

Results available from `https://das-g.github.io/django-girls-tutorial/django-girls-tutorial_<lang>.epub`, e.g. for English https://das-g.github.io/django-girls-tutorial/django-girls-tutorial_en.epub

closes #1785

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -23,10 +23,15 @@ jobs:
         uses: actions/checkout@v3
         with:
           persist-credentials: false
+      - uses: awalsh128/cache-apt-pkgs-action@v1
+        with:
+          packages: calibre
       - name: Install and Build
         run: |
           npm install
           npx honkit build
+          npx honkit epub
+          mv book_*.epub _book/
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
         with:
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,13 +25,14 @@ jobs:
           persist-credentials: false
       - uses: awalsh128/cache-apt-pkgs-action@v1
         with:
-          packages: calibre
+          packages: calibre rename
       - name: Install and Build
         run: |
           npm install
           npx honkit build
           npx honkit epub
-          mv book_*.epub _book/
+          rename 's/^book/django-girls-tutorial/' book_*.epub
+          mv *.epub _book/
       - name: Upload artifact
         uses: actions/upload-pages-artifact@v1
         with:
--- a/en/README.md
+++ b/en/README.md
@@ -9,6 +9,11 @@ Welcome to the Django Girls Tutorial! We are happy to see you here. :) In this t
 
 As with all unknown things, this is going to be an adventure - but no worries, since you already worked up the courage to be here, you'll be just fine. :)
 
+{% if output.name != "ebook" %}
+> Do you want to read this tutorial on your e-reader?
+> Then [download the Django Girls Tutorial as an e-book in ePub format](../django-girls-tutorial_{{ book.language }}.epub).
+{% endif %}
+
 ## Introduction
 
 Have you ever felt that the world is more and more about technology to which you cannot (yet) relate? Have you ever wondered how to create a website but have never had enough motivation to start? Have you ever thought that the software world is too complicated for you to even try doing something on your own?


.github/workflows/deploy.yml
```diff
@@ -23,10 +23,16 @@ jobs:
...
+      - uses: awalsh128/cache-apt-pkgs-action@v1
+        with:
+          packages: calibre
+          version: 1.0
       - name: Install and Build
         run: |
           npm install
           npx honkit build
+          npx honkit epub
+          mv book_*.epub _book/
```
ekohl, line 35: I like the filename suggestion, so something like this?
```suggestion
          rename book django-girls-tutorial book_*.epub
          mv *.epub _book/
```

.github/workflows/deploy.yml
```diff
@@ -23,10 +23,16 @@ jobs:
         uses: actions/checkout@v3
         with:
           persist-credentials: false
+      - uses: awalsh128/cache-apt-pkgs-action@v1
```
ekohl, line 26: I had not seen this action before. I'm not sure how much the cache really saves, but I do like the style

.github/workflows/deploy.yml
```diff
@@ -23,10 +23,16 @@ jobs:
...
+        with:
+          packages: calibre
+          version: 1.0
```
ekohl, line 29: Is it needed to specify a version? Reading through the source I see this is the cache version, not the calibre version. I'd suggest to drop it and only add it once we need it. You can also purge caches via the UI nowadays so it's not needed as much anymore.

ekohl (commented):
I wonder how people discover we have these. Is there anything that links to it?

.github/workflows/deploy.yml
das-g: I can't seem to get that to work. I've tried:
- just this suggested change das-g@3d8a98dfe
  → `rename: command not found` ([failed workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838171061))
- with Ubuntu package `rename` das-g@ec674e3
  → `Bareword "book" not allowed while "strict subs" in use at line 1, in:
book` ([failed workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838214073/jobs/8622550060#step:4:248))
- with Ubuntu package `util-linux` & command `rename.ul` das-g@a988e99
  → `rename.ul: command not found` ([failed workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838303349/jobs/8622679182#step:4:248))
- with Ubuntu package `util-linux` & command `rename` das-g@1fe6e5e
  → `rename: command not found` ([failed workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838364548))
- with Ubuntu package `rename`, command `file-rename` and escaped quotes around the word `book` das-g@61ae7d1
  → ["successful" workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838467289), but files not renamed, as can be seen [in the subsequent workflow](https://github.com/das-g/django-girls-tutorial/actions/runs/4838516940/jobs/8623011212#step:4:248) where I [made `mv` verbose](https://github.com/das-g/django-girls-tutorial/commit/c3246d7a83ec5ae5c6b534e43ed0f20e5b1d966a#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R36).

.github/workflows/deploy.yml
das-g: I used it so I don't have to remember how to make apt well-behaved for non-interactive use. From the top of my head, that'd at least involve
- use old `apt-get` commands instead of the new `apt` ones
- call `apt-get update` before `apt-get install`, as otherwise the packages might not be known
- use `-y` option for both these commands
- set environment variable `DEBIAN_FRONTEND` to `noninteractive` before all that

I haven't known this action before either, but seen it recommended in [this Stack Overflow answer](https://stackoverflow.com/a/74498019/674064) to the question [How to apt-get install in a GitHub Actions workflow?](https://stackoverflow.com/q/57982945/674064).

.github/workflows/deploy.yml
das-g: > I think my suggestion to call both rename and mv was a copy paste mistake.

Was it? We need the results to end up in the `_book/` directory for them to be deployed.

das-g:
> Is there anything that links to it?

Not yet. But https://github.com/DjangoGirls/tutorial/blob/master/LANGS.md?plain=1 has the paths to the HTML versions of the various translations hardcoded anyway, so I guess we could also include hardcoded links to the ePub files?

.github/workflows/deploy.yml
das-g: Ah, I wasn't aware that perl rename and util-linux rename have that much different conventions in how you specify the replacement.

ekohl:
I think that's a good idea.

PR #1806, opened 2023-10-08
Simplify Google Fonts link element

thibaudcolas:
Simplifies the `link` element used for Google Fonts, so it’s easier for beginners to follow:

- Its attributes should use the same quote style as other elements.
- `rel="stylesheet"` should come first, to match other `link` elements.
- No need for a protocol-relative URL, we can load the HTTPS version everywhere.
- `type="text/css"` does nothing, it can be removed.


das-g (approved):

das-g:
Thank you!

PR #1807, opened 2023-10-08
Provide more information for RunCode users choosing a workspace

thibaudcolas:
After selecting the "Blank" workspace type, there is another step to select a size. I think it helps if the tutorial explicitly says which size should be used. And acknowledges it takes a while for it to start.

Additionally, in "intro to command line", it’d be clearer for users if they reused the workspace they’ve created before, rather than having multiple ones to manage.


amakarudze (approved):
Thanks for tackling this @thibaudcolas. It was on my to-do list.

PR #1808, opened 2023-10-08
Simplify RunCode virtual environment installation and usage instructions

thibaudcolas:
This essentially copies the "Virtual environment" and "Working with virtualenv: Linux and OS X" over to RunCode. There is no need to install and use `virtualenv` on RunCode specifically, the built-in `python -m3 venv myvenv` works fine there as well.

And for other environments we use `source` rather than `.` to initialise the virtual environment, so we should do the same here for consistency.

### DIFF
--- a/en/cloud_development_setup/instructions.md
+++ b/en/cloud_development_setup/instructions.md
@@ -34,29 +34,23 @@ $ mkdir djangogirls
 $ cd djangogirls
 ```
 
-We will make a virtualenv called `myvenv`. 
+We will make a virtualenv called `myvenv` by running `python3 -m venv myvenv`.
+It will look like this:
 
-To create a new `virtualenv` on RunCode, you first need to install the `virtualenv` module. To do so, first you need to update
-the packages in your environment
->```
->$ sudo apt-get update -y
->```
-> then install `virtualenv` by running the command:
->```
->$ sudo apt-get install -y virtualenv
->```
+{% filename %}command-line{% endfilename %}
+```
+$ python3 -m venv myvenv
+```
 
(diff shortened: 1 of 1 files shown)


das-g:
Would it make sense to simply include `en/django_installation/instructions.md`, like we do over there in the all-encompassing "installation" chapter:
https://github.com/DjangoGirls/tutorial/blob/c3588ff3d1b48be4fb092c8dad86e3ce93feb5b9/en/installation/README.md#L61-L64
rather than duplicating the same instructions?

thibaudcolas:
Good point @das-g! That makes sense to me. My only concern would be whether it’s clear enough RunCode users are meant to follow the "Linux" instructions, but since that’ll be the case elsewhere in the tutorial, seems reasonable.

das-g (approved):

das-g:
> Would it make sense to simply include `en/django_installation/instructions.md`, like we do over there in the all-encompassing "installation" chapter:
> 
> https://github.com/DjangoGirls/tutorial/blob/c3588ff3d1b48be4fb092c8dad86e3ce93feb5b9/en/installation/README.md#L61-L64
> 
> rather than duplicating the same instructions?

> Good point @das-g! That makes sense to me. My only concern would be whether it’s clear enough RunCode users are meant to follow the "Linux" instructions, but since that’ll be the case elsewhere in the tutorial, seems reasonable.
> 
> Should I make that change now or wait for more feedback on this?

Actually, let's do that in a separate pull request. Your changes here seem like a clear improvement and uncontroversial to me, so let me merge them right away.

Including instructions that also contain windows-specific parts in a RunCode specific page might be more controversial, so let's separate that out.

PR #1809, opened 2023-10-08
remove suggestions to use Jupyter Lab workspace

### LINKED ISSUE #1805: Jupyter Lab workspace option is confusing in RunCode
### Issue description

For RunCode users, the tutorial suggests two alternative workspaces: "Blank" – which is VS Code Server on Ubuntu, and Jupyter Lab. I don’t understand why there is a need to present users with a choice here – why not recommend the "Blank" VS Code on Ubuntu for everyone?

### Operating system

RunCode

---

For people choosing Jupyter Lab, there are a few more points of friction:

> Altenatively, you can go to Workspaces → New Workspace → Jupyter Lab. This will open a Python prompt which is depicted by `>>>`, you can type `exit()` to get back to the Ubuntu terminal.

This doesn’t open a Python prompt, it opens to a Jupyter Lab dashboard view. There is a "Console - Python" option on there but it’s IPython within a bespoke interface, not a standard `>>>` Python prompt.

There is a "Terminal" option that starts a bash command line, from which you can use `python` or `python3`.

They also get affected by #1739, while people who chose the "Blank" workspace aren’t.

So I’m unclear why the tutorial suggests the two options?

das-g:
fixes #1805

Changes in this pull request:

- Remove suggestions to use Jupyter Lab workspace in RunCode, as (for this tutorial) there's probably no advantage to use a Jupyter Lab workspace over using RunCode's default VSCode based workspaces, but indeed some disadvantages, as #1805 mentions. (Fewer unneeded options to choose from also simplifies the work of coaches and of those maintaining the tutorial.)
- Remove all mentions of how to open a command line window or Python prompt in a Jupyter Lab workspace

- Completely remove the special case for RunCode users in the "Python prompt" section, as at that point, they've already been told to follow Linux instructions.

### DIFF
--- a/en/cloud_development_setup/instructions.md
+++ b/en/cloud_development_setup/instructions.md
@@ -15,8 +15,6 @@ which you can sign up with.
 ## Command Line
 To open the Ubuntu terminal on RunCode, go to Workspaces → New Workspace → Blank. This will open a new Visual Studio Code workspace which has an Ubuntu terminal in the bottom pane.
 
-Altenatively, you can go to Workspaces → New Workspace → Jupyter Lab. This will open a Python prompt which is depicted by `>>>`, you can type `exit()` to get back to the Ubuntu terminal.
-
 Ubuntu is a version of Linux so for all command line instructions later in the tutorial you can follow Linux instructions.
 
 ## Virtual Environment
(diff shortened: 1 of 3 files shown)


nikhiljohn10 (approved):
LGTM

thibaudcolas (commented):
This looks great to me, unless there is a reason why the alternative is there that I don’t understand.

ekohl (approved):

PR #1815, opened 2023-12-12
Fix variable substitution in German tutorial

mraerino:
It seems variable substitution only works in fenced not indented code blocks.

---

Changes in this pull request:

- Fix variable substitution in German tutorial


das-g:
Huh. We manage translations through Crowdin. I don't know whether we can control what kind of code block Markdown markup it produces for translations of code blocks.

Maybe we have to report this to Crowdin, so that they preserve code block style of the original language in translated Markdown documents.

das-g (approved):

PR #1816, opened 2024-01-07
Update Python and Django versions.

FlipperPA:
Changes in this pull request:

Python 3.6 is no longer supported, and 3.8 has a few more months. I took two people through the tutorial yesterday, and Python 3.10 is the most recent version supported by Python Anywhere for deployment. So I've set the Python version to 3.10, Django version to 4.2, and minimum version to 3.8.

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,12 +1,12 @@
 {
     "gitbook": ">=3.2.0",
     "variables": {
-      "py_version": "3.8",
-      "py_release": "3.8.6",
-      "py_min_version": "3.6",
-      "py_min_release": "3.6.8",
-      "pa_py_version": "3.8",
-      "django_version": "3.2.10"
+      "py_version": "3.10",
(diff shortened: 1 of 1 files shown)


das-g:
Aren't any changes to the instructions and/or screenshots needed due to the change from Django 3 to Django 4?

FlipperPA:
@das-g I don't think I remember seeing any screenshots, most were included in the template at run-time, which is fantastic. The only screenshot I remember seeing was this one:

https://github.com/DjangoGirls/tutorial/blob/master/en/python_installation/images/python-installation-options.png

I'm happy to update it, but it didn't seem that critical because I believe the dialog is the same in current Python. I'm happy to mock up the screenshot to say `Python 3.#.#` and `Add Python 3.#` to path, so we don't have to change it in the future?

xrmx:
A complete upgrade has been done in #1820

PR #1818, opened 2024-01-15
Update instructions.md

callpri:
Codelobster code editor is added

Changes in this pull request:

-
-
-

### DIFF
--- a/en/code_editor/instructions.md
+++ b/en/code_editor/instructions.md
@@ -20,6 +20,12 @@ Sublime Text is a very popular editor with a free evaluation period and it's ava
 
 [Download it here](https://www.sublimetext.com/)
 
+## Codelobster
+
+Codelobster is a cross-platform editor with many plug-ins for all popular frameworks.
+
+[Download it here](https://www.codelobster.com/)
+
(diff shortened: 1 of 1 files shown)


kerhac:
I'm not associated with DjangoGirls, just passing by, saw this MR and thought I might add my two cents.

This list is a list of suggested simple source-code editors, it's not a complete list and does not list IDE. There are a tone of code editors out there which all deserve to be listed here, now that would overload reader, and they would not know which one to pick. That's why it makes sense to keep the list short. I personally think the listed editors are well-chosen even though I don't use any of them. My preferred editor would be neovim which is not a good choice for beginners.

Now codeLobster is an IDE and not a source-code editor and therefor should also not be listed here just like PyCharm which is not listed either.

Regards Julian

das-g:
I (as a (somewhat inactive) maintainer of the tutorial) agree with https://github.com/DjangoGirls/tutorial/pull/1818#issuecomment-2118844032 and am thus closing this Pull Request as "won't fix". Feel free to re-open it if you can name compelling reasons why this Software should be listed anyway.

Issue #1739, opened 2021-09-06
Github has removed password authentication and now requires an access token


jlgimeno:
### Issue description

Beginning August 13, 2021, Github now requires an access token to access remote repositories.  See link [to docs here](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/about-authentication-to-github#https).

### Language

Not currently related to any language

### Operating system

This will apply to all OS's, as the issue is Github's.

das-g:
see three possibilities:
1. Let participants set up an SSH key and use SSH for cloning and pushing (for only pulling, as we do on PythonAnywhere, it shouldn't matter if the repo is public)
2. Use personal access tokens
3. Use GitHub CLI for cloning and pushing

Are there any other options?

We'd want to select whatever is most simple (and doesn't make the tutorial much longer) while also being reasonably fool-proof. I'm unsure which option strikes the best balance for these criteria, though.

DataCerealz:
I just did a remote event end of September.
We went for access tokens.
That's definitely the easiest way to go in my/our opinion.
They can be easily created at the same time the PythonAnywhere Token gets created during the installation party.

I wouldn't use GitHub CLI. Reason: there are already a lot of new tools for participants. Don't make it even harder.

I also would highly favour AGAINST SSH. Reason: there are a lot of new concepts for participants. Like: what are servers? what is a terminal? Python, HTML, CSS, Django? What even is a Framework? What is Git? ... and so on. It's almost impossible to go through the entire tutorial on a single workshop day (ignoring the installation party here as there is no coding involved). Also implementing / Teaching SSH would just add another step of complexity and another new concept to be taught.
And to be honest: I am a full time dev. Even a Tech Lead in my current Job. And I forget about SSH all the time. I use a tool to manage my keys as I constantly forget how it works or where they are on different operating systems so I just use a management tool. I think understanding and using and remembering this on top of all the other new stuff is just asking too much.
There is a lot of "fuss" around the real Django development already; like hosting, git, a code editor, terminal commands, and so on.
But the core of the tutorial is a working Django web app. To archive this in the really limited time of a full day for coding, I'd try to strap away as much as possible that's not strictly needed to achieve this goal.

nikhiljohn10:
[...]

PR #1821, opened 2024-03-18
add a GitHub auth tip

lilspad:
Changes in this pull request:

- Add a note in the "Configuring our site on PythonAnywhere" section in the "deploy" directory to:
    - Explain why password authentication might not work
    - Guide the user to use SSH instead

Notes: 

- I normally follow the github docs to create a new SSH key, but the bash console required small syntax adjustment:
    - As `pbcopy` is not defined, use `cat` to display the key and copy manually instead

- I personally ran into this issue at this exact point in the tutorial, but it's possible that there is a more appropriate place to mention this, earlier in the tutorial.

Hopefully my explanation makes sense and helps anyone else who might be struggling to access their github from the command line. Please let me know if there's a better way to contribute this info!

### DIFF
--- a/en/deploy/pythonanywhere.md
+++ b/en/deploy/pythonanywhere.md
@@ -29,6 +29,45 @@ Now we run the helper to automatically configure our app from GitHub. Type the f
 $ pa_autoconfigure_django.py --python={{ book.pa_py_version }} https://github.com/<your-github-username>/my-first-blog.git
 ```
 
+> **Note** This will ask for your GitHub username and password. GitHub no longer supports password authentication and running this command could display an error like this: 
+> ```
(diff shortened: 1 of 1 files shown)


lilspad:
I just realised this is already an open issue (#1739) sorry! 
I'll leave this request open in case this might help temporarily, but feel free to close it if not :)

dennybiasiolli:
@lilspad is this happening if the repo is set to "public" too?
I tried the tutorial a couple of weeks ago and it didn't ask for username/password when cloning the repo.
I will try again after creating a clean local virtual machine, to see if it's happening on localhost too

lilspad:
@dennybiasiolli Ah, you're absolutely right! This is not an issue if the repo is set to public. Thank you!

PR #1825, opened 2024-04-21
instructions to create github access token

0orja:
This error was frequently encountered when pushing/pulling from github


das-g:
Probably related to #1739

ekohl (approved):

ekohl:
> pulling from github

I noticed this too with one participant. With here approval we decided to make the tutorial public instead of private.

amakarudze (approved):

PR #1826, opened 2024-04-21
Update versions

zahidkizmaz:
Changes in this pull request:

- Update min python version to 3.9.19
- Update default python version to current latest 3.12.3
- Update pythonanywhere version to latest available in pythonanywhere 3.10
- Update django_version to latest LTS version 4.2.11

I tested the django code with python 3.12.3. Worked as expected.

### DIFF
--- a/sk/django_views/README.md
+++ b/sk/django_views/README.md
@@ -42,4 +42,3 @@ Ulož súbor, prejdi na http://127.0.0.1:8000/ a pozri sa, čo sa udeje.
 Ukazuje, že server aspoň znovu beží, ale nie je to úplne ono, že? Neboj sa, je to len chybová stránka, nič, čoho sa treba báť! Podobne ako chyby v konzole nám tieto chyby napovedajú, čo treba robiť. Dočítaš sa, že *TemplateDoesNotExist* (šablóna neexistuje). Poďme opraviť túto chybu tým, že vytvoríme šablónu v nasledujúcej kapitole!
 
 > Viac informácií o Django viewoch získaš v oficiálnej dokumentácii: https://docs.djangoproject.com/en/4.2/topics/http/views/
-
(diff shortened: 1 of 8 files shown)


sk/django_views/README.md
```diff
@@ -41,4 +41,5 @@ Ulož súbor, prejdi na http://127.0.0.1:8000/ a pozri sa, čo sa udeje.
 
 Ukazuje, že server aspoň znovu beží, ale nie je to úplne ono, že? Neboj sa, je to len chybová stránka, nič, čoho sa treba báť! Podobne ako chyby v konzole nám tieto chyby napovedajú, čo treba robiť. Dočítaš sa, že *TemplateDoesNotExist* (šablóna neexistuje). Poďme opraviť túto chybu tým, že vytvoríme šablónu v nasledujúcej kapitole!
 
-> Viac informácií o Django viewoch získaš v oficiálnej dokumentácii: https://docs.djangoproject.com/en/3.2/topics/http/views/
\ No newline at end of file
+> Viac informácií o Django viewoch získaš v oficiálnej dokumentácii: https://docs.djangoproject.com/en/4.2/topics/http/views/
```
ekohl, line 44: Also changing 1 file in slovak looks inconsistent. Was this intentional?

zahidkizmaz: Thank you for the review. It was a mistake on my side, I removed the new line.

ekohl: I still see the change.
```suggestion
> Viac informácií o Django viewoch získaš v oficiálnej dokumentácii: https://docs.djangoproject.com/en/3.2/topics/http/views/
```

zahidkizmaz: I am happy to remove it but his is a reference to old version. Why shouldn't we update it?

ekohl: Is this really the only thing that's needed to update the Slovak guide? If I look at the English guide it shows up in multiple places so I fear it's inconsistent

zahidkizmaz: I agree, upon further examination, I can see that other languages also use `3.2` in their links. I believe it would be much better if links didn't rely on hard-coded versions, but rather used a variable sourced from a `book.json`. 

I can imagine having an additional variable, such as: `django_version_major_minor = 4.2`, which would be utilized for generating links. This approach would prove helpful in keeping our links and Django version synchronized. I can see this change as a separate PR, since it's not directly related to the update itself. 
What are your thoughts?

With this, I plan to remove my changes from the Slovak guide within this PR.

amakarudze (approved):
LGTM!

PR #1835, opened 2024-05-18
Implement previews for GitHub pull requests

ekohl:
When a contributor submits a PR, we always perform a build. This takes it a step further and uploads that a custom surge.sh domain. It adds a sticky comment to link to that preview while also generating some diffs.  This means reviews easier.

In the implementation an additional preview step is added. This first builds the base (target of the PR) as the current. Then it downloads the generated preview that was added as an artifact in the original build step. Creating a reasonably sized diff was tricky, because there's a long Javascript line that includes the mtime, making it indeterministic.  That line isn't relevant anyway, so it's removed. The diff command also ignores the search index.

All of that is placed in the preview, making it readable. A sticky comment is added with a summary, making it easy to use. The sticky comment is updated for every push, rather than added a comment for every push. This keeps the PR conversation usable.

This work is based on what's in https://github.com/theforeman/foreman-documentation where I found previews and generated diffs extremely useful in reviewing things. We'll need to set up a surge.sh account and store the token as a secret.

I've done testing in https://github.com/ekohl/tutorial/pull/9 but I may still have missed some parts. For example, that was fro, the repo itself while this is a PR from a fork.

Another aspect is that it would be nice to reuse the tarball that was built to generate the website itself. From my research I found that https://github.com/marketplace/actions/download-workflow-artifact exists to download artifacts from another workflow. We'll need to increase the time to delete since it defaults to 1 day, which makes the cache rather inefficient. I also don't know if you can download artifacts from another repository (like with a PR). That's why it's now generating it inside the workflow itself.

I chose to not run the current website build in a separate workflow to avoid storing an additional artifact, but in theory we can generate both current and preview in parallel and then only do the diff in the preview step.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,3 +15,87 @@ jobs:
         run: |
           npm install
           npx honkit build
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: github-pages-pr-${{ github.event.pull_request.number }}
+          path: _book/**
+
+  preview:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - name: Set preview domain
+        run: echo "PREVIEW_DOMAIN=$(echo ${{ github.repository }} | tr / - )-${{ github.job }}-pr-${{ github.event.pull_request.number }}.surge.sh" >> $GITHUB_ENV
+
+      - name: Install surge
+        run: npm install surge
+
+      - name: Install diffstat and pygments
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y diffstat python3-pygments
+
+      # TODO: can this download from the existing pages or a cache?
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.base.repo.full_name }}
+          ref: ${{ github.event.pull_request.base.ref }}
+          persist-credentials: false
+
+      - name: Build current pages
+        run: |
+          npm install
+          npx honkit build
+          mv _book current
+
+      - name: Download preview pages
+        uses: actions/download-artifact@v4
+        with:
+          # TODO: store output in upload step?
+          name: github-pages-pr-${{ github.event.pull_request.number }}
+          path: preview
+
+      - name: Remove indeterminism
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -15,3 +15,87 @@ jobs:
...
+      - name: Create diff to current
+        run: |
+          diff -Nrwu --exclude search_index.json current/ preview/ | cat > preview/diff.patch
+          if [[ -s preview/diff.patch ]] ; then
+            pygmentize -o preview/diff.html -l diff -f html -O full preview/diff.patch
+            diffstat -l -p2 preview/diff.patch > diff.txt
+          fi
+
+      - name: Deploy to surge.sh
+        run: npx surge ./preview $PREVIEW_DOMAIN --token ${{ secrets.SURGE_TOKEN }}
```
ekohl, line 74: I just realized: this doesn't have access to the token since it runs from the PR context. I think that's why in `foreman-documentation it runs after `deploy` finished:
https://github.com/theforeman/foreman-documentation/blob/16ca70658a09af2b0358bd7c782edf2fc5327e3c/.github/workflows/preview.yml#L3-L8
Note it also has some more code to run if it failed vs passing. That kind of error message isn't posted now.

This needs a bit more work.

evgeni: You can use `pull_request_target` trigger, which has access to secrets. But then the whole workflow would need to run as that.
In foreman we do this two-stage approach which I also don't like, but the result is similar.

ekohl:
I finally had some time to play with previews again. In https://github.com/ekohl/tutorial/pull/11 I tested this and mostly found why it didn't work. The only issue I still have is that `diff.patch` still shows up as `Gone` (and HTTP code 410) and I can't quite explain why.

However, I think this already has value for people. It does require an admin to set up a [surge.sh](https://surge.sh/) token and I don't have those permissions. In my testing I created an account for myself but it's better if the DjangoGirls repo uses its own.

fltfx:
Hey @ekohl, would it be possible to make you an admin and then you could setup the token? This is a bit out of my expertise. I think you can contact @amakarudze

ekohl:
That works for me

PR #1848, opened 2024-09-25
Added information about Mac M1/M2/M3 git installation and Github access token instructions

fltfx:
Changes in this pull request:

- https://tutorial.djangogirls.org/en/deploy/: Added information on how to resolve using homebrew to install git for Mac M1/M2/M3 users as I think there is a bug that comes up for Mac users using the M1/M2/M3 Apple Silicon processors vs Mac Intel processor variants. We ran into an issue with a Macbook 2022 - M2 version with using brew to install git, so that's where this is stemming from.
- https://tutorial.djangogirls.org/en/deploy/: Multiple attendees ran into the issue of needing to create a Github personal access token to push their code for the first time. I added further instructions on how to generate that token and how to enter it.


kewtree1408 (approved):
Thanks for the contribution <3 
LGTM.

PR #1849, opened 2024-10-11
Remove gitter.im links from the homepage and the sidebar

mcagl:
The `gitter.im` chat seems to be abandoned between DjangoGirls events, and is filled with **very bad** spam (I'm not repeating here the things that are inside the chat), so maybe it's better to remove the links for now, if the chat isn't moderated, to avoid looking bad if someone clicks on the link(s) and enters there.

Changes in this pull request:

- Removed the `gitter.im` links from the `README.md` file
- Removed the `gitter.im` link from the sidebar

I did the work only in the English language file.

### DIFF
--- a/book.json
+++ b/book.json
@@ -9,9 +9,6 @@
       "django_version": "4.2.11"
     },
     "links": {
-        "sidebar": {
-            "Need help? Talk to us!": "https://gitter.im/DjangoGirls/tutorial"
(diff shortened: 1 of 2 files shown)


amakarudze (approved):
Thank you for your work on this @mcagl. Was considering get to it this weekend.

PR #1856, opened 2024-11-19
Use Ubuntu 24.04 for rename from util-linux

### LINKED ISSUE #1855: The build in the CI is broken
### Issue description

[This pipeline run](https://github.com/DjangoGirls/tutorial/actions/runs/11323400245/job/31485981412) did not build the site after the merge of [this PR](https://github.com/DjangoGirls/tutorial/pull/1852) that updated the tutorial to newer python/django, so the deployed website is still at the old versions.

### Language

It is a problem with the CI

### Operating system

It is a problem with the CI

ekohl:
<del>I'm unsure why this broke.</del> I also looked at https://honkit.netlify.app/ebook which mentions you can publish an ebook with a specific name, but figured this was the shortest path. It's untested now.

Right after submitting this I had a theory why it broke. On the first run it passed, but after that it started to fail. The first run it couldn't have cached the package with `awalsh128/cache-apt-pkgs-action` while it has on the later runs.

`/usr/bin/rename` is provided via alternatives:

```
Setting up rename (2.02-1) ...
update-alternatives: using /usr/bin/file-rename to provide /usr/bin/rename (rename) in auto mode
update-alternatives: warning: skip creation of /usr/share/man/man1/rename.1.gz because associated file /us
r/share/man/man1/file-rename.1p.gz (of link group rename) doesn't exist
```

I suspect `awalsh128/cache-apt-pkgs-action` ignores alternatives so `/usr/bin/rename` isn't present when a cache is used.

This updates to Ubuntu 24.04 and relies on `rename` from `util-linux` which bypasses the need to cache it altogether.

I can't think of an easy way to test this.

Fixes #1855

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -17,7 +17,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -25,7 +25,7 @@ jobs:
           persist-credentials: false
       - uses: awalsh128/cache-apt-pkgs-action@v1
         with:
-          packages: calibre rename
+          packages: calibre
       - name: Install and Build
         run: |
           npm install
@@ -46,7 +46,7 @@ jobs:
     environment:
       name: github-pages
       url: ${{ steps.deployment.outputs.page_url }}
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - name: Checkout
         uses: actions/checkout@v4


das-g:
> I can't think of an easy way to test this.

Can we make the stuff that runs for PRs (currently defined in `.github/workflows/test.yml`) more similar to that that would run on `master` (currently defined in `.github/workflows/deploy.yml`)? IMO the only difference should be that PRs shouldn't deploy to the production location.

That way, the checks for PRs would test almost everything of the productive CI/CD (e.g. also the renaming).

ekohl:
> Can we make the stuff that runs for PRs (currently defined in `.github/workflows/test.yml`) more similar to that that would run on `master` (currently defined in `.github/workflows/deploy.yml`)? IMO the only difference should be that PRs must not deploy to the production location.

https://github.com/DjangoGirls/tutorial/pull/1835 was the start for that, but I rarely have the time to work on Django things. That one was mostly done during a DjangoGirls workshop where the students didn't need a lot of guidance.

ekohl:
This is really weird: https://github.com/DjangoGirls/tutorial/actions/runs/11917391625/job/33212329587 failed on
> /home/runner/work/_temp/0bb45bc1-4c28-4c5b-a117-ee1a9563fd3c.sh: line 4: rename: command not found

But then later versions did deploy correctly. Almost like it was always using the old workflow file. Perhaps that is why the original change only started to break after the first run?

This is not how I thought GHA was supposed to work.

Issue #1858, opened 2024-11-26
[Portuguese] "Deploy!" translation needs changes


Anemaygi:
### Issue description

In the "Deploy!" step of the pt-br tutorial, two updates are required:

- GitHub no longer allows setting the origin using HTTP basic authentication. Instead, you should either generate a personal access token and include it in the URL or use the SSH address.

- Add the --python=3.10 flag to the pa_autoconfigure_django.py command. Without this, the wrong Python version may be used, leading to an error.

### Language

Portuguese (pt-br)

fltfx:
Thanks for the PR.

PR #1859, opened 2024-11-26
[Portuguese] Add "--python=3.10" flag to "pa_autoconfigure_django" command in the "Deploy!" step

Anemaygi:
Changes in this pull request:

- Add `--python=3.10` to `pa_autoconfigure_django` command in the "Deploy!" step of the portuguese tutorial. Without this flag, the incorrect version of Python is used, causing an error related to the Django version.

Solves one of the problems from #1858

### DIFF
--- a/pt/deploy/README.md
+++ b/pt/deploy/README.md
@@ -155,7 +155,7 @@ Agora vamos executar a ferramenta para configurar a nossa aplicação a partir d
 
 {% filename %}linha de comando do PythonAnywhere{% endfilename %}
 
-    $ pa_autoconfigure_django.py https://github.com/<your-github-username>/my-first-blog.git
+    $ pa_autoconfigure_django.py --python=3.10 https://github.com/<your-github-username>/my-first-blog.git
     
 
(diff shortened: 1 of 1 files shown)


dannluciano:
Recently I have the same issue.

fltfx:
@das-g Hi Raphael, can we do something about the required checks? Would like to merge this one. Thanks!

das-g:
> @das-g Hi Raphael, can we do something about the required checks? Would like to merge this one. Thanks!

I don't think I have the required access level to circumvent required checks. For that you'll probably have to ask [the owners](https://github.com/orgs/DjangoGirls/people?query=role%3Aowner) of the DjangoGirls GitHub organization.

As for why no status was reported by the build, I have no idea. Maybe @ekohl has some insight into that? https://github.com/DjangoGirls/tutorial/pull/1859/checks says **"Workflow runs completed with no jobs"**, which isn't what I'd expect from our CI setup.

das-g:
Closing to be able to re-open to force a rebuild.

das-g:
Reopening to force a rebuild.

PR #1866, opened 2025-03-12
Django Forms: rename field to attribute to be more accurate; fix typo

Oktosha:
In my previous fix #1864 to the Django Forms page I introduced an inaccuracy: I called an instance member "field" instead of "attribute". That pull request fixes that. Thank @das-g for noticing.

I also noticed a typo in the same sentence: I wrote "form" instead of "from". I fix it, too.


das-g (approved):
:+1:

PR #1877, opened 2025-09-20
Default Prompt on MacOS is now %, not $

hjwp:
Update prompt section to reflect that zsh on mac now defaults to %

NB:
* [x] Stacked from https://github.com/DjangoGirls/tutorial/pull/1876, need merge that first or rebase.

### DIFF
--- a/en/intro_to_command_line/prompt.md
+++ b/en/intro_to_command_line/prompt.md
@@ -1,16 +1,4 @@
-<!--sec data-title="Prompt: macOS and Linux" data-id="OSX_Linux_prompt" data-collapse=true ces-->
-
-
-If you're on Mac or Linux, you probably see a `$`, like this:
-
-{% filename %}command-line{% endfilename %}
-```
-$
-```
-<!--endsec-->
-
 <!--sec data-title="Prompt: Windows" data-id="windows_prompt2" data-collapse=true ces-->
-
 
 On Windows, you probably see a `>`, like this:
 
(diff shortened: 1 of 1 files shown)


das-g (commented):
AFAIR, that change in the prompt sign is due to MacOS having changed from Bash as the default shell to zsh. I wonder whether any other parts of the tutorial need updating due to that.

hjwp:
> AFAIR, that change in the prompt sign is due to MacOS having changed from Bash as the default shell to zsh. I wonder whether any other parts of the tutorial need updating due to that.

hey @das-g, so somewhat dauntingly the answer is probably yes.  at the moment, most listings are split into two:  "on windows, do this.  on mac or linux, do that".  and the prompt in the "mac or linux" section is a `$`.  so to fully feed through this change, theoretically we'd have to change every single set of command blocks from having two sections (windows+mac/linux) to having 3 sections -- windows, mac, and linux separately.

that feels like a big and painful change.  do we really want to go ahead and do it?  bearing in mind the only difference between mac and linux will be the prompt, not the commands?    can we think of some more lightweight way to incorporate this change?

hjwp:
i want to say, people have been following the tutorial just fine on macos for the several years now despite there being a `$` in their command listings, so maybe it's worth merging now and thinking about how to fix all the listings in a separate PR?

hjwp:
options for (later? separate pr?) fixes to all listings

1. turn all commands from 2 blocks to 3 blocks.  cumbersome but accurate!
2. switch all macos/linux commands from `$` to `%` - maybe there are more mac users than linux ones, and the linux ones are less likely to be confused?
3. leave them as-is with the `$` -- dollar is a fairly common convention in online tutorials/documentation
4. get rid of all the prompts -- they confuse people anyway and make copy-pasting hard.

no perfect solution!  wdyt?

das-g (approved):
Looks good to me!

das-g:
I was more worried that our current Linux/Mac commands might use bashisms (Bash-specific syntax or idioms) somewhere. If indeed people have followed the tutorial fine on Mac (and thus presumably zsh) that's probably not the case.

Additional sections just for the different prompt are IMO not worth it, even if they were (or maybe are, as that should be amendable to (classically, i.e. non-LLM) automated editing, with —say— Vim, `(s)ed`, `awk` or a simple Python script) easy to add. Also, separate sections for Linux and Mac might give the wrong impression that CLI skills taught by the tutorial are less transferable between UNIX-like operating systems (such as Linux, Mac OS and BSD) than they actually are.

PR #1899, opened 2026-06-02
Removed old manual import in the shell 

raffaellasuardini:
Changes in this pull request:

I removed the following lines since in Django 5.2 has the [automatic import for shell ](https://docs.djangoproject.com/en/6.0/releases/5.2/#automatic-models-import-in-the-shell) so there is no need to import the model

`>>> from blog.models import Post`


ekohl (approved):
Since we updated in https://github.com/DjangoGirls/tutorial/commit/5132e10b0bd4215622b8db1aab41a2973f2e9dd1 to 5.2, this should be OK. I didn't verify it myself, but trusting it based on the linked release note.

PR #1902, opened 2026-07-18
add a gitignore entry for pycharm users

hjwp:
Was sitting next to someone that's using pycharm and I saw a load of IDE stuff getting committed into their repo...


das-g (approved):
LGTM

